PR #10, opened 2017-03-24
Moved adapter code into gateway.

dhylands:
What type of commenting do we want on the functions (like in thing.js?)

Is there any type of standardized commenting style?

### DIFF
--- a/adapters.js
+++ b/adapters.js
@@ -1,0 +1,50 @@
+#!/usr/bin/env node
+
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var EventEmitter = require('events').EventEmitter;
+
+class AdapterManager extends EventEmitter {
+
+    constructor() {
+        super();
+        this.adapters = [];
+        this.things = [];
+    }
+
+    addThing(adapter, thing) {
+        //console.log('add_thing', thing);
+        this.things.push(thing);
+        this.emit('device-added', thing);
+    }
+
+    getThings() {
+        return this.things;
+    }
+
+    loadAdapters() {
+        var adapter_dir = './adapters';
+        var adapter_manager = this;
+        fs.readdir(adapter_dir, function file_list(err, filenames) {
+            if (err) {
+                console.error(err);
+                return;
+            }
+            for (var filename of filenames) {
+                if (path.extname(filename) !== '.js') {
+                    continue;
+                }
+                let adapter_filename = adapter_dir + '/' + filename;
+                console.log('Loading adapter', adapter_filename);
+
+                let adapter = require(adapter_filename);
+                adapter_manager.adapters.push(new adapter(adapter_manager));
+            }
+        });
+    }
+}
+
+module.exports = new AdapterManager();
+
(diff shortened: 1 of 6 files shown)


adapters.js
```diff
@@ -0,0 +1,50 @@
...
+class AdapterManager extends EventEmitter {
+
+    constructor() {
+        super();
+        this.adapters = [];
+        this.things = [];
+    }
+
+    addThing(adapter, thing) {
+        //console.log('add_thing', thing);
```
benfrancis: Remove commented code

adapters.js
```diff
@@ -0,0 +1,50 @@
...
+        this.things.push(thing);
+        this.emit('device-added', thing);
+    }
+
+    getThings() {
+        return this.things;
+    }
+
+    loadAdapters() {
+        var adapter_dir = './adapters';
```
benfrancis: camelCase for variable names and function names

test-server.js
```diff
@@ -0,0 +1,93 @@
...
+
+var express = require('express');
+var app = express();
+var adapter_manager = require('./adapters.js');
+var bodyParser = require('body-parser');
+
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(bodyParser.json());
+
+const ROOT = '/things';
```
benfrancis: Note that currently this clashes with app.js which also tries to respond to HTTP requests to this path. This is fine for a test but we need to decide where this logic lives and it should probably eventually live in an Express routes/ directory. Also I don't think it should directly call the AdapterManager to get a list of Things, that should probably be handled by a separate ThingManager which gets a list of Things from a database and calls the AdapterManager to look for new Things or get or set properties etc.

adapters/foo.js
```diff
@@ -0,0 +1,46 @@
...
+
+    setPropertyValue(name, value) {
+        // This function should propogate the value to the hardware
+        // Calling super.setPropertyValue will update the value cache
+        // and send a 'value-changed' event to any listeners.
+        super.setPropertyValue(name, value);
+    }
+}
+
+class FooAdapter {
```
benfrancis: These two classes should probably be in separate files. Also definitions of types of Thing should really be adapter (protocol) agnostic, with some kind of binding to each underlying protocol. So eventually thing definitions like this should probably live somewhere else outside of the adapters folder.

dhylands: The Thing class is adapter agnostic. FooThing is the specific Thing that needs to be adapter aware. So FooThing and FooAdapter are just dummy classes that I used.

I put them in the same file since they're intimately tied together. For real adapters, they can go in separate files. My intention was to allow the adapters directory to have subdirectories for multi-file adapters.

test-server.js
dhylands: Using test-server.js and app.js at the same time isn't intended. I run test-server.js just as a test for exercising the adapter stuff. As soon as app.js can do get/set etc then I plan on removing test-server.js

adapters/foo.js
benfrancis: Makes sense.

test-server.js
benfrancis: Understood.

PR #11, opened 2017-03-27
Start on a Thing database using SQLite3

benfrancis:
Load /things from an SQLite database rather than a static JSON file on disk, and populate it with some default things on first run. Next step will be adding/removing things to database, then adding things from adapters. Includes creating a /things Express router which handles HTTP requests to /things by calling the database.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules/
+db.sqlite3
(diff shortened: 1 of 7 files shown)


benfrancis:
> I'm assuming that we'll actually use an index at some point, otherwise it doesn't seem to be useful to have a db.

Yes, currently just an automatically incrementing ID but may change this to a UUID or something else and add extra fields. Still not sure what metadata needs storing vs. being generated at run time.

PR #14, opened 2017-03-28
MVC structure with a Things model and controller

benfrancis:
I moved things around a bit to follow an MVC type pattern, what do you think?

I created a Things controller (which is currently just a simple URL router) and a Things model (which currently just acts as a proxy to the database but will eventually do much more). When the database schema gets more complex the Database object could handle ORM and the models will handle business logic, calling the database and adapters on the back end where needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,3 +21,22 @@ Install dependencies:
 ```$ node app.js```
 
 Load ```http://localhost:8080``` in your web browser.
+
+
+## Source Code Structure
+
+* **models/** - Data model and business logic for Things
+* **views/** - Templates which are rendered and served by controllers
+* **controllers/** - App URL routes and their logic
+* **adapters/** - Network adapter device drivers (e.g. ZigBee, Z-Wave)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -21,3 +21,22 @@ Install dependencies:
...
+* **static/** - Static HTML, CSS, JavaScript & image resources for web app front end
+  * **css/** - Style sheets
+  * **images/** - Image files
+  * **js/** - JavaScript files for the web app front end
+  * **index.html** - The home page of the gateway
+  * **things.json** - A default list of devices to populate the database with
+  * **app.webmanifest** - W3C web app manifest for front end
+* **app.js** - The main back end script
+* **db.js** - Manages the SQLite3 database
+* **adapters.js** - Manages the various network adapters of the gateway (e.g. ZigBee, Z-Wave)
```
dhylands: I'm thinking I'll rename this to adapter-manager.js since that's the name fo the class it contains (so nothing for you to do).

Issue #5, opened 2017-03-23
List smart plugs


benfrancis:
As a user I want to see a list of devices connected to the gateway

benfrancis:
Closed by https://github.com/moziot/gateway/commit/3496bd8439671c55c4e2db044e99dd58590936a4

PR #16, opened 2017-03-29
Placeholder Add Thing screen - start on #5

benfrancis:
Mostly just placeholder UI for the Add Thing screen which will need hooking up to the adapter manager.


dhylands:
I'm going to say this is fine for now. I think we need to discuss the whole pairing thing once again, now I've got the zwave adapter actually working.

PR #17, opened 2017-03-30
Added zwave support

dhylands:
This also adds support for pairing and unpairing. I wound up having to make a thing for the zwave controller that has pair and unpair booleans. Setting the pair boolean to true starts pairing mode and you then push the button on the device.

I introduced the notion of a thing_id (which I may need to back out). I figured I'd do it and we can discuss it and see where it goes.

I wanted to support multiple zwave dongles, and I wanted to ensure that each thing has a unique "id". I think that using a name like 'temperature' is too generic and we can't guarantee that it's unique.

Running thing.py I did the following:
```
dhylands@dave-work ~/Dropbox/moziot/gateway  (add-zwave)
2000 >./thing.py --things
Id:Foo1 Name: thing1
Id:Foo2 Name: thing2
Id:zwave-cc9eff28-1 Name: zwave-cc9eff28-1
Id:zwave-cc9eff28-4 Name: cc9eff28-4-Switch
dhylands@dave-work ~/Dropbox/moziot/gateway  (add-zwave)
2001 >./thing.py --things --properties
Id:Foo1 Name: thing1
  temperature = 23
  humidity = 35
  led = false
Id:Foo2 Name: thing2
  temperature = 23
  humidity = 35
  led = false
Id:zwave-cc9eff28-1 Name: zwave-cc9eff28-1
  pair = false
  unpair = false
Id:zwave-cc9eff28-4 Name: cc9eff28-4-Switch
  4-37-1-0 = false
dhylands@dave-work ~/Dropbox/moziot/gateway  (add-zwave)
2002 >./thing.py --things --set 4-37-1-0=true
Id:Foo1 Name: thing1
Id:Foo2 Name: thing2
Id:zwave-cc9eff28-1 Name: zwave-cc9eff28-1
Id:zwave-cc9eff28-4 Name: cc9eff28-4-Switch
property_url = http://localhost:8088/things/zwave-cc9eff28-4/4-37-1-0
body = {"name": "value"}
  Set 4-37-1-0 = true
dhylands@dave-work ~/Dropbox/moziot/gateway  (add-zwave)
2003 >./thing.py --things --set 4-37-1-0=false
Id:Foo1 Name: thing1
Id:Foo2 Name: thing2
Id:zwave-cc9eff28-1 Name: zwave-cc9eff28-1
Id:zwave-cc9eff28-4 Name: cc9eff28-4-Switch
property_url = http://localhost:8088/things/zwave-cc9eff28-4/4-37-1-0
body = {"name": "value"}
  Set 4-37-1-0 = false
```


dhylands:
I also noticed that when I git Control-C in test-server.js that I get a segmentation fault as part of unloading the ZWave stuff.

So I need build a debug version of OpenZWave and dig into this a bit more.

adapters/zwave/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+var loadZWaveAdapters = require('./zwave-adapter.js');
```
benfrancis: What's the purpose of this file? Doesn't AdapterManager just load ZWaveAdapter directly?

thing.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
```
benfrancis: What's the purpose of this file? Shouldn't the web app front end do this? Is this another test script/kind of integration test?

adapters/zwave/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
```
benfrancis: What's the purpose of this file? Doesn't AdapterManager just load ZWaveAdapter directly?

adapters/zwave/zwave-node.js
```diff
@@ -0,0 +1,61 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+'use strict';
+
+var Thing = require('../../models/thing.js').Thing;
+
+class ZWaveNode extends Thing {
```
benfrancis: I don't think ZWaveNode should extend Thing, see my comments below.

benfrancis:
One other question - what do you need to do to get the Z-Wave adapter to work? Should we add instructions to the README on downloading/building openzwave?

thing.py
dhylands: Yeah - that's basically just a command line tool for exercising the web API. I plan on rewriting it in node. When I was working on foxbox, I found that it sometimes took quite a while between when the WebAPI was ready and the web interface was able to use it, and without a command line tool, I wouldn't have been able to test anything.

adapters/zwave/index.js
dhylands: When you require a directory, it loads the index.js file from the directory.
See https://nodejs.org/api/modules.html#modules_all_together for the full set of loading rules used by require.

dhylands:
You should just need to to do npm install .  I think I added the appropriate dependencies to package.json

dhylands:
OK - I don't mind changing the base class, but here's what I was expecting:

Adapter API exposes 2 objects: Adapter and Device (or Node or Thing). These base classes would define the API.

I think that having the adapter decide on the unique IDs makes sense. Many protocols already have things which are unique and creating a mapping from one unique ID to another doesn't really make sense to me. If you really want UUIDs then we can do that. it just adds an extra layer and the adapters now have to map everything from the UUID ID space to their own ID space.

I don't understand why we would want ZWaveBinarySwitch or ZigBeeOnOffSwitch or who would use it. I thought that having something like properties would alleviate the need for that. I certainly don't think we need them for ZWave. I'll have to wait for ZigBee to see.

Perhaps we should move the existing adapters tree into adapters/plugins and move the adapter-manager.js, and adapters specific code into there?

I also don't have a problem with having the adapter manager exposing the adapters

dhylands:
Question: How should doing a set/get look like from the client side when using our restful API? I'm assuming for the get that we'll be doing an http GET request on /thing/thingid/propertyid and we'll return some type of JSON response. Is the response just the value? Or is it the entire property with a value member?

Similarly for a set. Should we do PUT or POST? Should the content just be the value? Or should it be a dictionary with a value member?

benfrancis:
@dhylands wrote:
> Adapter API exposes 2 objects: Adapter and Device (or Node or Thing). These base classes would define the API.

That's fine, it's just if that's what's needed then we should probably create another Thing/Device/Node class in adapters/ separate to the one in model/. Tell me if you disagree but I think they're different layers of abstraction (Web Thing vs. Z-Wave Node, WoT vs. IoT). If the adapters were in a separate process or even a different language then ZWaveNode wouldn't be able to extend the web app's Thing base class anyway. As I said, it's probably my bad for moving your thing.js into model/ because I didn't understand what it was going to be used for.

> I think that having the adapter decide on the unique IDs makes sense. Many protocols already have things which are unique and creating a mapping from one unique ID to another doesn't really make sense to me. If you really want UUIDs then we can do that. it just adds an extra layer and the adapters now have to map everything from the UUID ID space to their own ID space.

I understand it's extra work to map the IDs. The reason I suggested UUIDs is that it would guarantee they were unique. Otherwise isn't it possible that two adapters could assign the same ID to two different Things? Also these IDs will be used in URLs and it might be a bit strange to expose the different protocol naming schemes at the web level?

> I don't understand why we would want ZWaveBinarySwitch or ZigBeeOnOffSwitch or who would use it. I thought that having something like properties would alleviate the need for that. I certainly don't think we need them for ZWave. I'll have to wait for ZigBee to see.

In general I think we have to have standard device types (extensible through schemas) because it isn't feasible to automatically generate a meaningful UI for an arbitrary combination of properties, actions and events. We need defined device types (e.g. a light switch, a door sensor or a motion sensor) which the UI knows how to represent.

My idea was that we could map standard ZigBee/Z-Wave/Bluetooth device types onto Web Thing device types using non-normative [protocol bindings](https://docs.google.com/document/d/1oyTM3NPjQOOohTqkjfq1klHalfcEn5w9Ull0_wadbJw/edit?usp=sharing). Are you thinking there's another way of achieving that?

> Perhaps we should move the existing adapters tree into adapters/plugins and move the adapter-manager.js, and adapters specific code into there?

Yes it probably makes sense to move adapter.js at least. I think adapter-manager.js could be fine where it is because it makes sense for it to live next to db.js, but maybe each adapter could have its own subfolder in adapters/ (e.g. adapters/zwave/) and the base classes could live directly in adapters/ ? Just a suggestion.

> I also don't have a problem with having the adapter manager exposing the adapters

Cool.

> Question: How should doing a set/get look like from the client side when using our restful API? I'm assuming for the get that we'll be doing an http GET request on /thing/thingid/propertyid and we'll return some type of JSON response. Is the response just the value? Or is it the entire property with a value member?
>
> Similarly for a set. Should we do PUT or POST? Should the content just be the value? Or should it be a dictionary with a value member?

See https://moziot.github.io/wot/#property-resource - this was my initial idea, but open to suggestions. Updating a resource should always be a PUT, creating a new resource is a POST.

I did realise it's probably good practice for a set (PUT) to echo back the value that was sent in its response body though (see https://github.com/moziot/wot/issues/26).

dhylands:
> I don't understand why we would want ZWaveBinarySwitch or ZigBeeOnOffSwitch or who would use it. I thought that having something like properties would alleviate the need for that. I certainly don't think we need them for ZWave. I'll have to wait for ZigBee to see.
> 
> In general I think we have to have standard device types (extensible through schemas) because it isn't feasible to automatically generate a meaningful UI for an arbitrary combination of properties, actions and events. We need defined device types (e.g. a light switch, a door sensor or a motion sensor) which the UI knows how to represent.

From the UI side of the world, I can see where this makes sense. But the user has to provide the semantic interpretation to know that something is a light switch or not. There isn't any way for ZWave to know that.

Basically in ZWave, a node has a bunch of attributes (i.e. it has a switch, it has a temperature sensor) and those attributes will pretty much map one-to-one to properties that we're going to expose. From the adapter point of view, it just doesn't know if something is a light switch or not. So at least for zwave, setting a property should wind up setting an attribute on a node, and the adapter doesn't need any more specialization than a node. The attributes contain all of the information necessary. Anything above that is user semantics.

dhylands:
I reworked things so that the terminology is adapter/device/attribute (thing->device property->attribute).

I left models/thing.js and created device.js.

I'll work on the following next (but in separate PRs)
- add discoverAdapters/discoverDevices
- get gateway to serve up adapters and things.
- add UUIDs instead of adapter/device ids
- fix seg fault in openzwave library

benfrancis:
> Basically in ZWave, a node has a bunch of attributes (i.e. it has a switch, it has a temperature sensor) and those attributes will pretty much map one-to-one to properties that we're going to expose. From the adapter point of view, it just doesn't know if something is a light switch or not. So at least for zwave, setting a property should wind up setting an attribute on a node, and the adapter doesn't need any more specialization than a node. The attributes contain all of the information necessary. Anything above that is user semantics.

Hmm, I was worried this might be the case. This makes things very challenging from the UI point of view, without having a hard coded list of every device model that might ever be connected to the gateway, which is what I've been trying to avoid.

As I've said before my hope was that [ZigBee clusters](http://www.zigbee.org/zigbee-for-developers/applicationstandards/zigbeehomeautomation/) (e.g. on/off switch, dimmer switch, door lock, smart plug, simple sensor, on/off light, dimmable light, light sensor, occupancy sensor, temperature sensor, thermostat etc.), [Z-Wave command classes](http://docs.smartthings.com/en/latest/device-type-developers-guide/z-wave-primer.html) (e.g binary switch, multi-level switch, binary sensor, multilevel sensor, meter, alarm etc.) and [Bluetooth profiles](https://www.bluetooth.com/specifications/adopted-specifications) (whenever they finally get around to defining home automation profiles) could be mapped directly onto web thing device types. By supporting a defined list of simple device types out of the box we could support any device which adhered to that device type (or underlying cluster/class/profile) without needed vendor-specific integration.

The reality turns out to be more complex because the truth is that devices are so diverse that many (perhaps most) devices support multiple clusters/classes/profiles. This diversity/lack of standardisation at the device type level makes it challenging to provide a meaningful user interface without implementing support for individual product models (although I haven't entirely lost hope yet).

In some ways ZigBee is even worse in that most controllers are designed to only pair with a pre-defined list of product models, whereas at least Z-Wave should pair with any device because it has a "basic command class". But on the HomeWeb project we think we found a way around this by configuring the controller to allow any ZigBee device to be paired, being careful to put security mitigations in place to make sure the user is pairing the device they think they're pairing.

So in summary, if a Z-Wave device doesn't follow a single device type (which it appears they don't), then it might be challenging for the adapter to just assign it a web thing device type and have the web app generate a thing description. You're probably right that more metadata needs passing up the stack so that the web app can provide something useful to the user. I suspect this may take some experimentation to get right.

benfrancis:
>  I reworked things so that the terminology is adapter/device/attribute (thing->device property->attribute).
>  I reworked things so that the terminology is adapter/device/attribute (thing->device property->attribute).
>
> I left models/thing.js and created device.js.
>
> I'll work on the following next (but in separate PRs)
>
> - add discoverAdapters/discoverDevices
> - get gateway to serve up adapters and things.
> - add UUIDs instead of adapter/device ids
> - fix seg fault in openzwave library
> I left models/thing.js and created device.js.
> 
> I'll work on the following next (but in separate PRs)
> 
> - add discoverAdapters/discoverDevices
> - get gateway to serve up adapters and things.
> - add UUIDs instead of adapter/device ids
> - fix seg fault in openzwave library

Great, thanks.

With the above discussion, I'm starting to better understand why you wanted to use web thing terminology more at the adapter layer, because in the absence of obvious standard device types the adapter may be best placed to generate the thing description. We're going to need to experiment with this because at some point we do need to assign a web thing device type and somehow map a device onto a standard schema (whether that's done automatically, based on product-specific mappings or via some user interaction), otherwise everything will just be of type "thing" and the UI will be pretty awful!

The separation in terminology between the layers is probably a bit silly, but I still think it makes sense for ZWaveNode not to extend Thing because eventually they may run in separate processes/languages and it helps provide a cleaner separation between layers.

Going to merge this.

BTW, should adapter.js live in adapters/ at the top level?

dhylands:
Right now, the adapters directory is assumed to only contain adapters. The adapter manager scan that directory for .js files or other directories and assumes that these are adapters.

I can move adapters.js into the adapters directory if we change the extension of the adapters to be something other than .js or we move them into another directory tree (which is why I mentioned having a plugins directory under adapters).

ZWave also supports a bunch of command classes: http://wiki.micasaverde.com/index.php/ZWave_Command_Classes so you can tell the difference between a thermostat and a switch.

But you can't tell anything about the type of switch (and ZigBee is the same way). i.e. a wall switch and an appliance switch both look the same, so you can't tell the difference between a light switch, or a heater switch.

PR #21, opened 2017-03-31
Add discover

dhylands:
This adds discoverAdapters and discoverDevices.

I started to look at hooking them in, but I wasn't sure exactly where you wanted them.

I also experimented a bit with jsDoc and added a bit of jsDoc documentation. If this looks reasonable, I'll continue to add documentation this way.

You can install jsdoc by doing `npm install jsdoc` and then just run `./node_modules/.bin/jsdoc .` and open `out/index.html` to see the generated documentation.

### DIFF
--- a/adapter-manager.js
+++ b/adapter-manager.js
@@ -1,8 +1,9 @@
 /**
- * AdapterManager.
+ * Manages all of the Adapters used in the system.
  *
- * Manages Adapter data model and business logic.
- *
+ * @module AdapterManager
+ */
+/**
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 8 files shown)


dhylands:
Added controllers for adapters/, adapters/:adapterId/, adapters/:adapterId/devices/ and adapters/:adapterId/devices/:deviceId.

These are primarily for debug to query information directly from the adapter rather than what's filterd by the db. This also fixes up some minor bugs in discoverDevices & discoverAdapters now that the functions are actually exercised.

benfrancis:
Cool.

Haven't reviewed yet because I think you omitted some files from the commit.

```
Error: Cannot find module './controllers/adapters_controller'
```

dhylands:
Right you are.

benfrancis:
I believe this was covered in another pull request

PR #22, opened 2017-04-03
Updated README to add more dependency information.

dhylands:
This adds the dependency information needed for open-zwave.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,82 @@
 # MozIoT Gateway
 Web of Things gateway.
 
-## Get Started
+## Prerequisites
+
+### Install pkg-config
+
+Under ubuntu:
+```
+sudo apt-get install pkg-config
+```
+
+Under OSX:
+[This page](http://macappstore.org/pkg-config/) covers installing pkg-config
+under OSX.
+
+### Recommended: install nvm
+
+nvm allows you to easily install different versions of node. To install nvm:
+```
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
+```
+
+Close and reopen your temrinal window. Use nvm to install node and set the
+default version.
+```
+nvm install v7.7.2
+nvm use v7.7.2
+nvm alias default v7.7.2
+```
+
+Verify that node and npm have been installed:
+```
+$ npm --version
+4.1.2
+$ node --version
+v7.7.2
+```
+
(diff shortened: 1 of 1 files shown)


benfrancis:
Thanks for documenting this.

I can't help but feel this could be made simpler, there are multiple ways of installing these pre-requisites and some of them aren't strictly necessary.

We could just write...

## Prerequisites
* [NodeJS](https://nodejs.org) ([Installation instructions](https://github.com/moziot/wiki/wiki/Installing-node))
* [OpenZWave](http://www.openzwave.com/) ([Installation instructions](https://github.com/moziot/wiki/wiki/Installing-openzwave))

...with links to their websites and a wiki page with our tips for installing on different operating systems.

For example, on Ubuntu I downloaded a Node binary from [nodejs.org](https://nodejs.org/en/download/) and OpenZWave from [openzwave.com](http://old.openzwave.com/downloads/) and it worked fine. No need for pkg-config or nvm or building openzwave myself. On Mac OS the openzwave-shared documentation says you can use HomeBrew to install openzwave with "brew install open-zwave", although I haven't been able to get it to work with that method.

benfrancis:
Proposed an alternative here https://github.com/moziot/gateway/pull/23

Let me know if you think this is better.

I'm also wondering whether we can make some dependencies optional. If someone only wants a Bluetooth gateway or a ZigBee gateway or a ZWave gateway there are some dependencies they won't need (e.g. building openzwave which adds complexity).

dhylands:
I didn't want to just point at the pages, because they offer a bunch of alternatives, I wanted to document what we recommend.

I certainly don't mind having the instructions elsewhere (perhaps on the wiki or in a separate README file (like README-openzwave.md) and have the front page just refer to that.

I also think that having a script to install all of the right things is also pretty much the way to go. I think we're going to need it once we start doing Travis builds/testing.

For zwave in particular, building from source seems to be the recommended approach. The distros all tend to have older copies.

I'm all for making stuff optional. We just need to figure out how to make it work. I'm pretty sure that for something like zwave, we could make the adapter test to see if the libopenzwave.so file is available, and if not, don't enable the zwave adapter.

Currently it won't add the adapter if it doesn't find a dongle, but I think it might still need the lib.

dhylands:
I filed #24

benfrancis:
Yeah I think you're right this is only get more complicated so we're going to need a build script.

PR #25, opened 2017-04-05
Added addSomeDevice and removeSomeDevice

dhylands:
I decided to call it addSomeDevice and removeSomeDevice.

I added extra support to the adapter_controller so that you can exercise addSomeDevice, removeSomeDevice and get/set values.

### DIFF
--- a/adapter-manager.js
+++ b/adapter-manager.js
@@ -1,8 +1,9 @@
 /**
- * AdapterManager.
+ * Manages all of the Adapters used in the system.
  *
- * Manages Adapter data model and business logic.
- *
+ * @module AdapterManager
+ */
+/**
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
@@ -14,6 +15,11 @@ var fs = require('fs');
 var path = require('path');
 var EventEmitter = require('events').EventEmitter;
 
+/**
+ * @class AdapterManager
+ * @classdesc The AdapterManager will load any adapters from the 'adapters'
+ * directory. See loadAdapters() for details.
+ */
 class AdapterManager extends EventEmitter {
 
   constructor() {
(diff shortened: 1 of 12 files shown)


controllers/adapters_controller.js
```diff
@@ -0,0 +1,195 @@
...
+      adapterList.push(adapter.asDict());
+    }
+    response.json(adapterList);
+  });
+});
+
+/**
+ * Cancel Action: add a new device to the adapter.
+ */
+adaptersController.delete('/addSomeDevice', (request, response) => {
```
benfrancis: This isn't very RESTful! Resource names should be nouns, not verbs, and the DELETE here doesn't actually delete the resource.

I can think of a couple of ways of implementing this as a REST API...


```
PUT /pairing_mode
{
  "pairing": true
}
```
```
GET /things/new
```
```
PUT /pairing_mode
{
  "pairing": false
}
```
**or**

Request...
```
POST /actions
{
  "name": "addNewDevice"
}
```
Response...
```
201 Created
{
  "name": "addNewDevice"
  "href": "/actions/1"
}
```
```
GET /things/new
```
```
DELETE /actions/1
```

controllers/adapters_controller.js
```diff
@@ -0,0 +1,195 @@
...
+    response.status(404).send('Adapter "' + adapterId + '" device "' + deviceId + '" not found.');
+    return;
+  }
+  response.status(404).send('Adapter "' + adapterId + '" not found.');
+});
+
+/**
+ * Get a particular device associated with a particular adapter.
+ */
+adaptersController.get('/:adapterId/devices/:deviceId/:attributeId', (request, response) => {
```
benfrancis: This code is really useful, but I think it's in the wrong place. It should be in a ThingController. Can we leave this getting/setting attributes code out of this controller and maybe keep it somewhere so we can implement it as part of the ThingController?

/things/:thingId/properties/:propertyName

dhylands: I need this just to do my own testing/exercising of the code. I don't care what the top level is called (maybe change it from /adapters to /debug or /test).

ThingController is going to return items from the DB. The adaptersController has nothing to do with the database and goes directly to the adapters. The json it returns also doesn't necessarily conform to the thing description, it's just exposing the internals.

controllers/adapters_controller.js
```diff
@@ -0,0 +1,195 @@
...
+    response.json(adapter.asDict());
+    return;
+  }
+  response.status(404).send('Adapter "' + adapterId + '" not found.');
+});
+
+/**
+ * Get a list of devices associated with a particular adapter.
+ */
+adaptersController.get('/:adapterId/devices/', (request, response) => {
```
dhylands: Yeah - the purpose of this is to facilitate testing of the adapter level APIs and isn't necessarily intended for public consumption.

controllers/adapters_controller.js
benfrancis: This web API is going to be exposed to the Internet, I don't think it's an appropriate place to do integration testing. Can we find a better way? Or as a minimum have a way of turning it off.

controllers/adapters_controller.js
dhylands: Well the action in this case is to initiate pairing. Since there can only be one pairing request I chose not to bother implementing a queue and returned an href which is the same as the GET. So you do a GET on /adapters/addSomeDevice to initiate pairing and you do a DELETE on /adapters/addSomeDevice to cancel the pairing action.

I could fabricate some dynamic actions stuff but that seemed like an awful lot of "make work" just for a debug feature.

app.js
```diff
@@ -10,6 +10,11 @@
 var express = require('express');
 var db = require('./db');
 var app = express();
+var adapterManager = require('./adapter-manager.js');
+var bodyParser = require('body-parser');
```
dhylands: It parses the body as JSON, so that you can use request.body['led'] or whatever the attribute name is. We could parse the body ourselves.

adapter-manager.js
```diff
@@ -13,38 +14,245 @@
...
+  }
+
+  /**
+   * @method addSomeDevice
+   * Initiates pairing on all of the adapters that support it.
+   * The user then presses the "button" on the device to be added.
+   * @returns A promise that resolves to the newly added device.
+   */
+  addSomeDevice() {
+    var deferredAdd = new Deferred();
```
dhylands: That's actually what I'm doing. Deferred is a just a wrapper around Promise. An instance of deferred contains the promise, and the pointers to the resolve and reject callbacks, so that I can pass around a single object rather than 3.

Ahh - I think I probably forgot to checkin deferred.js. It just looks like this:
```
class Deferred {
  constructor() {
    this.promise = new Promise((resolve, reject) => {
      this.resolve = resolve;
      this.reject = reject;
    });
  }
}
```

adapter-manager.js
```diff
@@ -13,38 +14,245 @@
...
+    }
+  }
+
+  /**
+   * @method addSomeDevice
+   * Initiates pairing on all of the adapters that support it.
+   * The user then presses the "button" on the device to be added.
+   * @returns A promise that resolves to the newly added device.
+   */
+  addSomeDevice() {
```
dhylands: Yeah - I renamed addDevice to deviceAdded. Maybe I should call it _deviceAdded. It's not really part of the public facing API of the adapter, but rather a class private function which the particular adapters should call.

I didn't want to call it onDeviceAdded because it wasn't hooked up as a listener.

controllers/adapters_controller.js
dhylands: We could make it so that the adaptersController is renamed to debug/test and only shows up if a specific argument is passed when we start the app. That reinforces the fact that it's really just for test/debug.

adapter-manager.js
benfrancis: Oh I see! I've never seen that before... Have you tried this without the promise npm module installed?

app.js
benfrancis: OK, I was just curious. If you think it's worth adding a dependency then let's keep it.

controllers/adapters_controller.js
benfrancis: Oh, so this isn't the API that the front end will call to initiate pairing then? I guess we could implement that how I described...

controllers/adapters_controller.js
benfrancis: I think ThingController will also need to call methods on AdapterManager in order to get and set properties on devices (not just interact with the database), and we'll need a controller that can start and stop pairing and get a list of adapters. I assumed that was what this code was for.

But if you really think we need a REST API for integration testing then yes, maybe it should be called test/ or similar.

adapter-manager.js
dhylands: I tried this:
```
rm -rf node_modules
remove promise from package.json
npm install
```
and the app still seem to run fine.

Using Deferred is useful if you need to resolve/reject outside of the Promise constructor.

app.js
dhylands: It turns out that you need a bodyParser of some kind in versions of node > 4. They're no longer included by default.

controllers/adapters_controller.js
dhylands: ok - this PR didn't include any changes for the thing_controller

dhylands:
I've moved most of what was in adapters_controller.js into debug_controller.js and applied the other comments.

This PR wasn't intended to implement anything in the thing API, just the adapter-manager & friends. Once this part is merged, then we can implement more in the things_controller.

dhylands:
I've been using HTTPRequestor extension for testing: https://addons.mozilla.org/en-US/firefox/addon/httprequester/

benfrancis:
Thanks Dave!

PR #26, opened 2017-04-06
Adapter List and Main Menu

benfrancis:
I added an adapter view to view the list of installed adapters and a main menu to switch between views.


dhylands:
Looks good. When I run this, all of my screens are empty (i.e. no things, no adapters, no settings). Is that expected?

benfrancis:
@dhylands  Hmm, no. Only Settings should be empty. Do you get any errors in the console?

dhylands:
And I just realized, adapters are really just a special kind of thing. We should be able to expose properties for the adapters (which would probably be in the form of config settngs). But I could definitely see having properties (which would then become like config settings). Things like which port to use (could be auto, or could require manually assigning), debug settings (i.e. enable/disable debug output).

dhylands:
> @dhylands Hmm, no. Only Settings should be empty. Do you get any errors in the console?

No errors. This is my complete console output:
```
2020 >./app.js 
Listening on port 8080
Loading Adapters from ./adapters/foo.js
Adapter: FooAdapter id Foo1 created
Loading Adapters from ./adapters/zwave
initialising OpenZWave addon (/home/dhylands/Dropbox/moziot/gateway/node_modules/openzwave-shared/lib/../build/Release/openzwave_shared.node)
Found ZWave port @ /dev/ttyACM0
Adapter: ZWaveAdapter id ??? created
Initialising OpenZWave 1.4.2476 binary addon for Node.JS.
	OpenZWave Security API is ENABLED
	ZWave device db    : /usr/local/etc/openzwave
	User settings path : /home/dhylands/Dropbox/moziot/gateway/node_modules/openzwave-shared/build/Release/../../
	Option Overrides : --ConsoleOutput false
ZWave: Driver Ready: HomeId: cc9eff28
ZWave: node8 valueAdded: 37:Switch -> false
ZWave: node1: Ready Aeotec ZW090 Z-Stick Gen5
ZWave: node1: name="ZWaveAdapter", type="Static PC Controller", location=""
ZWave: node8: Ready Leviton Unknown: type=1805, id=0334
ZWave: node8: name="zwave-cc9eff28-8-Switch", type="Binary Scene Switch", location=""
ZWave: Scan complete
```

dhylands:
Do I need to remove the database? Or maybe I have a stale file lying around?

PR #29, opened 2017-04-06
Add Thing

benfrancis:
Gets a list of new things to add via websocket and lets the user save a thing with their own name.

May require clearing the database with 'rm db.sqlite3'.


controllers/actions_controller.js
```diff
@@ -0,0 +1,35 @@
...
+ActionsController.post('/', function (request, response) {
+  if (!request.body.name) {
+    response.status(400).send();
+    return;
+  }
+  switch(request.body.name) {
+    case 'pair':
+      console.log('Actions Controller requested addNewDevice');
+      AdapterManager.addNewDevice();
+      //TODO: Create an action resource in the database and return its ID
```
dhylands: I'm not sure that these should be persisted in the database.

If we need to persist something then we probably want persistable actions versus non-persistable actions. Pairing for both ZigBee and ZWave would be in the "non-persistable" category, and would just need an array/dict in the server app.

benfrancis: Agree, I realised this later too. The queue probably just needs to exist in memory. Unless we find there's some reason to keep a log of actions or persist action requests over a reboot, which so far I don't think we have a need for.

benfrancis:
@dhylands I think this is ready for review, sorry this got quite big!

* Gets a list of new devices via a WebSocket (new devices pushed down the socket as they are paired)
* Allows the user to save a Thing in the database with a user provided name

Known issues:
* The add device dialog will appear to keep scanning for new devices forever. We need to add some timeouts and close websockets when we're done.
* Currently assumes all devices are on/off switches!
* Ideally I think we should move some business logic from the controller to the model layer and logic from the model layer to the adapter layer, but this is getting big so I suggest doing that in a follow-ups.

May need to delete your sqlite database as I changed the database schema a bit.

dhylands:
I was able to add a device. Were you unpairing using the debug controller?

When I do Add and press the inclusion button, My add devices screen sometimes gets 2 entries added to the bottom which say "My On/Off Switch". I haven't figured out why this happens yet, just thought I'd point it out.

I think that the add devices screen should mention something about pressing the button on zwave/zigbee devices.

dhylands:
I think this needs to be rebased on top of the stuff I just did. Otherwise, I think we should check it in so we can move forward (even if it isn't perfect).

benfrancis:
OK, let's get this landed, I need to build on top of it.

PR #31, opened 2017-04-07
Ran jshint on all files and cleaned up warnings

dhylands:
I added the .jshinrc file from gaia and added node: true to the end.


benfrancis:
Thanks

PR #33, opened 2017-04-07
Start and stop pairing from the UI

benfrancis:
Here's just the start and stop pairing parts, for review

### DIFF
--- a/controllers/actions_controller.js
+++ b/controllers/actions_controller.js
@@ -1,0 +1,74 @@
+/**
+ * Actions Controller.
+ *
+ * Manages the top level actions queue for the gateway (i.e. not actions on
+ * Things but on the gateway itself.)
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+'use strict';
+
+/* jshint unused:false */
+/* globals AdapterManager */
+
+var express = require('express');
+var Things = require('../models/things');
+var AdapterManager = require('../adapter-manager');
+
+var ActionsController = express.Router();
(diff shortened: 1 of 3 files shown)


controllers/actions_controller.js
```diff
@@ -0,0 +1,70 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+var express = require('express');
+var Things = require('../models/things');
+var AdapterManager = require('../adapter-manager');
+
+var ActionsController = express.Router();
+ActionsController.actions = [];
```
dhylands: And after Adding and cancelling, doing a GET on /actions returned [null]

benfrancis:
> I tested with a real ZWave device and after adding the device the "Scanning" page stayed up and I had to Cancel it.

Yep this is expected. I was hoping to implement the add device part at the same time as starting/stopping pairing but it's proving complex so I thought about getting this landed first. It just starts pairing when click the add button and cancels pairing when you click the cancel/back button.

> And after Adding and cancelling, doing a GET on /actions returned [null]

I'd expect it to be a an array of null entries where the action entries used to be, because cancelling pairing deletes the action entry from the array. But yes a dict would probably be better.

> We also need a way to delete the action when it's done via a success route.

I wasn't sure whether we want to keep a log of completed actions for any reason. Currently an action is only deleted if it is cancelled, but I realise this is a bit inconsistent.

Currently working on the changes you proposed, thanks.

benfrancis:
OK, I've addressed your comments. I still haven't quite figured out to do once a new device is returned so planning to address that in a follow-up if that sounds reasonable to you.

dhylands:
Waiting for a followup sounds fine.

PR #39, opened 2017-04-21
Change attributes to properties

dhylands:
This changes attributes to properties.

Each property (within a device) needs to have a unique id (assigned by the adapter).

I'm sure that more refinements will be required, but I figured I'd put these changes out as a start.


dhylands:
I found a couple spots I missed. I also decided that this was a good time to remove test-server.js since it won't work anymore after these changes.

dhylands:
ok - I've redone this in what I think is the proper way. I created a zwave-classifier.js which provides the mappings from properties to zwave values and figures out the wot type by looking at the command classes.

I've updated the debug controller and I can now turn on/off lights using /debug/device/DEVICEID/on

I started to add support for actions, but there isn't any handler right now.

I also fixed the problem with the name at add time not being filled in.

I haven't run into any segfaults so it might be specific to the Mac. My new Mac laptop should be showing up next week, and I'll see if I can reproduce anything.

benfrancis:
What's here looks good to me, but did you miss off zwave-classifier.js ?

FWIW the segfaults I experienced were on Linux, and using the precompiled openzwave binary, haven't seen them on the Mac.

dhylands:
Looks like I commited to my local repository and forgot to push. So the big changeset should be there now

dhylands:
I think that having getThing would probably be the right thing to do. Then we can expose getDevice using the debug controller, which has useful information.

For consistency, we should probably call it addNewThing and removeSomeThing. I don't think that I need the addNewDevice/removeSomeDevice APIs, so I'll rename them and change the context (i.e. promise will resolve to a thing)

I wasn't intending to set ConsoleOutput to true by default. I'll fix it. There is a way in OpenZWave to have a config file which controls this stuff, and these are really just overrides. For now, I'll just turn it off and defer figuring out the config file until after the demo.

I'll add the adapterManager APIs for getting/setting the properties.

dhylands:
Renamed addNewDevice/removeSomeDevice to addNewThing/removeSomeThing.
Added thing enumeration/property get/set to the debug controller.
Added some debug to Deferred (promises for adding/removing).
Added more functions to the AdapterManager
  - getThings, getThing, getProperty, getProperties,
    getPropertyValue, and setPropertyValue
addNewThing/removeSomeThing now resolve things rather than devices.
rebased to the latest master.

I tweaked models/thing.js and controllers/action_controller.js

deferred.js
```diff
@@ -7,12 +7,34 @@
...
+
+var id = 0;
+
 class Deferred {
   constructor() {
+    this.id = ++id;
     this.promise = new Promise((resolve, reject) => {
-      this.resolve = resolve;
-      this.reject = reject;
+      this.resolve_func = resolve;
```
benfrancis: nit: camelCase

adapter-manager.js
```diff
@@ -145,24 +145,97 @@ class AdapterManager extends EventEmitter {
...
+    if (device) {
+      return device.getProperty(propertyName);
+    }
+  }
+
+  /**
+   * @method getPropertyValue
+   * @returns Retrieves the value of the property named `propertyName` from
+   *          the thing identified by `thingId`.
+   */
```
benfrancis: Curious why we need both getProperty() and getPropertyValue() and why we have setPropertyValue() as opposed to just setProperty()?

deferred.js
dhylands: will fix

benfrancis:
Another note, I was expecting getProperty[value]() and setProperty[value]() to be async and return a promise which resolves when the value has been read or written successfully. Sync is fine to start with, but it would be nice to have a success/failure state for setting properties in particular (I'm guessing the get just reads from a cache?).

adapter-manager.js
dhylands: getProperty returns the property object of the thing (name, type, description). getPropertyValue returns the value of the object that property represents. i.e. that's what actually turn the light on/off.

dhylands:
That should be doable. The get would resolve/reject immediately for the devices we have.
We could have the set defer until the valueChanged event occurs.

Lets leave this sync for now and I'll open an issue so we can change it after the demo.

dhylands:
I filed issue #42 to have get/setPropertyValue return a promise

adapter-manager.js
benfrancis: Ah I see. So in Web Thing terminology getProperty() actually just gets the property description from the Thing resource whereas getPropertyValue() gets the value from the Property resource.

Note that in the current [Web Thing spec](https://moziot.github.io/wot/#property-resource) getting a Property resource returns JSON which looks like the following:

```
{
  "on": true
}
```

This is subject to change of course, if we come up with something better, but just FYI this is what I'm currently implementing on top of this code.

benfrancis:
> I filed issue #42 to have get/setPropertyValue return a promise

Thanks!

adapter-manager.js
dhylands: ok. so perhaps what I've called getProperty should really be called getPropertyDescription?
And getPropertyValue should be called getProperty?

I don't have a problem with having getPropertyValue return JSON instead of a plain return value (too much C background on my part :), and it probably makes sense anyways if there was something more complex than a simple value.

dhylands: I think we should do this renaming sooner rather than later. I'll get a PR together to do it today. I'll also do the json return as well.

benfrancis: Yep that all sounds good. I'm currently just working around it, so not a huge problem.

PR #44, opened 2017-04-26
Can turn a ZigBee light on and off.

dhylands:
This adds the initial zigbee adapter. It doesn't yet add devices, but it does scan them.


dhylands:
Right now, I think that the default pairing is enabled, so you can pair nodes with no UI interaction required.

I think that there needs to be some refactoring, and I think the whole queue can be redone using promises, but I figured I'd check it in now since it worked once for me !

PR #56, opened 2017-04-28
Improve the initial name for ZigBee

dhylands:
This also filters the dongle from the list of devices returned to
the adapter manager.

And I fixed the ZigBee on/off. It turns out it was broadcasting
the on/off to all switches.


dhylands:
r=me

I tested everything locally, so I'm just going to go ahead and land this for the demo.

PR #59, opened 2017-05-02
Add a couple of utilities

dhylands:
- list_port.js is a nodeJS program which will list all of the
  detected serial ports and some information about them.
- make-self-signed-cert.sh will create a self-signed certificate
  which will allow the gateway to be used on a local network
  without having to use the tunnel.


dhylands:
r=me

PR #60, opened 2017-05-04
ZigBee improvements

dhylands:
- Did some refactoring (still lots more to do)
- Now queries the initial value of the on/off switch
- Now updates the cached value when the manual switch is pressed on the device


benfrancis:
We really need some tests!

PR #61, opened 2017-05-05
Improve zwave

dhylands:
This needs #60.

 - Now picks up the correct initial state
 - Added support for the Aeotec SmartSwitch 6
   - By default, pressing the button doesn't report the change
   - detect the SmartSwitch and adjust the configuration to make it
     report button presses

### DIFF
--- a/adapters/zwave/zwave-classifier.js
+++ b/adapters/zwave/zwave-classifier.js
@@ -13,11 +13,12 @@
 // See; http://wiki.micasaverde.com/index.php/ZWave_Command_Classes for a
 // complete list of command classes.
 
-const COMMAND_CLASS_SWITCH_BINARY = 37;
-const COMMAND_CLASS_SWITCH_MULTILEVEL = 38;
-const COMMAND_CLASS_SWITCH_ALL = 39;
+const COMMAND_CLASS_SWITCH_BINARY = 37;     // 0x25
+//const COMMAND_CLASS_SWITCH_MULTILEVEL = 38; // 0x26
+//const COMMAND_CLASS_SWITCH_ALL = 39;        // 0x27
+const COMMAND_CLASS_CONFIGURATION = 112;    // 0x98
 
-const THING_TYPE_ON_OFF_SWITCH = "onOffSwitch";
+const THING_TYPE_ON_OFF_SWITCH = 'onOffSwitch';
 
 class ZWaveClassifier {
 
@@ -25,8 +26,8 @@ class ZWaveClassifier {
     }
 
     classify(node) {
-        for (var valueId in node.values) {
-            var value = node.values[valueId];
+        for (var valueId in node.zwValues) {
+            var value = node.zwValues[valueId];
 
             if (value.class_id == COMMAND_CLASS_SWITCH_BINARY) {
                 // This looks like an on/off switch
@@ -37,7 +38,7 @@ class ZWaveClassifier {
     }
 
     initOnOffSwitch(node, valueId) {
-        node.type = "onOffSwitch";
+        node.type = THING_TYPE_ON_OFF_SWITCH;
         node.properties = [{
             'name': 'on',
             'type': 'boolean',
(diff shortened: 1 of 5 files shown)


adapters/zwave/zwave-classifier.js
```diff
@@ -13,20 +13,21 @@
 // See; http://wiki.micasaverde.com/index.php/ZWave_Command_Classes for a
 // complete list of command classes.
 
-const COMMAND_CLASS_SWITCH_BINARY = 37;
-const COMMAND_CLASS_SWITCH_MULTILEVEL = 38;
-const COMMAND_CLASS_SWITCH_ALL = 39;
+const COMMAND_CLASS_SWITCH_BINARY = 37;     // 0x25
+//const COMMAND_CLASS_SWITCH_MULTILEVEL = 38; // 0x26
+//const COMMAND_CLASS_SWITCH_ALL = 39;        // 0x27
```
benfrancis: Did you mean to leave these commented out?

adapters/zwave/zwave-classifier.js
```diff
@@ -46,7 +47,28 @@ class ZWaveClassifier {
             'name': 'toggle',
             'description': 'Toggles On/Off',
         }];
-        node.propertyMap = {'on': valueId };
+        node.propertyMaps = {'on': valueId };
+        node.values.on = node.zwValues[valueId].value;
+
+        if (node.zwInfo.productId == 0x0060 &&
+            (node.zwInfo.productType & 0xff) == 0x03) {
```
benfrancis: This is ringing some alarm bells for me. Do we have to write code like this for every model of Z-Wave smart plug on the market? Is there a more generic way to do this that might work for multiple smart plugs?

adapters/zwave/zwave-classifier.js
dhylands: I commented them out because the linter was complaining that they're not being used. Which is true for now. They will be used soon as we add more support for more types of things.

adapters/zwave/zwave-classifier.js
dhylands: Not that I'm aware of. The reporting behaviour of the switch is controlled by a config which defaults to "Don't report anything when the switch is pressed", which allows the state of the light to become out of sync.

I won't know how prevelant this will be until we experiment with different brands of modules. My Leviton switches, for example, don't change the light and don't report anything when the button is pressed (so at least its consistent).

From what research I did, there is no standard for having config options, and what config options exist seem to be entirely vendor specific.

So, we need someway of specifying this in order to get "sensible" behaviour. Obviously, it doesn't need to exist buried in the code like this. You could have some type of external file which has quirks. For that matter, you could probably even have thing mappings be described purely using data.

benfrancis: I'm grateful that you figured out how to make this work, but I really don't think we should bury vendor-specific hacks in the code like this. I'd suggest we either need generic code which can provide a sensible default configuration for all Z-Wave switches (preferable, but maybe not possible), or a mechanism to list vendor-specific default configurations (maybe in some declarative form).

I think it makes sense to have a classifier which maps standardized Z-Wave device types (groups of command classes) onto Web Thing types, but if this kind of vendor-specific configuration is needed to make Z-Wave smart plugs sensibly then it's likely people are going to want to extend this with their own device configurations.

I'd suggest our focus should be on designing a some kind of add-on mechanism for adapters (and specific device configurations if necessary), rather than hard coding these quirks into the classifier.

dhylands: I have no issues with (and agree with) having some type of generic mechanism. Do you want me to remove this? Or leave it in until the generic mechanism is implemented?

dhylands: After implementing this a few times (I now have zigbee, zwave, and gpio onOffThings, I think that we need to factor some of this out, but I'll leave that for later.

benfrancis: Yes my suggestion would be to remove vendor-specific hacks from the classifier and try to stick to mapping standardized classes, until we have a sensible mechanism for catering to those quirks.

dhylands:
OK - I removed the vendor specific code, but I personally would have left it in and removed it when we had the mechanism, since now you need to be very careful at any demo since the UI state and the switch state can easily get out of sync.

PR #62, opened 2017-05-06
Added adapter for GPIO output

dhylands:
This adds an adapter for output GPIOs. To use, copuy the config/gpio-config-example.js to config/gpio-config.js and edit appropriately.

Example of it running can be found here:
https://www.youtube.com/watch?v=sODPG_wM4jY

### DIFF
--- a/adapters/gpio/gpio-adapter.js
+++ b/adapters/gpio/gpio-adapter.js
@@ -1,0 +1,118 @@
+/**
+ *
+ * GpioAdapter - an adapter for controlling GPIO pins.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+'use strict';
+
+var fs = require('fs');
+var Gpio = require('onoff').Gpio;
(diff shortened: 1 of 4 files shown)


dhylands:
The configuration file is just an example file, which is why its called gpio-config-example.js. It needs to be called gpio-config.js to be actually used (and we shouldn't ever checkin gpio-config.js). I put it in tree just to keep things simple for now. I tested under linux and OSX and it just doesn't load itself if the resource isn't accessible. I figured it would be easier to have it in tree for time being. Once our APIs have more or less stabilized, then we can look at moving it out of tree.

dhylands:
Perhaps we need an adapters-config.js which lists the adapters to load and their locations. This would allow various adapters to be used from separate repositories without having to copy them to the adapters directory.

benfrancis:
> Perhaps we need an adapters-config.js which lists the adapters to load and their locations. This would allow various adapters to be used from separate repositories without having to copy them to the adapters directory.

I'm thinking of something much more dynamic than that. Like an Add-ons system in the UI where you can download adapters (and maybe new device types) from an add-ons directory.

The add-ons mechanism could be via node modules, zip files (like Wordpress plugins), apt packages, snap packages... lots to discuss.

Anyway, I guess we can land this in tree until we have that mechanism figured out and have decided which adapters should ship by default.

PR #63, opened 2017-05-07
Add a test program for testing the REST API.

dhylands:
This adds a REST client which can be used to exercise the API. Some examples of things that can be done (I had the GPIO adapter enabled with no ZWave or ZigBee dongle).
```
470 >./test.js https://rpi3.local:4443 --list
URL             Name Type        Properties
=============== ==== =========== ==========
/things/gpio-18 led  onOffSwitch on        

dhylands@dylands-38202 ~/Dropbox/moziot/gateway/test (add-test)
471 >./test.js https://rpi3.local:4443 --list --toggle
URL             Name Type        Properties
=============== ==== =========== ==========
/things/gpio-18 led  onOffSwitch on        

led property on has value false
led setting property on to true
led property on has value true
led setting property on to false
dhylands@dylands-38202 ~/Dropbox/moziot/gateway/test (add-test)
472 >./test.js https://rpi3.local:4443 --list --thing=led
URL             Name Type        Properties
=============== ==== =========== ==========
/things/gpio-18 led  onOffSwitch on        

dhylands@dylands-38202 ~/Dropbox/moziot/gateway/test (add-test)
473 >./test.js https://rpi3.local:4443 --list --thing=foo
No things found
dhylands@dylands-38202 ~/Dropbox/moziot/gateway/test (add-test)
474 >./test.js https://rpi3.local:4443 --list --thing=led --set on=true
URL             Name Type        Properties
=============== ==== =========== ==========
/things/gpio-18 led  onOffSwitch on        

led setting property on to true
dhylands@dylands-38202 ~/Dropbox/moziot/gateway/test (add-test)
475 >./test.js https://rpi3.local:4443 --list --thing=led --set on=false
URL             Name Type        Properties
=============== ==== =========== ==========
/things/gpio-18 led  onOffSwitch on        

led setting property on to false
```
`--list` causes all of the matching things to be printed.
`--toggle` toggles all matching things, waits a second and then toggles them again.
`--set on=true` Turns all matching things on.
`--set on=false` Turns all matchining thing off.
`--thing foo` Filters the things so that only things with 'foo' in the name match.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,8 @@
     "xbee-api": "^0.5.2",
     "zcl-id": "^0.3.2",
     "zcl-packet": "^0.2.2"
+  },
+  "devDependencies": {
+    "node-rest-client": "^3.1.0"
   }
 }
--- a/test/gateway.js
+++ b/test/gateway.js
@@ -1,0 +1,84 @@
+#!/usr/bin/env node
+/*
+ * MozIoT Gateway App.
+ *
+ * Back end main script.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+
+var Thing = require('./thing');
+var Client = require('node-rest-client').Client;
+
+var DEBUG = false;
+
+class Gateway {
+  constructor(url) {
+    if (url.slice(-1) === '/') {
+      // Remove trailing /
+      url = url.slice(0, -1);
+    }
+    this.server_url = url;
+    if (DEBUG) {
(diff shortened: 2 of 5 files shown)


dhylands:
It's pretty standalone, so it could live in its own tree. It does use a few of the utility files, so we'd have to duplicate those if we put it in its own repo. i'm happy either way.

I was intending to use it to verify that the basic gateway API still works. For example, being able to toggle all of the ZigBee and Zwave devices you can be reasonably confident I didn't break anything.

benfrancis:
Ah OK, then we should start to investigate automated regression testing. I'm sure there are existing tools out there for testing REST APIs.

benfrancis:
Alright let's land this until we have automated testing figured out. I'm just concerned about cluttering the repo with random scripts.

PR #70, opened 2017-05-10
Action model

benfrancis:
This creates a model layer for Actions and moves business logic out of the controller and into the model where it should be.

Note: This is top level actions on the gateway, not actions on Things which are part of the Web Thing API and will come later.

### DIFF
--- a/models/action.js
+++ b/models/action.js
@@ -1,0 +1,29 @@
+/**
+ * Action Model.
+ *
+ * Manages Action data model and business logic.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+
+var Actions = require('../models/actions');
+var config = require('../config');
+
+var Action = function(name) {
+  this.id = Actions.generateId();
+  this.name = name;
+  this.href = config.ACTIONS_PATH + '/' + this.id;
+};
+
+Action.prototype.getDescription = function() {
+  return {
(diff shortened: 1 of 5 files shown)


models/action.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+'use strict';
+
+var Actions = require('../models/actions');
+
+var Action = function(name) {
+  this.id = Actions.generateId();
+  this.name = name;
+  this.href = '/actions/' + this.id;
```
benfrancis: Yep, good idea! This shouldn't be hard coded in the model.

models/actions.js
```diff
@@ -0,0 +1,88 @@
...
+   */
+  add: function(action) {
+    var id = action.id;
+    this.actions[id] = action;
+    switch(action.name) {
+      case 'pair':
+        AdapterManager.addNewThing().then(function(thing) {
+          Things.handleNewThing(thing);
+        }).catch(function(error) {
+          console.error('Error trying to add new thing ' + error);
```
benfrancis: Good point.

config.js
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+'use strict';
+
+exports.THINGS_PATH = '/things';
+exports.NEW_THINGS_PATH = '/new_things';
+exports.ADAPTERS_PATH = '/adapters';
+exports.ACTIONS_PATH = '/actions';
+exports.STATIC_PATH = 'static';
+exports.DEBUG_PATH = '/debug';
```
dhylands: There is also a config directory. Currently, it just has the gpio config. Perhaps this should be a file in the config directory, and have index.js

benfrancis:
Done, thanks.

PR #73, opened 2017-05-12
Thing model

benfrancis:
This builds on #70 to create a Thing model (the current thing.js was unused).

Rather than blindly passing around JSON data, the Thing model actually parses a Thing Description and creates a Thing object from it. You can then ask for a valid Thing Description with thing.getDescription().

The Things model internally keeps track of a Map of Things by Thing ID and can be asked for a list of Thing Descriptions to return in the API.

This will make the handling of Thing Descriptions more robust and in future the Thing object can be extended with methods so that the model is not just a container for data.

### DIFF
--- a/models/things.js
+++ b/models/things.js
@@ -10,15 +10,17 @@
 
 'use strict';
 
+var Thing = require('./thing.js');
 var Database = require('../db.js');
-var adapterManager = require('../adapter-manager');
+var AdapterManager = require('../adapter-manager');
+var Config = require('../config');
 
 var Things = {
 
   /**
    * A Map of Things in the Things database.
    */
-  things: {},
+  things: new Map(),
 
    /**
     * A collection of open websockets listening for new things.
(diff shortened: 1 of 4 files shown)


models/things.js
```diff
@@ -28,21 +30,38 @@ var Things = {
...
     *
-    * @return Promise which resolves with a list of Thing objects.
+    * @return {Promise} which resolves with a Map of Thing objects.
     */
    getThings: function() {
      return new Promise((function(resolve, reject) {
        Database.getThings().then((function(things) {
          // Update the map of Things
-         this.things = {};
+         this.things = new Map();
```
dhylands: Interesting. I didn't realize that there was a Map and what the differences were. From my reading it looks like I should be using Map in several places in the adapter where I currently use an object {}

benfrancis: @dhylands  Yes I wasn't sure whether NodeJS supported Map because it's a relatively recent feature of JavaScript and doesn't have universal support in browsers. They have both advantages and disadvantages though.

PR #85, opened 2017-05-23
Add support for removing devices

dhylands:
This adds the core functionality for removing a device.


dhylands:
Unpairing and pairing now both work for zigbee.

benfrancis:
Thanks!

PR #87, opened 2017-05-30
Changes properties to be a dictionary (object) rather than an array.

### LINKED ISSUE #86: Web Thing Description - Properties, actions and events as objects
Change properties, actions and events members of the Web Thing Description from arrays to objects, as specified in the latest Web Thing API draft https://moziot.github.io/wot/

dhylands:
Fixes #86 
This PR assumes that #85 will land first.

### DIFF
--- a/models/things.js
+++ b/models/things.js
@@ -79,10 +79,11 @@ var Things = {
            if(!storedThings.has(connectedThing.id)) {
              connectedThing.href = Config.THINGS_PATH + '/' + connectedThing.id;
              if (connectedThing.properties) {
-               connectedThing.properties.forEach(function(property) {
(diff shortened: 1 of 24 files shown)


models/things.js
```diff
@@ -74,15 +74,17 @@ var Things = {
        this.getThings().then((function(storedThings) {
          // Get a list of things connected to adapters
          var connectedThings = AdapterManager.getThings();
+         console.log('connectedThings =', connectedThings);
```
benfrancis: Remove this logging?

dhylands: Whoops - I thought I got them all - sigh

PR #88, opened 2017-05-30
Add jshint-all.sh script and run it on the repository.

dhylands:
Question: Should be do this as part of Travis?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: node_js
+
+sudo: required
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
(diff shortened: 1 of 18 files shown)


benfrancis:
Yes, I think this could be part of our automated testing/continuous integration.

For now, should this script live in tools/ or tests/ ?

dhylands:
If we're going to make it part of CI, then I'd be inclined to move it into tests. We can hook it into package.json so that you do:
```
npm run jshnt
```
and that will run the jshint-all.sh script. We'd also hook it into:
```
npm test
```
so that it would run on Travis. I'll update this PR to include the travis bits.

benfrancis:
Nice!

PR #90, opened 2017-06-01
Remove Thing - closes #4

### LINKED ISSUE #4: Remove Thing
As a user I want to disconnect a Thing from the gateway when I no longer need it

benfrancis:
UI and API to remove a Thing.

Note that this just removes the Thing from the database so for ZigBee and Z-Wave devices the Thing will return to the new things list so it can be re-added. This doesn't cover unpairing the device which may be treated as a separate interaction, I think we need a bit more thought on how the UI for that should work.

### DIFF
--- a/static/js/on-off-switch.js
+++ b/static/js/on-off-switch.js
@@ -43,7 +43,7 @@ OnOffSwitch.prototype.view = function() {
 /**
  * Update the on/off status of the on/off switch.
  */
-Thing.prototype.updateStatus = function() {
+OnOffSwitch.prototype.updateStatus = function() {
   if (!this.onPropertyUrl) {
     return;
   }
@@ -66,7 +66,7 @@ Thing.prototype.updateStatus = function() {
 /**
  * Show on state.
  */
-Thing.prototype.showOn = function() {
+OnOffSwitch.prototype.showOn = function() {
   this.element.classList.remove('off');
   this.element.classList.add('on');
 };
@@ -74,7 +74,7 @@ Thing.prototype.showOn = function() {
 /**
  * Show off state.
  */
-Thing.prototype.showOff = function() {
+OnOffSwitch.prototype.showOff = function() {
   this.element.classList.remove('on');
   this.element.classList.add('off');
 };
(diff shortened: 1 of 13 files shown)


static/js/on-off-switch.js
```diff
@@ -102,37 +104,37 @@ Thing.prototype.handleClick = function(e) {
...
-  }).bind(this))
-  .catch(function(error) {
-    console.error('Error trying to turn on switch: ' + error);
-  });
+OnOffSwitch.prototype.turnOn = function() {
+ this.showTransition();
+ this.properties.on = null;
+ var payload = {
+   'on': true
+ };
```
dhylands: nit: Looks like only 1 character indention. Was 2 intended?

dhylands:
Apparently I fail at filling out the review thing. I meant to add these comments and they didn't show up:

The Remove seems a bit "in your face" on the desktop, but I'm ok to leave that until we get a UX person :)

I tried on m my phone, and there seems to be a layout issue. See: https://people-mozilla.org/~dhylands/gateway/pr-90/Screenshot_20170601-113709.png The name of the devide being removed is almost completely obscured by the Remove button.

Unrelated to this PR, the layout of the main screen also has some issues on my phone:
https://people-mozilla.org/~dhylands/gateway/pr-90/Screenshot_20170601-113808.png

So the "change I was requesting" was to fix the layout on the remove screen to not obscure the name of the thing being removed.

benfrancis:
Good catch. That layout bug doesn't occur in Firefox.

Shame on you for using Chrome, and shame on me for not testing in Chrome... ;)

Sorry about my laziness in the design, I only designed the Things screen to show a maximum of two on/off switches for the purposes of the demo, and didn't bother making the context menu more responsive for desktop because I think it's going to be a temporary solution. But maybe I should be less lazy :)

dhylands:
> Shame on you for using Chrome

That's a long story. I used to use Firefox mobile, but several of the websites I use on a daily basis caused Firefox to crash routinely. I got tired of having 3 or 4 crashes while trying to read a news article. So I switched to Chrome. I guess I should go back and see if they ever did fix that bug. I filed a bug report along with a ton of crash reports. The particular issue was related to scrolling and had been open for quite a few months with no movement, which also motivated the move to Chrome.

benfrancis:
Created follow up #92 to fix the unrelated layout bug

benfrancis:
Addressed review comments

PR #91, opened 2017-06-02
Add a test adapter and tests for thing_controller.js

dhylands:
This should test all of the code paths in thing_controller.js

I got rid of the old files from test.js (test.js, gateway.js, and thing.js) and added a new test.js which has mocha tests in it.


adapters/test-adapter/test-adapter.js
```diff
@@ -0,0 +1,155 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+'use strict';
+
+var Adapter = require('../adapter');
+var Device = require('../device');
+var Property = require('../property');
+
+class TestProperty extends Property {
```
benfrancis: Would it be accurate to refer to the TestAdapter as a MockAdapter?

MockProperty
MockDevice
MockAdapter

tools/make-self-signed-cert.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -x
+
 openssl genrsa -out privatekey.pem 2048
-openssl req -new -sha256 -key privatekey.pem -out csr.pem
+openssl req -new -sha256 -key privatekey.pem -out csr.pem -subj '/CN=www.toizom.com/O=MozIoT Gateway/C=US'
```
dhylands: This is just to generate the certificate without requiring user interaction. I just picked a name. It seems to work fine. I tested it locally, and it seems to still just treat it as self-signed. I think that any data would be fine.

adapters/test-adapter/test-adapter.js
dhylands: Yeah - I could go for that

adapter-manager.js
```diff
@@ -306,13 +306,25 @@ class AdapterManager extends EventEmitter {
     for (var adapterName in config.adapters) {
       var adapterConfig = config.adapters[adapterName];
 
-      if (adapterConfig.enabled) {
+      var loadAdapter = adapterConfig.enabled;
+      var notLoadReason = 'disabled';
+      if (process.env.NODE_ENV === 'test') {
```
dhylands: I like the idea of having a testing config versus a regular config. I'll take a stab at consolidating the config into a single file. There is a bunch of stuff in the config that isn't really config, but rather constants (like /things, etc), so I think I'll split those out of config too,

app.js
```diff
@@ -63,7 +63,16 @@ app.use(bodyParser.urlencoded({ extended: false }));
...
-db.open();
+var dbName = './db.sqlite3';
+if (process.env.NODE_ENV === 'test') {
+  // Use a different database for testing.
+  dbName = './test-db.sqlite3';
+  // Always start out with no database (for testing)
+  if (fs.existsSync(dbName)) {
+    fs.unlinkSync(dbName);
+  }
+}
```
dhylands: I think that its possible to do this in the test.js file. I just need to do it before the require('../app') line.

config/index.js
```diff
@@ -17,6 +17,7 @@ exports.ADAPTERS_PATH = '/adapters';
 exports.ACTIONS_PATH = '/actions';
 exports.STATIC_PATH = 'static';
 exports.DEBUG_PATH = '/debug';
+exports.TEST_PATH = '/test';
```
dhylands: I originally had a test_controller assuming that the tests would be running out-of-process, but the tests run in-process so I removed the test_controller and missed removing this.

test/test.js
```diff
@@ -1,131 +1,165 @@
...
+    chai.request(server)
+      .get(config.THINGS_PATH + '/test-1/properties/on')
+      .end((err, res) => {
+        res.should.have.status(200);
+        res.body.should.have.a.property('on');
+        res.body.on.should.be.eql(true);
+        done();
+      });
+  });
+});
```
dhylands: I personally have always preferred hypens because you don't need to press the SHIFT key.

test/test.js
```diff
@@ -1,131 +1,165 @@
...
+let chaiHttp = require('chai-http');
+let server = require('../app');
 
-// Get some decent error messages for unhandled rejections. This is
-// often just errors in the code.
-process.on('unhandledRejection', (reason) => {
-  console.log('Unhandled Rejection');
-  console.error(reason);
-});
+let config = require('../config');
```
dhylands: It looks like if we use the config module: https://github.com/lorenwest/node-config then it allows for default configuration settings and overrides for production/development/test based on the NODE_ENV environment variable.

So I'm going to suggest that we switch to that.

dhylands:
I think that I addressed all of the comments.

benfrancis:
This is great, thanks. I'm getting a bit weary of all the npm module dependencies we have, but this config module does seem very helpful!

PR #96, opened 2017-06-09
Login

benfrancis:
Adds user authentication against an email address and password stored in the database.

Authentication is currently disabled by default because it relies on a default user being configured and a hard coded session secret. These defaults will be removed when first time setup (#9) is implemented (in #97) so that the user can specify their own username and password and a unique secret is generated for each gateway on first use.

To enable authentication to try it out just change authentication.enabled to true in config/default.js. You'll know it's working because you'll be redirected to the login page when you try to load the front end and will need to use username "user@example.com", password "password" to log in.

Authentication uses Passport.js and authenticates user access to the front end. Authentication of API calls is not included, this will be implemented in a follow up (#102).

I've added an integration test for the only API change (get user), but testing is still weak elsewhere (need to figure out how to test front end pieces).

index.html and login.html live in a new "views" folder. They're still just static HTML (no need for a template engine yet) but this assures they are authenticated and can't be accessed via the static file controller.

authentication.js handles configuring the authentication system, defining the authentication strategy and authenticating individual requests.

I've moved the router configuration into router.js because app.js was getting a bit big.

This is quite a big pull request, it took me a while to figure out how to use passport.js and the feature will make more sense once we have first time setup, so let me know if you have any questions.


config/default.js
```diff
@@ -37,4 +37,13 @@ module.exports = {
...
+  authentication: {
+    enabled: false,
+    defaultUser:  {              // DO NOT USE THIS IN PRODUCTION
+      email: 'user@example.com',
+      password: 'password',
+      name: 'Example User'
+    },
+    secret: 'top secret 51'      // DO NOT USE THIS IN PRODUCTION
+  }
 };
```
dhylands: By default, config assumes NODE_ENV is set to development, so we could put this in development.js and put a different set of defaults in production.js (and then we would arrange for NODE_ENV to be set to production for a real install)

config/test.js
```diff
@@ -25,4 +25,12 @@ module.exports = {
     filename: './test-db.sqlite3',
     removeBeforeOpen: true,
   },
+  authentication: {
+    enabled: false,
+    defaultUser:  {              // DO NOT USE THIS IN PRODUCTION
```
dhylands: Since this is in test.js it will only ever get used when NODE_ENV is set to test.

views/login.html
```diff
@@ -0,0 +1,23 @@
...
+  </head>
+  <body>
+    <img id="wordmark" src="../images/moziot-wordmark.png" />
+    <form id="login-form" action="/login" method="post">
+      <input type="email" name="email" placeholder="Email"></input>
+      <input type="password" name="password" placeholder="Password"></input>
+      <button id="login-button" type="submit">Log in</button>
+    </form>
+  </body>
+</html>
```
dhylands: It would be nice if this had a "Remember Me" checkbox, so that I don't have to re-enter the username and password everytime (probably best to add this as a separate PR)

config/test.js
benfrancis: That was intentional :)

config/default.js
benfrancis: At the moment I think both would be the same anyway. These values have to be provided but we want authentication disabled by default. The follow up bugs will implement the features which allow these to be set dynamically, at which point it makes sense to remove this from the default config and a separate development.js and production.js is a great idea.

views/login.html
benfrancis: I would like that too. I would also like a forgotten password feature! Will file bugs.

PR #103, opened 2017-06-16
/new_things improvements

hobinjk:
This fixes several issues with the behavior of /new_things. First, it updates `handleNewThing` to send the Thing instead of a description of the thing. This fixes a case where pairing a new device would be impossible due to the description's lack of its corresponding thing's id.

Second, it replaces the one-shot AdapterManager promise with a listener on 'thing-added' events. This is explained in more detail at https://github.com/moziot/gateway/pull/99#issuecomment-309088070 and #99 in general.

It finally adds a test for the updated behavior which will fail if either fix is not present.

### DIFF
--- a/models/things.js
+++ b/models/things.js
@@ -112,11 +112,9 @@ var Things = {
     * @param Object New Thing description
     */
    handleNewThing: function(newThing) {
-     var thing = new Thing(newThing.id, newThing);
-
      // Notify each open websocket of the new Thing
      this.websockets.forEach(function(socket) {
-       socket.send(JSON.stringify(thing.getDescription()));
+       socket.send(JSON.stringify(newThing));
      });
    },
 
--- a/models/things.js
+++ b/models/things.js
@@ -143,4 +143,8 @@ var Things = {
    }
  };
 
+AdapterManager.on('thing-added', function(thing) {
+  Things.handleNewThing(thing);
+});
+
 module.exports = Things;
--- a/test/common.js
+++ b/test/common.js
@@ -12,6 +12,7 @@
 /* globals assert */
 
 process.env.NODE_ENV = 'test';
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 
 var chai = require('chai');
 global.chai = chai;
--- a/app.js
+++ b/app.js
@@ -100,4 +100,6 @@ server.listen(https_port, function() {
   adapterManager.loadAdapters();
 });
 
+app._server = server;
+
 module.exports = app; // for testing
--- a/models/actions.js
+++ b/models/actions.js
@@ -64,8 +64,7 @@ var Actions = {
     action.status = 'pending';
     switch(action.name) {
       case 'pair':
-        AdapterManager.addNewThing().then(function(thing) {
-          Things.handleNewThing(thing);
+        AdapterManager.addNewThing().then(function() {
           action.status = 'completed';
         }).catch(function(error) {
           action.status = 'error';
--- a/test/integration/things-test.js
+++ b/test/integration/things-test.js
@@ -6,6 +6,7 @@
 require('../common');
 
 var Constants = require('../../constants');
+const WebSocket = require('ws');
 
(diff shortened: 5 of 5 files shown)


models/things.js
```diff
@@ -145,4 +145,8 @@ var Things = {
    }
  };
 
+AdapterManager.on('thing-added', function(thing) {
+  Things.handleNewThing(thing);
+});
```
benfrancis: See my comments in #99 regarding pairing multiple devices at a time. In principle this makes sense, but we need to think about the potential implications for other adapters.

models/things.js
```diff
@@ -116,7 +116,7 @@ var Things = {
 
      // Notify each open websocket of the new Thing
      this.websockets.forEach(function(socket) {
-       socket.send(JSON.stringify(thing.getDescription()));
+       socket.send(JSON.stringify(thing));
```
hobinjk: Both NewThingsController.get('/') and NewThingsController.ws('/') send directly JSON-ed things (https://github.com/moziot/gateway/blob/master/controllers/new_things_controller.js#L24 and https://github.com/moziot/gateway/blob/master/controllers/new_things_controller.js#L42). I didn't realize this was not the intended behavior and can modify the PR accordingly.

The problem I'm trying to solve here is that the Add New Thing UI needs to have some id for the Thing it is trying to save for the `POST /things` to work. This works in the normal cases of NewThingsController.get('/') and NewThingsController.ws('/') because they send the full Thing object for all Things currently in the database. Where it doesn't work is when a Thing is added with the Add New Thing UI open. In this case, Things.handleNewThing will be invoked and will call getDescription before sending the Thing object. Neither of the other instances of sending Thing objects calls getDescription.

This matters because if the Philips Hue is paired while the Add New Thing UI is open it will add a Thing for each bulb. The adding of these Things is communicated to the UI through  handleNewThing which invokes getDescription.

test/common.js
```diff
@@ -12,6 +12,7 @@
 /* globals assert */
 
 process.env.NODE_ENV = 'test';
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
hobinjk: The WebSocket test needs to connect to the original https server because `express-ws` is bound to it. All the other tests are okay because `chai-http` creates a new http server by default.

models/things.js
hobinjk: I was a bit wrong about the behavior of `Things.getNewThings()`. The Thing descriptions returned by it are actually from `device.asThing`, not true Thing objects.

I'm not sure what the best way to fix this is, but I don't think that adding the `id` key to `getDescription` is correct. Manually adding back the id key also feels less-than-ideal.

benfrancis: There's a difference between a "new thing" and a "thing". New things come from the adapter manager and get their thing description from Device.asThing(). Things come from the database and get their description from Thing.getDescription().

New things have an internal ID in their thing description until they are turned into things in the database which have a URL to identify them in the web API. Things using the Thing Description format we're aiming to standardise do not have an ID property because their URL is their ID.

What's happening here is the handleNewThing() is using the Thing constructor to validate the description of the new thing, which ends up stripping out the internal ID. I don't actually know how this is working because the client side needs the ID to generate the POST request https://github.com/moziot/gateway/blob/master/static/js/add-thing.js#L115 (it might be that you've found a bug).

I'd suggest that either you remove the call to the Thing() constructor in handleNewThing and pass the new thing description (including its id) directly to the websocket message as is done elsewhere, or you simply add back in the ID after it has been validated.

It's good to be getting extra eyes on this code to find these inconsistencies :)

hobinjk: Thanks for the clarification! I think I'll go with the first approach. I do think I've found a bug :)

models/things.js
hobinjk: Something I want to note here because I just confirmed it for myself: the emitted `thing-added` event is sending over the correct `device.asThing()` format, not a raw Thing object.

hobinjk:
Superseded by #108

PR #111, opened 2017-06-20
Create user

hobinjk:
To-do:
- [x] Automatically login created user
- [x] Tests

### DIFF
--- a/controllers/login_controller.js
+++ b/controllers/login_controller.js
@@ -13,15 +13,24 @@
 var express = require('express');
 var passport = require('passport');
 var path = require('path');
+var Users = require('../models/users');
 
 var LoginController = express.Router();
+
+const viewsRoot = path.join(__dirname, '../views');
 
 /**
  * Get the login page.
  */
 LoginController.get('/',
   function(request, response) {
-    response.sendFile('login.html', { root: path.join(__dirname, '../views') });
+    Users.getUsers().then(users => {
+      if (users.length > 0) {
+        response.sendFile('login.html', { root: viewsRoot });
+      } else {
+        response.sendFile('create_user.html', { root: viewsRoot });
+      }
+    });
   }
 );
 
--- a/db.js
+++ b/db.js
@@ -177,6 +177,41 @@ var Database = {
         }
       });
     }).bind(this));
+  },
+
+  /**
+   * Get all Users stored in the database.
+   *
+   * @return {Promise<Array<User>>} resolves with a list of User objects
+   */
+  getUsers: function() {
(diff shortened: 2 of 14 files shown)


hobinjk:
I think this is ready to go. One potential concern is that it reuses `/login` for the create user page which seems logical to me but could be split off to `/create_user` or `/new_user` if desired.

controllers/login_controller.js
```diff
@@ -13,15 +13,24 @@
...
  * Get the login page.
  */
 LoginController.get('/',
   function(request, response) {
-    response.sendFile('login.html', { root: path.join(__dirname, '../views') });
+    Users.getUsers().then(users => {
+      if (users.length > 0) {
+        response.sendFile('login.html', { root: viewsRoot });
+      } else {
+        response.sendFile('create_user.html', { root: viewsRoot });
```
benfrancis: I'm not 100% convinced about re-using the login route for this but I think this is probably OK for now while it's the only step in the first time setup.

db.js
```diff
@@ -177,6 +177,41 @@ var Database = {
...
+  },
+
+  /**
+   * Create a user
+   * @param {User} user
+   * @return {Promise<User>}
+   */
+  createUser: function(user) {
+    return new Promise((resolve, reject) => {
+      this.db.run('INSERT INTO users (email, password, name) VALUES (?, ?, ?)',
```
benfrancis: Let's add a name as well.

hobinjk:
Thanks for the comments! I've addressed the UI feedback and squashed the commits

benfrancis:
Oh, did you change a font size somewhere? Everything suddenly got very big.

screenshot from 2017-06-21 15-01-42

hobinjk:
@benfrancis Oh wow, that is dramatic! I did change a font size to make it look closer to the mock-up on my computer. I can scale everything down a bit

benfrancis:
I think the sizes on the login form were working OK? They were designed so it would still fit on a 320x480 mobile, but I realise the mockup doesn't have dimensions on it!

hobinjk:
I've pushed a commit with slightly smaller sizes so that things like that email input are a bit more properly sized. It also allows the "Passwords do not match" error to show up in a better location.

screen shot 2017-06-21 at 10 34 11

benfrancis:
FWIW these are the dimensions on the mockup

create_user_account-dimensions

How does that look?

hobinjk:
@benfrancis I've updated to use those dimensions, the only change was to make the Welcome title smaller. I think it would be good to bundle Fira Sans so that there's a bit more consistency

PR #118, opened 2017-06-21
Add Log Out route and menu item

### LINKED ISSUE #94: Logout
As a user I want to log out of my gateway so that nobody else can use it from this browsing session

hobinjk:
Closes #94


hobinjk:
@benfrancis I agree that using GET is suspect. My justification was that it's what all the other Passport apps do, but that's pretty weak.

I'm going to try styling a submit button to look like the Log out link, since that's what I should have done in the first place

hobinjk:
@benfrancis Updated to use POST. I feel much better now :)

benfrancis:
Nice :)

PR #131, opened 2017-06-23
Rename MozIoT Gateway to Things Gateway by Mozilla

benfrancis:


hobinjk:
The organization name (or whatever the O field is) in `tools/make-self-signed-cert.sh` also refers to MozIoT if you want to completely get rid of the old name

benfrancis:
> The organization name (or whatever the O field is) in tools/make-self-signed-cert.sh also refers to MozIoT if you want to completely get rid of the old name

@hobinjk Yeah I saw that, but as far as I know it doesn't actually do anything!

benfrancis:
Going to land this as it's basically just renaming comments and labels for branding consistency, no functional changes.

PR #165, opened 2017-06-29
bluetooth

KevinGrandon:
image

Just a WIP/test code of getting bluetooth working in an adapter using node-hid. I'm not aware of the direction we want to go, but here's a little bit of code. Do what you will with it, closing is totally fine. :)

### DIFF
--- a/adapter-manager.js
+++ b/adapter-manager.js
@@ -238,6 +238,16 @@ class AdapterManager extends EventEmitter {
   /**
    * @method handleDeviceAdded
    *
+   * Called when we discover a device, but have not paired with it yet.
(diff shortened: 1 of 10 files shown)


link-webcreations:
Hello,

Sounds really interesting, to pair an audio Bluetooth device, or a moisture/temperature sensor for instance.
Is this totally aborted ?

Regards

benfrancis:
Not aborted, we just need someone to implement a Bluetooth adapter which can expose some properties/actions/events of a useful device.

Part of the problem is that there are no standard adopted profiles for smart home in Bluetooth yet. It might be possible to support audio or health profiles, but a moisture/temperature sensor will probably require device-specific adapter code.

Issue #209, opened 2017-07-10
Switch Settings and Log Out menu entries


benfrancis:
Log Out should be the last entry

PR #212, opened 2017-07-10
Switched Settings and Log Out menu entries - #209

igilham:
This covers the functional change. I'm not aware of the testing situation yet, so I don't know if this is likely to upset the CI.


benfrancis:
Great, thanks!

PR #224, opened 2017-07-22
Scripts to build a rapberry pi gateway on the host.

dhylands:


dhylands:
See: https://github.com/mozilla-iot/wiki/wiki/Building-an-image-on-the-host for progress on this so far.

dhylands:
I created a total of 3 wiki pages which describe the various steps:
[Main](https://github.com/mozilla-iot/wiki/wiki/Creating-an-image-file-for-the-Raspberry-Pi-on-the-host)
the 2 sub-pages are linked from the main page and are called:
[Creating the base image for the Raspberry Pi](https://github.com/mozilla-iot/wiki/wiki/Creating-the-base-image-file-for-the-Raspberry-Pi)
[Building the Raspberry Pi on the host](https://github.com/mozilla-iot/wiki/wiki/Building-the-Raspberry-Pi-gateway-on-the-host)

hobinjk:
While creating the base image, the document suggests that `prepare-base.sh` will start the gateway after a reboot. However, `prepare-base.sh` doesn't set up the mozilla-iot directory with a clone of the gateway in it.

In Building the Raspberry Pi Gateway on the Host, `which` complains about not having a `-p` option. I also had a problem where because the docker command to create the rpxc script is run under sudo, the docker daemon is then owned by root, requiring all following docker commands to also be run as root.

Otherwise, everything looks good so far! It will take me a while to test the final few steps because dd-ing my 16 GB sd card is only going at 3MB/s

dhylands:
Ahh. On my build machine I have which aliased to type (rather than executing /usr/bin/which). `type` is a bash builtin and takes the -p option. So I just need to change `which` to `type` and that should fix that. I confirmed that this works on the RPi.

I'll have to play with the docker stuff. My id is a member of the docker group, but I still ran into issues trying to run docker as non-root, but I'll go back and revisit this to see if this can be resolved somehow.

dhylands:
> While creating the base image, the document suggests that prepare-base.sh will start the gateway after a reboot. However, prepare-base.sh doesn't set up the mozilla-iot directory with a clone of the gateway in it.

Yeah - this is a cheat. It bascially sets up to execute the gateway, but won't actually run the gateway until your run an image that was populated using the agg-gateway.sh script.

Did this cause any problems? I didn't try booting up the base image, and I assumed that trying to start the non-existant mozilla-iot would just cause a failure that could be ignored.

dhylands:
ok - I updated the 2 wiki pages, so that it references `type` instead of `which` and I removed/revised the comment about having a working gateway.

hobinjk:
@dhylands The missing gateway didn't cause any problems. I'm testing the `add-gateway.sh` script now. So far I've noticed that the instructions should include `kpartx` on the list of packages to install.

Good luck with docker!

hobinjk:
As someone with a 16GB sd card, it would be nice to have the images trimmed to not include the unused space but I have no idea how that would work.

dhylands:
Yeah - Booting up the system seems to auto-expand under some circumstances. 

I haven't yet experimented to see what exactly does that. We might be able to add a 3rd partition after the second to cause the amount that the 2nd partition gets expanded to be limited.
I've been using a 2Gb card and that seems to be manageable.

benfrancis:
I haven't had chance to try out this build process, but I have a couple of questions:
1) This sounds like quite a manual process, what additional work is needed to automate this as part of an automated build server?
2) Do you have any further thoughts about how we could move the Raspberry Pi specific stuff to a separate GitHub repo so that the gateway repo can be hardware/OS agnostic? I'm assuming none of this work makes that any harder?

hobinjk:
When testing the generated image I ran into some issues with first-time setup which appear to be image-specific. I visited the gateway URL and was first prompted to create a new user. After creating a user I was directed to the normal gateway page. It was only after refreshing that page that I was prompted to set up the tunnel. The tunnel setup failed with this error message:
```
le.challenges[tls-sni-01].loopback should be defined as function (opts, domain, token, cb) { ... } and should prove (by external means) that the ACME server challenge 'tls-sni-01' will succeed
[le/lib/middleware.js] created middleware
```
@andrenatal do you know what could be going on here?

benfrancis:
@hobinjk That sounds like https://github.com/mozilla-iot/gateway/issues/242

Do you have https://github.com/mozilla-iot/gateway/pull/245/commits/9bf96d890c4b46846578803e33759e983418694e ?

hobinjk:
@benfrancis Good point! I don't have that commit (this branch isn't based on the latest master)

dhylands:
> This sounds like quite a manual process, what additional work is needed to automate this as part of an automated build server?
> Do you have any further thoughts about how we could move the Raspberry Pi specific stuff to a separate GitHub repo so that the gateway repo can be hardware/OS agnostic? I'm assuming none of this work makes that any harder?

@benfrancis Probably the biggest manual step is creating the base image. I expect that we'd only need to update the base image once every month or so, or whenever a security update dictates. The base image basically has the rapberry pi image, along with a the systemd file needed to start the gateway, and the init.d file needed to redirect the ports. Neither one of those should need to be updated with any regularity.

Setting up the docker container is also a bunch of manual steps.

Once you have the base image and the docker container then creating the final image is really just a case of running 2 additional scripts - one which builds the raspberry pi version of openzwave and the pi version of all of the native node modules (this happens in the docker container) and the second script takes that output and populates a copy of the base image to create a final gateway image.

I've only tried this locally under linux so far, I'll need to figure out what additional steps need to be done to make this be part of some type of travis job.

dhylands:
rebased to the latest master

dhylands:
I'm not sure what happened here, but this is totally messed up, and I need to find the correct commits and fix this,.

dhylands:
> I'm not sure what happened here, but this is totally messed up, and I need to find the correct commits and fix this.

The merge I did was from an out-of-date copy of the gateway project. PR #256 merges the version of the files that were originally in the PR.

Issue #199, opened 2017-07-07
binarySensor Web Thing type


benfrancis:
As a user I want to connect a door sensor to the gateway so that I know when my door is open or closed

benfrancis:
Implemented by https://github.com/mozilla-iot/gateway/pull/226

Issue #185, opened 2017-07-05
Subscribe to properties over WebSocket


benfrancis:
As a user I want the state of my on/off switches to update instantly without having to reload the page

benfrancis:
Closed by #230 (doesn't include the front end part)

Issue #34, opened 2017-04-10
Subscriptions with WebSockets


benfrancis:
As a web app developer I want to subscribe to events and property changes of Things hosted by the gateway using WebSockets so that I don't have to keep polling the server.

benfrancis:
HTTP does not have a pub/sub mechanism and relies on polling, long-polling or webhooks to be notified of changes of state of resources on the server.

Our current draft of the Web Thing API [suggests](https://moziot.github.io/wot/#subscriptions-with-websockets) a way of subscribing to events and property changes by upgrading an HTTP GET request on a property or event resource to a WebSocket connection so that the server can push changes back to the client.

In #27 we have a similar requirement to subscribe to a list of new devices as the pairing process happens. This is the first time we're implementing this type of mechanism.

The current proposal in the Web Thing API spec could result in multiple open websockets per client. For a complex UI which needs to be notified of many different types of state change on multiple Things this could reach quite a large number, which might not scale well.

We need to think about the best way to observe changes of Things using web technologies, ideally without having to poll the server.

The CoAP protocol has a proposed [observe](https://tools.ietf.org/html/rfc7641) extension for this purpose, but HTTP doesn't have an equivalent.

The Web Thing Model proposal uses a [subscription](http://model.webofthings.io/#subscriptions-resource) mechanism which can use webhooks or websockets and creates subscription resources on the server which can be DELETEd to cancel a subscription. It appears that an implementation of this specification would also result in multiple websockets per client, but the [reference implementation](https://github.com/webofthings/webofthings.js) does not appear to fully implement that part of the spec.

The current draft of the W3C's Web Thing Description spec has the concept of [Events](https://w3c.github.io/wot-thing-description/#event) and notes that CoAP has a native mechanism for subscribing to events, but suggests that HTTP would have to poll the server.

benfrancis:
I'm starting to wonder whether you should just open one WebSocket per Thing...

```
GET /things/myThing
Upgrade: websocket
...
```

With the option of subscribing to all things...

```
GET /things
Upgrade: websocket
...
```

Or a subset of things

```
GET /things/?type=onOffSwitch
Upgrade: websocket
...
```

If you're subscribing to a whole Thing rather than just a single property, the challenge is then how to do that without having to invent a whole pub/sub sub-protocol to use over WebSockets and keep it feeling like a natural extension of REST (like CoAP has), rather than a completely separate API.

[...]

Issue #204, opened 2017-07-10
Thing Detail View


benfrancis:
As a user I want to read/write all of a Thing's properties, request actions and receive events all in the same view of a Thing.

Currently the only kind of UI we provide for Things is a simple on/off switch. This issue is about developing this into a more complex detail view for a Thing which may have support multiple properties, actions and events.

Each Thing should have its own unique URL in the URL bar when viewing the web app in the browser.

benfrancis:
Closed by #229.

Currently only accessible by manually typing the URL because we currently only support the on/off switch thing type which only has one property so the extra step would be redundant. Will add the rest of the UI when we add more complex device types.

PR #226, opened 2017-07-24
Add binary sensor web thing - #199

igilham:
For a starting point, I've just copied the existing OnOffSwitch code and stripped out the interactive elements (click event, turnOn, turnOff).

There are obviously better ways to structure the code but I wanted to open discussion before making any kind of architectural change. I don't do much JavaScript development so I'm interested in feedback on how you guys want to structure the code (inheritance, composition, duplication etc.)

Adding a second Thing type also makes it rather obvious that something needs to be done with #185 (#34). The thing's status (and the UI) should update in response to an event from the device.


benfrancis:
Hi Ian,

This is great, thank you for working on this.

I think you're taking the right basic approach by having the `BinarySensor` class extend the `Thing` base class.

We can probably get away with having a similar UI to the onOffSwitch type here, maybe just with a different icon and without the interaction logic as you say. Once we have more complex thing types I think we're going to need a [Thing Detail View](https://github.com/mozilla-iot/gateway/issues/204) which can represent multiple properties for a given thing. I'm working on that now.

As for the WebSockets API, that's also on my plate for the current sprint starting with #185 and #186, although I've been a bit delayed on that front. The WebSockets API is also needed for James H's work on the rules engine so definitely high priority.

As well as this front end code, we'll also need adapters on the back end like the ZigBee and Z-Wave adapters to support binary sensors like door sensors and generate Web Thing Descriptions for them.

Honestly, not much thought went into the front end architecture for 0.1 as we just wanted something working, but this is something I'm paying much more attention to for the 0.2 release. I'm currently working on a single page app with an app shell and a client side URL router. Suggestions welcome.

I think you're safe to carry on in the direction you're already going in, but please follow up here with any further questions or suggestions.

Ben

igilham:
I've started looking through the adapters etc. for references to the onOffSwitch so I can add support for the binarySensor. There isn't a lot of information to go on, but it looks like the ZWave side should be simple enough. The Zigbee HA adapter is written in a fairly opaque way right now and I'm not familiar enough with the terminology to be able to adapt it to a new thing type. Is there any decent documentation on that?

It looks like I've also missed a reference/include somewhere (Travis log). Any chance you can point to the mistake?

DavidKarlas:
Since you probably don't have Zigbee device and adapter... Imho, you should just open issue that Zigbee binarySensor support is missing... and someone else should implement it(might be me)...

igilham:
I think this is ready for review now. I raised issue #235 for Zigbee support as suggested. It would be useful to have a new default icon for sensors but I don't really have the skill to produce something suitable. Perhaps we should raise another issue for that?

benfrancis:
Thanks! Will take a look. @dhylands Would you mind reviewing the Z-Wave parts?

With regards to the icons, let's use these for now until we figure out the detail view UI:
 
on-off-sensor.png
on-off-sensor

on-off-sensor-off.png
on-off-sensor-off

on-off-sensor-on.png
on-off-sensor-on

igilham:
I've added the graphics.

dhylands:
FYI: You should be working on a branch rather than working on master. https://guides.github.com/introduction/flow/ By doing this it makes it much easier to keep your branch up-to-date and prevents all of the merge commits.

src/adapters/zwave/zwave-classifier.js
```diff
@@ -44,16 +44,18 @@ class ZWaveClassifier {
         }
     }
 
-    initOnOffSwitch(node, valueId) {
-        node.type = THING_TYPE_ON_OFF_SWITCH;
+    initOnOffThing(node, valueId, type) {
+        node.type = type;
         node.properties.set('on',
                             new ZWaveProperty(node, 'on', 'boolean', valueId));
```
dhylands: I don't like the fact that we're using the same property name for a sensor as a switch. And on doesn't necessarily make sense in many contexts. For example, with a door switch, does on == true mean that the door is open or that the door is closed? I can't find an example, but it's possible that a device implements both command classes, in which case using the same property name would be problematic.

The zwave documentation refers to event-triggered and event-not-triggered, so I'd be inclined to use a property name of "triggered", and still remain a boolean.

igilham: I've made this change. It may be worth considering consistent naming with the CSS, graphics, Thing class etc. as well.

igilham:
@dhylands, I've been wondering about how to best avoid all the merge commits etc. so thanks for the tip. Unfortunately nothing in the linked page says anything about that so it's left as an exercise for the reader to figure out. I've always assumed that since `master` is just another branch in another copy of the repo, there should be no difference from using that vs using another named branch.

I've heard about squashing commits before but have yet to find any instructions that work without risking all the work that went into a branch. It all seems rather fragile and destructive. I'd appreciate it if you can point to an example that would work, given that there's now a bunch of interspesed commits from various merges on this branch.

igilham:
I've had a go at squashing. Hopefully it works...

benfrancis:
@dhylands If we change the property name we should also update the spec http://iot.mozilla.org/wot/#binarysensor-type Don't disagree about "on" being a bit confusing in some cases though.

Just for reference, in the ZigBee spec "simple sensor" devices have a "binary input" cluster with a single "presentValue" attribute which for door sensors is 1 for open and 0 for closed.
>  When the Binary Input (Basic) cluster is used for a open/close sensor (contact
> sensor) the value of presentValue attribute needs to be specified so devices can be
> interoperable.
> For a contact sensor the meaning of the presentValue attribute is:
> 1 = Open
> 0 = Closed

So I guess "triggered" would be equivalent to "1" or "open", i.e. triggered = true when the two parts of the sensor are pulled apart. Not sure what other types of binary sensor there are...

@igilham Your squashing appears to have worked :) It is good practice to develop new work in a branch, though I've not personally noticed the difference it makes with merge commits. I have found that using `git rebase` tends to work better than `git pull` when merging upstream into my local branch though. Linus Torvalds once said that the big easter egg in git is that every other command can be expressed using `git rebase` :P On the GitHub side we just need to hit the "Rebase and merge" button rather than the "Create a merge commit" button when we land.

benfrancis:
@igilham BTW I agree about consistent naming of files and CSS classes with the thing type to avoid confusion.

Other than that, are we ready to land @dhylands ?

dhylands:
Yep - I'm happy now.

I normally always use `git pull --rebase`. Squashing isn't as important now that github has the squash and merge, although I still do it to amalgamate everything and I find it makes reviewing diffs easier.

`git rebase -i` is the magic. Here's an article on doing that. http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

When I was just learning git, I always made a backup of the directory tree, so at least I had a squashed version to compare to.

Another thing I've often done is to create a second branch (or record the hash) that points to the same place as the the tip of the branch you're about to squash. Then if you totally mess up, you can just checkout that second branch and voila, everything is magically back (including all of the history) to where things were before you started the squash.

benfrancis:
@igilham Would you prefer to change the file names and classnames to "binary-sensor" before landing or in a follow-up?

igilham:
I've renamed the files as suggested. Might as well be consistent.

benfrancis:
Great, thanks Ian!

PR #229, opened 2017-07-26
Thing Detail View

benfrancis:
This pull request includes two commits, the first splits routing between a server side Express router (for API requests) and a client side Page.js router (for app requests). The second introduces a basic Thing detail view which currently you can only access by typing in the URL of a Thing directly (until we have more complex Thing types and a UI that makes more sense).

In order to give things URLs in the web app like /things/{thingId} in a nice RESTful way, we need to be able to differentiate between API requests (JSON & WebSockets) and web app requests (HTML etc.).

The first commit splits the API and app routing so that app routing is handled mostly by a client side URL router so we can have a single page app with real URLs which can eventually also work offline using Service Workers.

It turns out this is something Express isn't actually very good at. The best approach I've found so far is to write our own content negotiation middleware which splits routing between API requests and app requests by adding a pseudo path prefix (which the end consumer never sees) based on the contents of the HTTP Accept header. It felt kind of hacky at first, but it actually works quite well.

One issue with the client side router is how to properly handle 404s when a resource doesn't exist, but this seems to be an issue with client side routing in general.

The client side router uses the very lightweight [page.js](https://visionmedia.github.io/page.js/) library which I've manually added to a /static/lib folder. Maybe not the best way of managing dependencies going forward but I didn't want to get into client side dependency management just yet.

### DIFF
--- a/static/js/things.js
+++ b/static/js/things.js
@@ -16,14 +16,21 @@
 var ThingsScreen = {
 
   NO_THINGS_MESSAGE: 'No devices.',
+  THING_NOT_FOUND_MESSAGE: 'Thing not found.',
 
   /**
    * Initialise Things Screen.
+   *
+   * @param {String} thingId Optional thing ID to show, otherwise show all.
    */
-  init: function() {
+  init: function(thingId) {
     this.thingsElement = document.getElementById('things');
     this.addButton = document.getElementById('add-button');
-    this.showThings();
+    if (thingId) {
+      this.showThing(thingId);
+    } else {
+      this.showThings();
+    }
     window.addEventListener('_thingchange', this.showThings.bind(this));
     this.addButton.addEventListener('click',
       AddThingScreen.show.bind(AddThingScreen));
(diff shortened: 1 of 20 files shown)


benfrancis:
@hobinjk This is still a bit rough and I haven't fixed & written tests yet, but I wanted to get your eyes on it so that you have time to provide input before tomorrow when I try and finish it off! Feedback welcome.

static/js/things.js
```diff
@@ -62,5 +70,45 @@ var ThingsScreen = {
...
+  /**
+   * Display a single Thing.
+   *
+   * @param {String} id The ID of the Thing to show.
+   */
+  showThing: function(id) {
+    const opts = {
+      headers: {
+        'Authorization': `Bearer ${window.API.jwt}`,
+        'Accept': 'application/json'ave-
```
hobinjk: Minor typo here

hobinjk:
This is really good work! My main question is where the thing detail view will be exposed to the user. I agree that the 404 case definitely seems to be less than ideal.

benfrancis:
@hobinjk  OK, I've fixed the typo and fixed and added integration tests.

The way I imagine the UI working is that you have the Things screen which just shows you a group of (non-interactive) things, then you tap on a Thing to view the Thing detail view including a visualisation of all of its properties and actions. See below:

things_and_thing_detail

This doesn't make a whole lot of sense to implement now because we only have on/off switches which have a single property so the extra step will seem redundant. I think the next step is to implement support for more complex device types where this UI will make sense. This is also something else we might want to get UX input on!

If you find a solution to the 404s & client side routing let me know, but I think this is a problem with client side routing in general. We could do a hybrid of client side and server side routing so the server could return a 404 for a URL we know to be invalid, but subsequent navigations wouldn't make an actual HTTP request so wouldn't get an actual HTTP response. Maybe implementing Service Workers can make the client side routing more like server side routing with real HTTP requests which are proxied through the Service Worker.

igilham:
I like the proposed detail view UI but it might be useful to retain a button on the Thing list view for simple toggle devices like switches, sirens etc. Just a thought. That's essentially the approach taken by the SmartThings mobile app and it works quite well.

benfrancis:
Thanks, I've fixed that bug. Going to land :)

PR #230, opened 2017-07-27
Send propertyStatus message over Thing websocket

hobinjk:


hobinjk:
There was some fun debugging because the ws route middleware silently swallows any errors thrown by the handler function

benfrancis:
This looks great, thank you! BTW, one thing which isn't currently included in the WebSocket API spec is the ability to communicate with multiple Things over one WebSocket. I think we may eventually want this (specifying a Thing by its URL in the message somehow) so that we don't have to open a WebSocket per Thing.

Anyway, hopefully this unblocks you on the rules engine a bit and we can also use this in the front end :)

Nice job.

hobinjk:
Thanks!

If your reason for wanting to be able to communicate with multiple things over one websocket is performance, I did some initial benchmarking. My raspberry pi 3 was able to have ~150 WebSockets open at a time when doing an intensive broadcast-to-every-websocket benchmark and even had over 2500 websockets open when each socket only had to echo the client's message.

If you want that as a potential simplification of websocket clients I can let you know what my feelings are after using the current connection-per-Thing API in the Rules Engine.

benfrancis:
> If your reason for wanting to be able to communicate with multiple things over one websocket is performance, I did some initial benchmarking. My raspberry pi 3 was able to have ~150 WebSockets open at a time when doing an intensive broadcast-to-every-websocket benchmark and even had over 2500 websockets open when each socket only had to echo the client's message.

Interesting.

> If you want that as a potential simplification of websocket clients I can let you know what my feelings are after using the current connection-per-Thing API in the Rules Engine.

Heh, OK.

PR #231, opened 2017-07-28
Redirect HTTP to HTTPS when available

### LINKED ISSUE #206: Redirect HTTP to HTTPS
If the gateway is only running on port 443/4443 then port 80/8080 should redirect. Currently you have to manually type https://

hobinjk:
The redirector forbids anything that may have credentials in it while
redirecting everything else to HTTPS. Additionally refactors the server
startup flow to be a bit more straightforward. Most notably, the
`deferredWsRoutes` hack is no longer necessary.

Fix #206

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -11,7 +11,8 @@
 module.exports = {
   cli: false,
   ports: {
-    https: 0 // 0 = find a free open port
+    https: 0, // 0 = find a free open port
+    http: 0,
   },
   adapters : {
     mock: {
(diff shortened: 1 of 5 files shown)


hobinjk:
@benfrancis Thanks! I haven't had an opportunity to test it with Let's Encrypt and I forgot to bring my Raspberry Pi here so I won't be able to test it fully. What I did do was test the approximate HTTP transitioning to HTTPS with HTTP redirect which worked

PR #233, opened 2017-07-31
Fix leak of event listener by propertyStatus API

hobinjk:


benfrancis:
Good catch. I'm terrible at remembering to remove listeners.

PR #236, opened 2017-08-03
Implement setProperty

hobinjk:
This incorporates a fair bit of WebSocket-related work, culminating in implementing support for setProperty.

The other parts are a refactor of the WebSocket test functions to help future work, a fix for a bug in the `/new_things` websocket endpoint,  and a fix for a leak in propertyStatus.

### DIFF
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -144,6 +144,39 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
   websocket.on('close', function() {
     AdapterManager.removeListener('property-changed', onPropertyChanged);
   });
+
+  websocket.on('message', function(requestText) {
+    let request = JSON.parse(requestText);
+    let device = AdapterManager.getDevice(thingId);
+    if (!device) {
+      websocket.send(JSON.stringify({
+        messageType: Constants.ERROR,
+        data: {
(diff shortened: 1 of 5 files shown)


hobinjk:
There's some kind of horrific race condition where whichever test runs second is failing unless there's a console.log in the webSocketSend function. This may take longer than anticipated

hobinjk:
I fixed the race condition, added more tests, and this feature should be good to go. The race condition was that a message could be sent on a websocket before `express-ws` was finished passing the newly opened websocket to its handler function

src/controllers/things_controller.js
```diff
@@ -144,6 +144,52 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
...
     AdapterManager.removeListener('property-changed', onPropertyChanged);
   });
+
+  websocket.on('message', function(requestText) {
+    let request = null;
+    try {
+      request = JSON.parse(requestText);
+    } catch(e) {
+      websocket.send(JSON.stringify({
+        messageType: Constants.ERROR,
```
benfrancis: Constants.ERROR doesn't actually seem to be set, did you forget to check in constants.js?

hobinjk: Nice catch! I've added the constant to the PR

PR #246, opened 2017-08-14
Remove jest-transform; Enable strict null checks

hobinjk:
`jest-transform` was making `jest --coverage` only report coverage for
TS files and had no discernable effect otherwise. Enabling [strict null
checks](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html)
makes TypeScript more useful while requiring a little bit more
annotation in some cases.


hobinjk:
For example:
```
let x: number;
x = null;
```
is prohibited by strictNullChecks. Without strictNullChecks, every type declaration in TypeScript has "|null|undefined" appended to it, which essentially loses most of the benefits of declaring the type in the first place.

hobinjk:
@dhylands With strictNullChecks you'd have to declare it as a nullable variable with `var x : Something|null`, then any further uses of `x` would have be behind guards like `if (x) x.method()`.
Alternatively, TypeScript can detect something like this:
```javascript
var x : Something;
x.method(); // not allowed, x has not been assigned to
x = new Something();
x.method(); // TS knows this will be okay
```

PR #252, opened 2017-08-16
Integrate Rules Engine

hobinjk:
Issue JWT token to rules engine; add menu item; configure rules engine's
gateway property.


hobinjk:
The "configure rules engine's gateway property" is the main thing I feel vaguely sad about as ideally it would auto-configure based on which server the gateway decides to use

config/default.js
```diff
@@ -55,5 +55,6 @@ module.exports = {
     domain: 'mozilla-iot.org',
     pagekite_cmd: './pagekite.py',
     port: 443
-  }
+  },
+  gateway: 'https://localhost:4443'
```
benfrancis: So how does this work with the SSL tunnel?

src/constants.js
```diff
@@ -20,6 +20,7 @@ exports.LOGIN_PATH = '/login';
 exports.LOG_OUT_PATH = '/log-out';
 exports.SETTINGS_PATH = '/settings';
 exports.DEBUG_PATH = '/debug';
+exports.RULES_ENGINE_PATH = '/rules-engine';
```
benfrancis: If the rules engine is meant to be a pluggable module, it seems strange to hard code this here.

src/router.js
```diff
@@ -46,6 +47,17 @@ var Router = {
       }
     });
 
+    try {
+      let rulesEngine = require('rules-engine/index.js');
+      // Issue a token without an associated user
```
benfrancis: Again, rules engine code hard coded in the gateway means it isn't really a pluggable module, we'd have to add this code for every third party module.

src/views/index.html
```diff
@@ -77,6 +77,7 @@
     <nav id="main-menu" class="hidden">
       <ul>
         <li><a id="things-menu-item" href="/things" class="selected">Things</a></li>
+        <li><a id="rules-engine-menu-item" href="/rules-engine">Rules Engine</a></li>
```
benfrancis: Let's add an icon for this

benfrancis:
Here's an icon for the rules engine menu item:

rules-icon

Not very imaginative, I know! Let me know if you have any better ideas for an icon ;)

src/router.js
```diff
@@ -46,6 +47,17 @@ var Router = {
...
+      JSONWebToken.issueToken(-1).then(jwt => {
```
benfrancis: So this JWT isn't associated with a user? How does that work exactly? What happens when the user logs out? What about when another user logs in? Just asking because I don't understand the JWT stuff very well.

src/views/index.html
hobinjk: Thanks for the icon! What are your thoughts on how to best approach hiding this when the rules engine isn't enabled?

config/default.js
hobinjk: This string is a configuration for `node-fetch` for the rules engine's server code to talk to the gateway. This means it doesn't break when the server is tunneling. Thank you for pointing this out because there's a bug in the client-side javascript of the rules engine where this same string occurs when it is subject to the tunnel

src/router.js
hobinjk: When a user is removed all of the JWTs associated with their user ID are deleted from the database. When a user logs out only the JWT associated with that session is deleted.

By not being associated with any user this JWT is effectively permanent unless it is manually revoked which is a bit problematic. If we want to use this pattern in the future it would be good to have expiring tokens and to delete all tokens with a user id of -1 when starting the server

src/views/index.html
benfrancis: We could add a checkbox in an "experiments" section of settings which defaults to off https://docs.google.com/presentation/d/1rMeg6hep8JQflqwbiNIbhSoK9StCeVZ2npphmQHKda4/edit#slide=id.g24e68acc65_0_9 which could eventually become an add-ons section.

Ideally the icon would be provided by the rules engine add-on itself, presumably an add-on would specify it in a manifest a bit like web extensions add buttons to the toolbar https://developer.mozilla.org/en-US/Add-ons/WebExtensions/user_interface/Browser_action

benfrancis: FWIW I've started on an Experiments section for Settings in https://github.com/mozilla-iot/gateway/pull/254 - unless you find a way to automatically detect rules engine for now.

PR #254, opened 2017-08-17
Show/hide floorplan view

benfrancis:
The actual floorplan view isn't ready yet, but here is code for an "experiments" section for settings, a way to persist gateway settings to disk and a settings item to show and hide the experimental floorplan view with a checkbox. I hope we can use this for other experiments like speech too.


hobinjk:
The first thing I'm seeing is that this is missing the changes to `yarn.lock` caused by installing `node-persist`

benfrancis:
Oh, that would be because I'm still using npm! lol.

benfrancis:
Thanks

PR #259, opened 2017-08-25
Remove incorrect yarn installation instructions

hobinjk:


hobinjk:
I feel iffy advocating for something the Yarn people frown upon (even if though it's the most widely-applicable solution)

Issue #258, opened 2017-08-24
Position thing on floorplan


benfrancis:
As a user I want to position a thing on a floorplan of my home

PR #264, opened 2017-09-05
Upload floorplan - closes #249

### LINKED ISSUE #249: Upload floorplan
As a user I want to upload a floorplan of my home so I can arrange things spatially on a map of my home

benfrancis:
This pull request provides the ability for a user to upload a floorplan of their home. The floorplan is used as a background image which things can be arranged on top of.


hobinjk:
The uploading functionality looks good to me. I think it could use an onboarding hint to prompt you to click the edit icon as it was a bit of an intimidating blank blue screen at first.

benfrancis:
> I think it could use an onboarding hint to prompt you to click the edit icon as it was a bit of an intimidating blank blue screen at first.

I agree, see the mockups in #249. I haven't figured out a neat way to implement that yet, I thought I might do it as part of #258 as the hint can cover both. Also, hopefully the blank screen will be less blank once there are "things" on it. I don't want users to have to "add" things twice so the intention is that things added in the things screen automatically appear on the floorplan to be positioned.

PR #267, opened 2017-09-07
Basic Over the Air Updates of node_modules and gateway source

hobinjk:
See https://github.com/hobinjk/gateway/blob/archive-node_modules/doc/how-to-ota.md for more information.

Everything appears to work

### todo:
- [x] Turn extract-ca-archive.sh and create-ca-archive.sh into functions
- [ ] Report upgrade status to user (may be punted to future work)


hobinjk:
@dhylands If you have the time could you take a quick look at this to make sure it looks reasonable?

tools/check-for-update.js
```diff
@@ -1,41 +1,55 @@
...
-const util = require('util');
 
-const childProcess = require('child_process');
-const exec = util.promisify(childProcess.exec);
+const exec = require('child_process').exec;
 
 let pkg = JSON.parse(fs.readFileSync('package.json', {encoding: 'utf-8'}));
 
-fetch('https://api.github.com/repos/mozilla-iot/gateway/releases').then(res => {
+fetch('https://api.github.com/repos/hobinjk/gateway/releases').then(res => {
```
dhylands: Perhaps the root of the repo should go into a variable? That way there is only one place to change.

Presumably having this point to hobinjk is just temporary?

tools/upgrade.sh
```diff
@@ -2,6 +2,18 @@
...
+# wget $node_modules_archive_url
+
+pwd
+
+gateway_old=./gateway
+cp -r $gateway_old /tmp/gateway
+rm -r /tmp/gateway/node_modules
+$gateway_old/tools/extract-ca-archive.sh gateway-*.tar.gz /tmp
+$gateway_old/tools/extract-ca-archive.sh ../node_modules-*.tar.gz /tmp/gateway
+# TODO: will sudo ever prompt for password?
```
dhylands: The file: `/etc/sudodoers.d/010_pi-nopasswd` contains the following:
```
pi ALL=(ALL) NOPASSWD: ALL
```
which is what makes the pi user not need to enter a password.

doc/how-to-ota.md
```diff
@@ -0,0 +1,41 @@
...
+   "gateway", this command should do the trick: `tar czvf gateway.tar.gz
+   gateway`
+5. Run `yarn install` in the new clone
+6. Move `gateway.tar.gz` into the clone
+7. Run [tools/create-release-archives.sh](https://github.com/mozilla-iot/gateway/blob/master/tools/create-release-archives.sh), producing `gateway-<hash>.tar.gz`
+   and `node_modules-<hash>.tar.gz`
+8. Create a release using the version from step 1 as its tag name
+9. Attach the release archive to this release
+10. Publish the release when ready or save as a draft
+
```
dhylands: Presumably we could create a script which automates the above?

doc/how-to-ota.md
```diff
@@ -0,0 +1,41 @@
...
+1. Download and verify `gateway-<hash>.tar.gz`
+   and `node_modules-<hash>.tar.gz`
+4. Copy the current gateway into `/tmp/gateway_new`
+5. Extract `gateway-<hash>.tar.gz` into `gateway_new`, overwriting any matching
+   files
+6. Replace `gateway_new/node_modules` with the new release's `node_modules`
+7. Stop `mozilla-iot-gateway.service`
+8. Move the current gateway to `gateway_old`
+9. Move the new gateway to `gateway`
+10. Start `mozilla-iot-gateway.service`
```
dhylands: We should make sure that we test out of space failures. On a 2Gb card, I don't think that there is space for 2 copies of node_modules.

image/prepare-base.sh
```diff
@@ -150,4 +150,11 @@ END
...
-echo "Done"
\ No newline at end of file
+# Check for an update every day
+sudo sh -c 'cat > /etc/cron.daily/mozilla-iot-gateway-update.sh' <<END
+#!/bin/bash
+su pi -c 'node /home/pi/mozilla-iot/gateway/tools/check-update.js'
+END
+sudo chmod +x /etc/cron.daily/mozilla-iot-gateway-update.sh
+
+echo "Done"
```
dhylands: Since this is running under the pi user, I think that we could launch this process from within the gateway itself. We just need to be careful to launch it as a detached process. This would allow the scheduling to be configured from within the gateway rather than relying on cron. Although I like the fact that cron will run regardless. Even if the gateway has badly crashed.

tools/extract-release.sh
```diff
@@ -0,0 +1,10 @@
+gateway_old=./gateway
+$gateway_old/tools/extract-ca-archive gateway-*.tar.gz gateway_new
+pushd gateway_new
+$gateway_old/tools/extract-ca-archive ../node_modules-*.tar.gz node_modules
+cp -r $gateway_old/db.sqlite3 $gateway_old/settings ./
+popd
+systemctl gateway.service stop
```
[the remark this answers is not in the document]
dhylands: Shouldn't this also be copying the *.pem files and the tunneltoken file?

tools/extract-node_modules-archive.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+digest=$1
```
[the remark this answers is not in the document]
hobinjk: I've added error checking to all of the user-facing scripts

tools/extract-node_modules-archive.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ "$digest" != "$check_digest" ]; then
+  echo "Digest mismatch: $digest != $check_digest"
+  exit -1
+fi
+
+if [ -d node_modules ]; then
+  rm -r node_modules
+fi
+
+tar xzf "node_modules-$digest.tar.gz" ./node_modules
```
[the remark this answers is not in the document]
hobinjk: I updated the extraction process so it doesn't care about the exact path (after banging my head against this problem repeatedly)

tools/create-ca-archive.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+base_name=$1
```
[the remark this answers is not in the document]
hobinjk: I turned it into a function named makeContentAddressedArchive

tools/extract-release.sh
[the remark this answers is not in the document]
hobinjk: I changed the approach used in `tools/upgrade.sh` where instead of copying single files it copies the entire `gateway` directory then uses tar to overwrite the files. I also moved the `systemctl gateway.service stop` to much earlier in the process so there will be enough time for the gateway to shut down cleanly.

tools/check-for-update.js
hobinjk: I've factored it out into a variable to make changing it easier. In the current revision it should be back to mozilla-iot

tools/upgrade.sh
hobinjk: Great! Thanks for the info

doc/how-to-ota.md
hobinjk: Yes, this is definitely a place where a script would help out

doc/how-to-ota.md
hobinjk: Yeah, this is one of the concerns I don't know how to handle yet. Do you have any ideas for how to handle this? I can bail if extraction fails, but I'm not sure if there's anything better to do.

image/prepare-base.sh
hobinjk: I can switch the scheduling over to the gateway if you prefer.

If the gateway has crashed it should be picked up by the OnFailure handler which would then run `tools/rollback.sh`. The time when `cron` would be most useful is if `gateway_old` is also broken.

doc/how-to-ota.md
dhylands: bailing if the extraction fails would be a good start. We may just need to stipulate a minimum size for the sdcard in order to allow updates to be performed.

image/prepare-base.sh
dhylands: Using cron is probably fine for now.

tools/extract-release.sh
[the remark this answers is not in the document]
dhylands: Perhaps just add a comment then where that copy occurs that it will copy the pem files, tunneltoken, and db. That way somebody searching will find it and it will act as a reminder in case somebody tries to change it.

doc/how-to-ota.md
hobinjk: I've updated it to do that, I'll try to test it to make sure it doesn't explode

hobinjk:
@dhylands This should be reasonably complete now. I tested on the base 2GB filesystem and it did run out of space and fail, but it left everything in a clean and working state.

tools/rollback.sh
```diff
@@ -1,8 +1,29 @@
...
+   local filename=$1
+   local changed=`stat -c %Y "$filename"`
+   local now=`date +%s`
+   local elapsed
+
+   let elapsed=now-changed
+   # if less than 60 * 60 * 24 * 14 seconds have passed
+   if [ $elapsed -lt 1209600 ]; then
+     return 1
+   else
```
dhylands: nit: following the mozilla coding standards, I'd drop the else and unindent the return 0
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#General_C.2FC.2B.2B_Practices

tools/rollback.sh
```diff
@@ -1,8 +1,29 @@
...
+     return 0
+   fi
+}
+
+if [ -d "gateway_old" ] && [ ! -z `recentEnough "gateway_old"` ]; then
```
dhylands: nit: Personally, I prefer to use $(recentEnough ...) rather than backticks. I find the $() notation makes it much easier to match up when expressions get more complex.

tools/rollback.sh
```diff
@@ -1,8 +1,29 @@
...
+  if [ -d "gateway_failed" ]; then
+    rm -fr gateway_failed
```
dhylands: Since you're using rm -f you don't need to check if the directory exists or not. You can just use `rm -rf gateway_failed` and that will succeed whether gateway_failed exists or not.

tools/upgrade.sh
```diff
@@ -25,11 +25,14 @@ gateway_old=./gateway
 # node_modules
 cp -r $gateway_old /tmp/gateway
 if [ -d "/tmp/gateway/node_modules" ]; then
-	rm -r /tmp/gateway/node_modules
+  rm -r /tmp/gateway/node_modules
```
dhylands: You could just use rm -rf here and remove the test

tools/upgrade.sh
```diff
@@ -25,11 +25,14 @@ gateway_old=./gateway
...
 fi
 
 die() {
   sudo systemctl start mozilla-iot-gateway.service
+  rm gateway-*.tar.gz
```
dhylands: These rm's should probably use rm -f so that they won't fail if the files don't exist for some reason

PR #268, opened 2017-09-07
WIP for Integration of the Intent Parser

tamarahills:
Still some TODOs and cleanups left, but wanted to push this earlier rather than later.


package.json
```diff
@@ -42,8 +42,9 @@
     "node-persist": "^2.1.0",
     "onoff": "^1.1.2",
     "openzwave-shared": "^1.3.2",
+    "request-promise": "^4.2.1",
     "serialport": "^4.0.7",
-    "sqlite3": "^3.1.8",
+    "sqlite3": "^3.1.9",
```
benfrancis: Any particular reason for this change? @hobinjk Will this interfere with yarn at all? I'm unsure how that works.

src/constants.js
```diff
@@ -19,6 +19,7 @@ exports.ACTIONS_PATH = '/actions';
 exports.LOGIN_PATH = '/login';
 exports.LOG_OUT_PATH = '/log-out';
 exports.SETTINGS_PATH = '/settings';
+exports.COMMAND_PATH = '/command';
```
benfrancis: I'd suggest calling this `/commands` (plural) for consistency. RESTfully speaking it's a collection of command resources you're adding a command to.

src/controllers/command_controller.js
```diff
@@ -0,0 +1,185 @@
...
+'use strict';
+
+const express = require('express');
+const Constants = require('../constants.js');
+const rp = require('request-promise');
+
+const CommandController = express.Router();
+
+
+//TODO:  move the Auth key somewhere else!
```
benfrancis: Yeah this definitely can't be hard coded.

src/controllers/command_controller.js
```diff
@@ -0,0 +1,185 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+
+const express = require('express');
+const Constants = require('../constants.js');
+const rp = require('request-promise');
+
+const CommandController = express.Router();
```
benfrancis: I'd suggest `CommandsController` rather than `CommandController`

webpack.config.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
   entry: './src/app.js',
   target: 'node',
   node: {
-    __dirname: false,
+    __dirname: true,
```
benfrancis: Just curious, what is this for?

src/app.js
```diff
@@ -117,6 +119,23 @@ function getOptions() {
   return options;
 }
 
+/**
+ * The command parser talks to the server over the public HTTP/WS API,
+ * the gateway needs to configure it with a JWT and a server address
+ * @param {http.Server|https.Server} server
+ */
+function commandParserConfigure(server) {
```
benfrancis: I'm not sure whether this logic should live in app.js or in `CommandsController` itself, by just passing the server object to `CommandsController` directly, or via `Router.configure()`. Maybe it is better to inject the jwt rather than have `CommandsController` import `JSONWebToken` itself though, what are you doing for the rules engine @hobinjk ?

benfrancis: Reviewing #263 and I realised this code was copied from there :) Will discuss in #263

package.json
hobinjk: Yes, this change requires this PR to have `yarn install` run and the changes to `yarn.lock` committed. The current version of sqlite3 in the lockfile is `3.1.8`.

webpack.config.js
hobinjk: This is from #263 where the existing false setting made `__dirname` return incorrect results in some cases (I think setting it to false means that `__dirname` is always equal to 'src/'). I just updated #263 to not require this change and instead put its files in the proper location

benfrancis:
BTW Tamara, you should really fork this repo and create a feature branch in your fork rather than create a feature branch in the main repo.

src/controllers/command_controller.js
andrenatal: Yes, maybe we can add the key here: https://github.com/mozilla-iot/gateway/blob/master/config/default.js

package.json
andrenatal: I had a problem with sqlite 3.1.8 too and only had it fixed after update to the latest version.

PR #270, opened 2017-09-11
View and position things on floorplan - closes #250 closes #258

### LINKED ISSUE #250: View Things on Floorplan
As a user I want to see my things on a floorplan of my home

### LINKED ISSUE #258: Position thing on floorplan

benfrancis:
This pull request adds SVG views for Things to represent them in the floorplan view and allows things to be positioned on the floorplan and have their co-ordinates persisted in the database. SVG is cool!

This isn't quite as polished as I would like, but I've run out of time.

I'm not sure saving the floorplan co-ordinates directly in the thing descriptions in the database is necessarily the best place for this data long term, but it serves the purpose for now.

I'm currently trimming the length of thing labels to fit the available space, because SVG text is hard.

Also ideally you wouldn't have to leave the floorplan view to manipulate a thing (e.g. turn switch on and off), eventually I'd like to have a UI design where the thing's properties expand out inside the floorplan view in a nicely animated fashion. In the meantime I might file a follow-up bug to add a back button to the thing detail view.


static/js/on-off-switch.js
```diff
@@ -16,31 +16,52 @@
...
   return '<div class="thing on-off-switch">' +
          '  <div class="thing-icon"></div>' +
          '  <span class="thing-name">' + this.name + '</span>' +
          '</div>';
 };
 
 /**
+ * SVG view for on/off switch.
+ */
+Thing.prototype.svgView = function() {
```
hobinjk: If this is the SVG view for the on/off switch shouldn't this be OnOffSwitch.prototype?

benfrancis: Huh, yes. I thought I fixed this once.

static/js/thing.js
```diff
@@ -37,24 +41,56 @@ var Thing = function(description) {
...
+  if (format != 'svg') {
+    this.element.addEventListener('contextmenu',
+      this.handleContextMenu.bind(this));
+  }
   return this;
 };
 
 /**
- * HTML view for Thing.
+ * HTML view for uknown Thing.
```
hobinjk: typo

hobinjk:
Another issue is that `static/uploads/floorplan.svg` is in this PR but is in the .gitignore. I think it would be better to put the default floorplan.svg in a different location because otherwise anyone who uploads a floorplan.svg will have to deal with git complaining

static/js/thing.js
benfrancis: Fixed.

benfrancis:
Thanks for the review.

I've fixed the typo and stopped things being dragged outside the floorplan.

I don't like the default floorplan being at uploads/floorplan.svg either. I did it this way to avoid having to store a boolean in a database somewhere to say whether or not a floorplan has been uploaded but I agree it isn't ideal because apart from the .gitignore strangeness you also lose the default floorplan file when you upload a new one. Can we file a follow-up bug to improve this because I don't think I have time to implement it today? I'd also prefer it if the textual hint didn't appear in edit mode because it's kind of confusing.

hobinjk:
Okay, I'll file a follow-up

PR #272, opened 2017-09-12
Partial fix for the 0.1-p1 release

dhylands:
This addresses an issue where loadAdapters gets called twice, once when creating the http server and a second time when starting the https server.

### DIFF
--- a/adapter-manager.js
+++ b/adapter-manager.js
@@ -28,6 +28,7 @@ class AdapterManager extends EventEmitter {
     this.devices = {};
     this.deferredAdd = null;
     this.deferredRemove = null;
+    this.adaptersLoaded = false;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


adapter-manager.js
```diff
@@ -357,11 +364,16 @@ class AdapterManager extends EventEmitter {
...
+    if (!this.adaptersLoaded) {
+      // The adapters are not currently loaded, no need to unload.
+      return;
+    }
     for (var adapterId in this.adapters) {
       var adapter = this.adapters[adapterId];
       console.log('Unloading', adapter.name);
       adapter.unload();
     }
+    ths.adaptersLoaded = false;
```
hobinjk: Typo, s/ths/this

hobinjk:
Otherwise LGTM. I think you can point this PR at the 0.1-p1 branch using GitHub, though I don't remember the exact location of that option

dhylands:
Yeah - I think I need to close this PR and create another one in order to get it branched properly.

dhylands:
This PR was superceded by #274

PR #273, opened 2017-09-12
Add a intent parser that parses a text into an intent  and then execu…

tamarahills:
…tes the command.

### DIFF
--- a/src/controllers/commands_controller.js
+++ b/src/controllers/commands_controller.js
@@ -1,0 +1,210 @@
+/**
+ * Commands Controller.
+ *
+ * Manages HTTP requests to /commands.
+ *
+ * Grammar that the parser understands:
+ *  Turn the <tag> light <on|off>
+ *  Turn <tag> <on|off>
+ *  Shut <tag> <on|off>
+ *  Shut the <tag> light <on|off>
+ *
+ * <tag> must match the .name property of one of the /things objects
+ *  in order for the command to be executed.
+ *
+ * Sample curl command to test from the command line:
+ * curl -H 'Authorization:Bearer '<JSONWebToken>'
+ *        -H "Content-Type: application/json"
+ *        -H "Accept: application/json"
+ *        -d '{"text":"turn the kitchen light on"}'
+ *        https://localhost:4443/commands
+ *        -k
+ *
+ * HTTP Response Codes        Status message
+ *              201           Command Created
+ *              400           Text element not defined
+ *              404           Thing Not Found
+ *              406           Unable to understand command
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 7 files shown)


andrenatal:
@tamarahills you need to run `yarn instal` and commit `yarn.lock` in order to make travis pass

src/controllers/commands_controller.js
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Parses the intent for a text sentence and sends to the API.ai intent
+ * parser to determine intent.  Then executes the intent as an action on the
+ * thing API.
+ */
+CommandsController.post('/', function (request, response) {
+
+  if(!request.body || request.body['text'] === undefined) {
+    response.status(400).send('Text element not defined');
```
hobinjk: As an API endpoint, errors should be sent back in a consistent JSON format like `{error: 'Message goes here'}`.

src/controllers/commands_controller.js
```diff
@@ -0,0 +1,207 @@
...
+};
+
+/**
+ * Parses the intent for a text sentence and sends to the API.ai intent
+ * parser to determine intent.  Then executes the intent as an action on the
+ * thing API.
+ */
+CommandsController.post('/', function (request, response) {
+
+  if(!request.body || request.body['text'] === undefined) {
```
hobinjk: `request.body['text']` should be written as `request.body.text`

src/controllers/commands_controller.js
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Formats the body for the request to API.ai and returns the String
+ */
+function getAiBody(text) {
+  let body = {
+    'v':'20150910',
+    'query': '',
+    'lang':'en',
+    'sessionId':'b083ae95-e39a-4954-a05b-84687f50f790',
+    'timezone':'2017-07-24T10:45:52-0400'
```
hobinjk: Is hardcoding this okay?

src/router.js
```diff
@@ -76,6 +76,8 @@ var Router = {
       require('./controllers/users_controller'));
     app.use(API_PREFIX + Constants.UPLOADS_PATH,
       auth, require('./controllers/uploads_controller'));
+    app.use(API_PREFIX + Constants.COMMANDS_PATH,
```
hobinjk: missing `auth` middleware

src/controllers/commands_controller.js
tamarahills: fixed.

src/controllers/commands_controller.js
tamarahills: Added these as a message in JSON format.

src/router.js
tamarahills: fixed.

src/controllers/commands_controller.js
tamarahills: I removed this as it's optional and we don't have anything depending on TZ ATM.

PR #274, opened 2017-09-12
0.1 p1 release

dhylands:
This fixes a problem with trying to load the adapters twice, once when starting the http server and a second time when starting the https server.


dhylands:
This commit was reviewed by @hobinjk in #272. This PR was setup to branch off of 0.1-p1 rather than master, so carrying the r+

PR #275, opened 2017-09-13
Backport run-app.sh from master

dhylands:
This is needed by systemctl scripts which are installed using the prepare-base.sh script.


dhylands:
r=me

PR #281, opened 2017-09-18
Use WS API to update property status

hobinjk:
Adds a `ws` property to Thing in the frontend. The OnOffSwitch then listens for `propertyStatus` messages, using them to update the UI. This does not yet use the `setProperty` message to replace the PUT requests.

### DIFF
--- a/static/js/on-off-switch.js
+++ b/static/js/on-off-switch.js
@@ -77,20 +77,33 @@ OnOffSwitch.prototype.updateStatus = function() {
       'Accept': 'application/json'
     }
   };
-  fetch(this.onPropertyUrl, opts).then((function(response) {
+
+  fetch(this.onPropertyUrl, opts).then(function(response) {
     return response.json();
-  }).bind(this)).then((function(response) {
-    this.properties.on = response.on;
-    if (response.on === null) {
-      return;
-    } else if(response.on) {
-      this.showOn();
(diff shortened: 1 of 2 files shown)


static/js/on-off-switch.js
```diff
@@ -77,23 +77,36 @@ OnOffSwitch.prototype.updateStatus = function() {
...
+  }
+  this.properties.on = data.on;
+  if (data.on === null) {
+    return;
+  }
+  if (data.on) {
+    this.showOn();
+  } else {
+    this.showOff();
+  }
```
dhylands: Should this code and the code for drawing it in the first place be factored?

Perhaps have a show function which is called from both places?

hobinjk: The code for drawing the initial state of the OnOffSwitch uses `onPropertyStatus` as well [here](https://github.com/mozilla-iot/gateway/pull/281/files#diff-5343f48079b432eb1ed84e1f67cdc038R84) then it's updated in response to messages [here](https://github.com/mozilla-iot/gateway/pull/281/files#diff-3760c5a1e06ccd33ec83b97e6a996154R41)

dhylands: I was probably looking at the unpatched code and missed that it was changed in this PR :)

PR #283, opened 2017-09-21
Add a UI for OTA updating

hobinjk:
Looks like this:
screen shot 2017-09-21 at 2 22 03 pm

screen shot 2017-09-21 at 2 43 37 pm

To test, run `image/prepare-base.sh` to put the unit descriptions into place, then replace "mozilla-iot" with "hobinjk" in `src/controllers/updates_controller.js` and `tools/check-for-update.js`.

I'd recommend also commenting out the OnFailure in `mozilla-iot-gateway.service` so that any crashes caused by local development will not make the Gateway think it has to rollback an update. Running `sudo systemctl disable mozilla-iot-gateway.check-for-update.timer` will disable automatic updates for now.

### DIFF
--- a/src/controllers/updates_controller.js
+++ b/src/controllers/updates_controller.js
@@ -1,0 +1,117 @@
+const childProcess = require('child_process');
+const fs = require('fs');
+const fetch = require('node-fetch');
+const semver = require('semver');
+const PromiseRouter = require('express-promise-router');
+
+const UpdatesController = PromiseRouter();
+
+function readVersion(packagePath) {
+  return new Promise(function(resolve, reject) {
+    fs.readFile(packagePath, {encoding: 'utf8'}, function(err, data) {
+      if (err) {
+        reject(err);
+        return;
+      }
+      let pkg = JSON.parse(data);
+      if (!semver.valid(pkg.version)) {
+        reject(new Error('invalid gateway semver: ' + pkg.version));
+        return;
+      }
+      resolve(pkg.version);
+    });
+  });
+}
+
+function stat(path) {
+  return new Promise(function(resolve, reject) {
+    fs.stat(path, function(err, stats) {
+      if (err) {
+        if (err.code === 'ENOENT') {
+          resolve(null);
+        } else {
+          reject(err);
+        }
+      } else {
+        resolve(stats);
+      }
+    });
+  });
+}
+
+/**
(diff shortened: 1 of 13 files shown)


hobinjk:
@dhylands I forgot to write tests for the updates controller, I'll re-request review when I've fixed this

src/controllers/updates_controller.js
```diff
@@ -104,10 +104,14 @@ UpdatesController.get('/status', async function(request, response) {
 });
 
 UpdatesController.post('/update', async function(request, response) {
-  const child = childProcess.spawn('node', ['tools/check-for-update.js'], {
-    detached: true,
-    stdio: 'ignore'
-  });
+  const child = childProcess.spawn(
+    process.argv[0],
```
dhylands: If a relative path is used, then we should probably also save away the current working directory at the time of program launch, or save away an absolute version of argv[0] at program start (just in case the current directory gets changed).

static/js/settings.js
```diff
@@ -238,20 +238,23 @@ var SettingsScreen = {
...
             isDown = true;
           }
+          setTimeout(checkStatus, 5000);
         });
-      }, 1000);
+      }
+      checkStatus();
     }).catch(function() {
       updateNow.textContent = 'Error';
     });
```
dhylands: So how frequently does this wind up checking github? I'd guess that we should probably only check a few times (maybe even just once) per day.

We should probably also check if the user requests a check.

src/controllers/updates_controller.js
hobinjk: `process.argv[0]` is secretly [process.execPath](https://nodejs.org/api/process.html#process_process_execpath) which is thankfully always absolute.

static/js/settings.js
hobinjk: This ends up checking every time the page is loaded, which is how I got rate-limited by GitHub. I definitely want to add some form of caching before landing this

hobinjk:
Thanks!

The version comparisons are done by [the semver package](https://www.npmjs.com/package/semver). The check-for-update.js script assumes that the most chronologically recent release that is not a prerelease or draft is also the one with the highest semver as its tag name. Upon thinking about this further I'm going to sort by semver to be better at filtering out things like 0.1-p1.

PR #284, opened 2017-09-21
Reinstate commands tests

hobinjk:
The test runner ignored `commands_test` because of the underscore. The
tests worked perfectly once updated for the command -> commands
renaming.


hobinjk:
It tripped me up too, it turns out that `jest` when run on its own is configured to only recognize files ending in `-test.js` so Travis silently ignored the commands_test.js file.

PR #286, opened 2017-09-22
Test Rules Engine's rule execution

hobinjk:
This takes advantage of the fact that the rules engine is part of the gateway to perform a complete test of its rule execution. Notably this also solves an intermittent issue where `storage.init` calls in tests could race and make everything sad.

### DIFF
--- a/src/rules-engine/Engine.js
+++ b/src/rules-engine/Engine.js
@@ -16,18 +16,18 @@ class Engine {
 
   /**
    * Get a list of all current rules
-   * @return {Array<Rule>} rules
+   * @return {Promise<Array<Rule>>} rules
    */
   getRules() {
     let rulesPromise = Promise.resolve(this.rules);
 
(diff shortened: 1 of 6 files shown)


dhylands:
Admittedly, I didn't try this to perform an update.

hobinjk:
Thanks for reviewing! The way I understand async/await is that `async` means a function can use `await` and will return a promise. `await` then runs the promise to completion and returns whatever it resolves to.

```javascript
async function doThing() {
  let stuff = await thing();
  return await otherThing(stuff);
}
```
is equivalent to
```javascript
function doThing() {
  return thing().then(stuff => {
    return otherThing(stuff);
  });
}
```

hobinjk:
@dhylands I think that second comment is for #283 ?

dhylands:
Doh - yeah (second comment was for #283)

PR #287, opened 2017-09-22
Initial implementation of Adapter IPC

dhylands:
See the wiki page https://github.com/mozilla-iot/wiki/wiki/Adapter-IPC for some description of the IPC mechanism.


dhylands:
I haven't written any tests yet, and there are still some things that needs to be cleaned up (see the wiki page).

src/adapters/plugin/plugin-server.js
```diff
@@ -0,0 +1,83 @@
...
+  onMsg(msg) {
+    this.verbose &&
+      console.log('AdapterManagerServer rcvd:', msg);
+
+    switch (msg.messageType) {
+
+      case 'registerAdapter':
+        var adapter = this.registerAdapter(msg.data.id);
+        this.ipcSocket.sendJson({
+          messageType: 'resgisterAdapterReply',
```
hobinjk: Typo

src/adapters/plugin/plugin-server.js
```diff
@@ -0,0 +1,83 @@
...
+          },
+          id: msg.id,
+        });
+        break;
+
+    }
+  }
+
+  registerAdapter(adapterId) {
+    if (adapterId in this.adapters) {
```
hobinjk: I think `this.adapters.hasOwnProperty(adapterId)` would be better in case an adapter id happens to be something weird like `toString`

src/adapters/plugin/adapter-proxy.js
```diff
@@ -0,0 +1,137 @@
...
+var idBase = 0;
+
+/**
+ * Class used to describe an adapter from the perspective
+ * of the gateway.
+ */
+class AdapterProxy extends Adapter {
+
+  constructor(adapterManager, adapterId, pluginServer) {
+    super(adapterManager, '???');
```
hobinjk: Something more explanatory than '???' would be nice

src/adapters/plugin/adapter-manager-proxy.js
```diff
@@ -0,0 +1,139 @@
...
+const DEBUG = false;
+
+class AdapterManagerProxy extends EventEmitter {
+
+  constructor (pluginClient) {
+    super();
+
+    this.pluginClient = pluginClient;
+
+    this.on('property-changed', (property) => {
```
hobinjk: I think this should be Constants.PROPERTY_CHANGED

src/adapters/plugin/adapter-manager-proxy.js
```diff
@@ -0,0 +1,139 @@
...
+      DEBUG && console.log('proxy: Got "property-changed" notification for',
+                           property.name, 'value:', property.value);
+
+      this.pluginClient.sendNotification('propertyChanged', {
```
hobinjk: It would be good to have all constant strings like this defined somewhere. I think the best approach would be to change Constants.PROPERTY_CHANGED to be 'propertyChanged' then use it everywhere

src/adapters/plugin/adapter-proxy.js
```diff
@@ -0,0 +1,137 @@
...
+
+    idBase += 1;
+    this.idBase = idBase * 1000;
+    this.ipcSocket.idBase = this.idBase;
+  }
+
+  startPairing(timeoutSeconds) {
+    DEBUG && console.log('AdapterProxy: startPairing',
+                         this.name, 'id', this.id);
+    this.sendMsg('startPairing', {timeout: timeoutSeconds});
```
hobinjk: As above, all strings should be constants where possible

src/adapters/plugin/ipc.js
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      var data = JSON.parse(bufStr);
+    } catch (err) {
+      console.error('Error parsing message as JSON');
+      console.error('Rcvd: "' + bufStr + '"');
+      console.error(err);
+      return;
+    }
+    DEBUG && console.log('IPC:', this.name, 'Rcvd:', data);
+    var deferredReply = this.deferredReply;
```
hobinjk: Depending on how many requests we can have at once, I think it might be good to replace deferredReply with a map from id to Deferred

dhylands:
@hobinjk I think I finally got the build to go green.

The hanging problem was caused by not closing the nanomsg sockets. My solution wound up being to add an unloadAdapter message which the adapter responds to with adapterUnloaded and then the adapter closes the sockets on its end.

The gateway side keeps track of open adapters and closes stuff as it goes. I had some changes where I made PluginServer be an Adapter, but I wound up backing those changes out (so you can ignore that change).

The very first commit is the one you would have seen previously. All of the remaining ones are changes I made since then.

The procedure for loading a plugin is now that the plugin needs a config entry with plugin: true set. You can then run
```
node adapter-loader.js test_plugin
```
to load the test_plugin via the config file.

src/adapters/plugin/plugin-server.js
```diff
@@ -0,0 +1,88 @@
...
+    return adapter;
+  }
+
+  unregisterAdapter(adapterId) {
+    this.adapters.delete(adapterId);
+    if (this.adapters.size == 0) {
+      this.ipcSocket.close();
+      setTimeout(() => {
+        // This delay allows some of the console logs to 
+        // show up, and allows the tests to run.
```
hobinjk: This timeout doesn't seem to be doing anything with its empty body

src/adapters/plugin/property-proxy.js
```diff
@@ -0,0 +1,91 @@
...
+   * the value passed in.
+   */
+  setValue(value) {
+    return new Promise(resolve => {
+      console.log('PropertyProxy: setProperty property:', this.name,
+                  'for:', this.device.id,
+                  'to value:', value);
+
+      this.device.adapter.sendMsg(
+        'setProperty', {
```
hobinjk: nit: should be constant

config/default.js
```diff
@@ -28,6 +28,11 @@ module.exports = {
         },
       }
     },
+    test_plugin: {
```
hobinjk: Is this development-only code that should be removed?

src/adapters/adapter-constants.js
```diff
@@ -0,0 +1 @@
+../constants.js
```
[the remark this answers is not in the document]
dhylands: It has to do with webpack. constants.js winds up being included in the webpack and not as a separate file. Basically, everything from the adapter and its dependencies don't get processed by webpack, and when the adapter gets loaded, there is no constants.js file (it's merged into the build/gateway.js file). The simplest workaround seemed to be to add the adapter-constants.js and making it a symlink seemed to work. I think that the "ideal" solution would be to figure out how to redirect the require to grab the constants from the gateway.js webpack, I just haven't figure out how to do that. I'll file an issue so that we don't forget.

src/adapters/test-plugin/index.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * index.js - Loads the Test adapter.
```
hobinjk: I think the test-plugin-adapter should be renamed to `example-plugin-adapter` so that its purpose is clearer

src/adapters/adapter-constants.js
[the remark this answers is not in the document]
dhylands: I created #308

src/adapters/plugin/plugin-server.js
dhylands: Except that if I remove it, then when I run npm test locally, random tests fail.

src/adapters/test-plugin/index.js
dhylands: Sounds good.

src/test/common.js
```diff
@@ -49,25 +49,31 @@ let testId = uuid();
 let storageDir = config.get('settings.directory') + '-' + testId;
 
 beforeAll(async () => {
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
```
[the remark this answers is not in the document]
dhylands: I'll check and see if its still needed. The root cause was that the nanomsg sockets needed to be closed or jest would wind up hanging because it detected "resources were still in use".

src/constants.js
```diff
@@ -29,10 +29,36 @@ exports.RULES_PATH = '/rules';
...
 
-exports.SET_PROPERTY = 'setProperty';
-exports.REQUEST_ACTION = 'requestAction';
-exports.ADD_EVENT_SUBSCRIPTION = 'addEventSubscription';
-exports.PROPERTY_STATUS = 'propertyStatus';
+exports.ADAPTERS_CONFIG = 'adapters';
+
 exports.ACTION_STATUS = 'actionStatus';
+exports.ADAPTER_ADDED = 'adapterAdded';
+exports.ADAPYTER_UNLOADED = 'adapterUnloaded';
```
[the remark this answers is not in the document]
dhylands: Hmm. I wonder how that doesn't cause an error? Or maybe its causing an unreported error.

src/adapters/zwave/zwave-adapter.js
```diff
@@ -18,10 +18,10 @@ var ZWaveModule = require('openzwave-shared');
 const DEBUG = false;
 
 class ZWaveAdapter extends Adapter {
-  constructor(adapterManager, port) {
+  constructor(adapterManager, adapterId, port) {
     // We don't know our id yet. So we set it to ??? and once we get the
     // driverReady notification, which will then set the id based on the homeId.
```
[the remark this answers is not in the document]
dhylands: Turns out that the comment is actually correct, and I need to switch it back to using something like ??? or ***unknown***. ZWave and ZigBee both support multiple adapters and the adapter id isn't known until after the dongle is opened and some info is read from the dongle.

src/adapters/plugin/plugin-client.js
```diff
@@ -0,0 +1,106 @@
...
+    this.verbose &&
+      console.log('ManagerIpcAddr =', this.managerIpcAddr);
+    this.managerIpcSocket = new IpcSocket('PluginClientServer', 'req',
+                                          this.onManagerMsg.bind(this));
+    this.managerIpcSocket.connect(this.managerIpcAddr);
+
+    // Register ourselves with the server
+    this.verbose &&
+      console.log('Registering with server');
+
```
hobinjk: nit: trailing whitespace

src/controllers/things_controller.js
```diff
@@ -200,11 +200,12 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
     }));
   }
 
-  AdapterManager.on('property-changed', onPropertyChanged);
+  AdapterManager.on(Constants.PROPERTY_CHANGED, onPropertyChanged);
   Actions.on(Constants.ACTION_STATUS, onActionStatus);
 
   websocket.on('close', function() {
-    AdapterManager.removeListener('property-changed', onPropertyChanged);
+    AdapterManager.removeListener(Constants.PROPERTY_CHANGED,
```
hobinjk: nit: trailing whitespace

src/adapters/plugin/plugin-server.js
dhylands: I was able to get the intermittent failures with the timeout removed, so I'm going to go ahead and remove it.

PR #292, opened 2017-09-26
Fix storage init

dhylands:
I had to temporarily disable the rules-engine index-test.js since its now failing 100% of the time.

Does it require storage to be cleared when storage.init is called?


hobinjk:
See #294. The fact that the rules engine test still intermittently fails with this approach means that it unfortunately doesn't fix the root of the problem

dhylands:
Closing in favor of #296

PR #294, opened 2017-09-26
Mock node-persist

hobinjk:
Jest runs all tests in parallel. If any test clears `node-persist` while
another test is initializing it, the two calls can race and cause the
initialization of `node-persist` to see an inconsistent filesystem
state.

This mocks `node-persist` with an in-memory store to fix the race
conditions.

### DIFF
--- a/__mocks__/node-persist.js
+++ b/__mocks__/node-persist.js
@@ -1,0 +1,33 @@
+const storage = jest.genMockFromModule('node-persist');
+
+let data = {};
+
+storage.init = function() {
+  return Promise.resolve();
+};
+
+storage.clear = function() {
+  data = {};
+  return Promise.resolve();
+};
+
+storage.getItem = function(key) {
+  if (!data.hasOwnProperty(key)) {
+    return Promise.resolve();
+  }
+  // Duplicate node-persist's copying logic
+  let copiedValue = JSON.parse(JSON.stringify(data[key]));
+  return Promise.resolve(copiedValue);
+};
+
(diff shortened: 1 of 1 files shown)


__mocks__/node-persist.js
```diff
@@ -0,0 +1,33 @@
+const storage = jest.genMockFromModule('node-persist');
+
+let data = {};
+
+storage.init = function() {
```
dhylands: I noticed that some places pass in options when they call init. If they're all just to specify a path, then it might be ok to ignore it.

I suppose another way to deal with this would be to have each test use a different storage file.

__mocks__/node-persist.js
```diff
@@ -0,0 +1,33 @@
...
+  return Promise.resolve();
+};
+
+storage.clear = function() {
+  data = {};
```
dhylands: Isn't this still going to cause problems when test1 clears storage and test2 had written something to storage and was expecting it to be there?

hobinjk: Jest runs each test in its own node process (approximately) so each test gets its own mock storage

__mocks__/node-persist.js
hobinjk: Yeah, this is based on the fact that every use of init uses the same directory. If there were two calls to init with different directories whichever call occurred later would overwrite the first directory for that particular instance of Node.

An alternative approach to this PR would be to have every test use `test-settings/{randomly-generated-id}` as its directory instead of `test-settings`

PR #303, opened 2017-09-28
Add support for Thing actions and events

### LINKED ISSUE #187: Add event subscription over WebSocket
As a developer I want to add a subscription to an event over a websocket so that I can be notified when an event occurs without having to poll the server

### LINKED ISSUE #188: Receive event over WebSocket
As a developer I want to receive an event over a WebSocket so I can be notified when an event occurs without having to poll the server

### LINKED ISSUE #190: Receive action status over WebSocket
As a developer I want to be notified of a change in state of an action over a WebSocket so that I don't have to poll the server to know when my requested action has been completed

### LINKED ISSUE #237: Implement API for thing actions
As a complement to #189 and #190 it should be possible to create thing-specific actions via an API. Ideally it can reuse parts of the existing gateway actions model.

### LINKED ISSUE #238: Implement API for thing events
For #188 and #187 there needs to be support for creating and managing events in the Thing model

hobinjk:
Close #238, close #237, close #190, close #188, close #187

### DIFF
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -159,7 +159,15 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
       thing.removeEventSubscription(onEvent);
     });
   }).catch(function() {
-    console.warn('WebSocket opened on nonexistent thing', thingId);
+    console.error('WebSocket opened on nonexistent thing', thingId);
+    websocket.send(JSON.stringify({
+      messageType: Constants.ERROR,
+      data: {
+        status: '404 Not Found',
+        message: 'Thing ' + thingId + ' not found',
+      }
+    }));
+    websocket.close();
   });
 
   function onPropertyChanged(property) {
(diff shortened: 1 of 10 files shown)


src/controllers/things_controller.js
```diff
@@ -204,22 +229,31 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
...
-          }
-        }));
-      });
+
+    switch (request.messageType) {
+      case Constants.SET_PROPERTY: {
+        let setRequests = Object.keys(request.data).map(property => {
+          let value = request.data[property];
+          return device.setProperty(property, value);
+        });
```
dhylands: I don't quite understand what this is trying to do?
Is it just setting multiple properties at the same time for a single device?

hobinjk: Yes, given a request of the form
```javascript
{
  messageType: 'setProperty',
  data: {
    on: true,
    hue: 120
  }
}
```
it will set both properties

hobinjk:
@dhylands Can you review the latest changes to this?

PR #304, opened 2017-09-29
Lower bcrypt rounds in tests

hobinjk:
This keeps the same security in the default configuration while reducing the amount of time required to run the tests by 1/3.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -62,5 +62,5 @@ module.exports = {
     pagekite_cmd: './pagekite.py',
     port: 443
   },
-  bcryptRounds: 10
+  bcryptRounds: 0
 };
--- a/config/default.js
+++ b/config/default.js
@@ -61,5 +61,6 @@ module.exports = {
     domain: 'mozilla-iot.org',
     pagekite_cmd: './pagekite.py',
     port: 443
-  }
+  },
+  bcryptRounds: 10
 };
--- a/config/default.js
+++ b/config/default.js
@@ -61,6 +61,5 @@ module.exports = {
     domain: 'mozilla-iot.org',
     pagekite_cmd: './pagekite.py',
     port: 443
-  },
-  bcryptRounds: 0
+  }
 };
--- a/src/passwords.js
+++ b/src/passwords.js
@@ -7,6 +7,9 @@
  */
 
 const bcrypt = require('bcryptjs');
+const config = require('config');
+
+const rounds = config.get('bcryptRounds');
 
 module.exports = {
   /**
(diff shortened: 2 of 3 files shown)


config/default.js
```diff
@@ -61,5 +61,6 @@ module.exports = {
     domain: 'mozilla-iot.org',
     pagekite_cmd: './pagekite.py',
     port: 443
-  }
+  },
+  bcryptRounds: 10
```
dhylands: I think it would be better to allow this to not be specified and just use the default. I know that generally the default changes every couple years so that bcrypt takes approx the same amount of time with new compute hardware. By explicitly setting this, we'll eventually fall behind.

src/passwords.js
```diff
@@ -15,7 +18,7 @@ module.exports = {
    * @return {Promise<String>} hashed password
    */
   hash: function(password) {
-    return bcrypt.hash(password, 10);
+    return bcrypt.hash(password, rounds);
```
dhylands: According to the documentation, the 'rounds' argument is actually the length of the salt and not the number of rounds.
https://github.com/dcodeIO/bcrypt.js#hashs-salt-callback-progresscallback

genSalt is the function that takes the number of rounds.

src/passwords.js
```diff
@@ -26,7 +29,7 @@ module.exports = {
    * @return {String} hashed password
    */
   hashSync: function(password) {
-    return bcrypt.hashSync(password, 10);
+    return bcrypt.hashSync(password, rounds);
```
dhylands: Same as previous comment

src/passwords.js
hobinjk: The documentation disagrees with their implementation. The second argument is fed to genSalt as the number of rounds.

dhylands: I see that looking at the implementation now. I still that we should pass in a rounds of undefined and let the library pick the default rather than us providing it.

hobinjk: I agree that leaving it up to the library is a good idea. I can look for test performance gains elsewhere

dhylands: I don't have an issue with lowering the number of rounds for testing purposes. I just didn't like the idea of specifying it explicitly as our default (so in my opinion the unpatched version is flawed).

hobinjk: I've updated this PR setting bcryptRounds to `0` which will cause the library to default to 10. Unfortunately `bcrypt.hash(password)` and `bcrypt.hash(password, null)` don't work with bcrypt.js and `bcrypt.hash(password, undefined)` causes `config` to complain

dhylands:
Can't you use something like:
```js
var rounds;
if (config.has('bcryptRounds')) {
  rounds = config.get('bcryptRounds');
}
```

hobinjk:
@dhylands Done! I really like that approach

PR #306, opened 2017-09-29
Add generate-release.sh

hobinjk:


tools/generate-release.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Generates release archives based on the current Gateway revision
+
+if [ ! -e "package.json" ]; then
+  echo "Usage: ./tools/generate-release.sh"
+  echo "Error: missing package.json"
+  exit 1
+fi
+
+yarn
```
dhylands: I'm confused. Wouldn't you want to run yarn on the newly cloned gateway? And similarly, shouldn't the check for package.json be done on the cloned copy?
Cloning will check out the latest which may not match with the current version in the tree that the script is running in.

hobinjk: `git clone ./ gateway` means clone a clean copy of the current tree into a new folder named gateway. Most importantly, this means that if I run this while checked out to a branch, the newly created clone will already be checkout out to that branch but without any modifications or untracked files.

Yarn could be run in either directory, but running it in the original gateway means that it is probably only verifying that packages exist instead of building everything from scratch.

dhylands: Doh - That makes sense. I didn't actually appreciate at all of the arguments to git clone. Makes perfect sense now.

PR #313, opened 2017-10-05
Adds user-visible widgets for creating IFTTT rules

hobinjk:
Now you can just drag-and-drop in IFTTT hooks and events! To use events (the IFTTT effect), you will need to specify iftttKey in config/default.js

iftt-web-request

ifttt-rule


benfrancis:
(Maybe we can demo this but not land it in the gateway)

hobinjk:
Closing since the way we want to do this is with a first-party IFTTT integration instead of hacking things together

PR #325, opened 2017-10-10
Rules Mobile Friendliness

### LINKED ISSUE #323: Rules UI doesn't scale well on mobile
The rules list doesn't really fit on the screen on mobile.

screenshot_20171010-161406

hobinjk:
Improves the scaling of the rules engine on tiny (320 x 480) devices. Fix #323. This is based on #313 because it has all the new class names and widgets

### DIFF
--- a/static/css/rule.css
+++ b/static/css/rule.css
@@ -342,7 +342,7 @@
   display: none;
 }
 
-@media only screen and (max-width: 700px) and (min-height: 600px) {
+@media only screen and (max-width: 700px) and (min-height: 400px) {
   #drag-hint {
     flex-direction: column;
   }
--- a/static/css/rules-common.css
+++ b/static/css/rules-common.css
@@ -31,6 +31,8 @@
 .device-icon {
   margin: 2.7rem;
   padding: 0;
+  width: 4.8rem;
+  height: 4.8rem;
 }
 
 .rule-edit-button, .rule-delete-cancel-button, .rule-delete-confirm-button {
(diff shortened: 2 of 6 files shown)


benfrancis:
@hobinjk When you say "based on", does this require the commits from #313 or can it be landed independently?

hobinjk:
@benfrancis It was originally built on top of #313 but I changed that and it is now completely separate

benfrancis:
Screenshots below.

Portrait:
screenshot_20171012-160859

Landscape:
screenshot_20171012-160904

hobinjk:
Thanks for the feedback! I agree that my approach was a bit too hacky. I've changed this to be a bit cleaner. It no longer adjusts the base font size. Instead, it puts the rules cards into a centered grid which scales well down to around 400px width. On smaller widths the only thing cut off is the enable/disable switch.

I would consider the ideal to be the cards dynamically becoming taller and skinnier while still appearing in a vertical list.

hobinjk:
It now looks like this. The first rule will be cut off on a sufficiently skinny phone (like a pixel), but that isn't a significant problem in my opinion.
screen shot 2017-10-12 at 12 53 20

screen shot 2017-10-12 at 12 53 25

benfrancis:
Yeah this is better, thanks. One slight issue is that with 3 or more cards the + button can obscure the enable/disable switch on the final card and it's impossible to scroll far enough to reach it so you have to try to squeeze the switch between the + button and the bottom of the screen to make it clickable!. Maybe a bit of padding at the bottom would fix that.

hobinjk:
@benfrancis Added padding to the bottom to get them to not overlap

benfrancis:
\o/ LGTM.

PR #327, opened 2017-10-11
Add support for multiple adapters per plugin

dhylands:


dhylands:
Tested with the zwave adapter (although I only have one dongle).
This also gets rid of all of the test issues that I had by using inproc IPC rather than ipc (which uses named pipes).

src/adapters/plugin/adapter-manager-proxy.js
```diff
@@ -99,6 +102,10 @@ class AdapterManagerProxy extends EventEmitter {
         this.pluginClient.sendNotification(Constants.ADAPTER_UNLOADED, {
           adapterId: adapter.id,
         });
+        return;
+
+      case Constants.UNLOAD_PLUGIN:
```
hobinjk: Can you add this message to the Adapter IPC documentation in the wiki?

src/adapters/plugin/adapter-proxy.js
```diff
@@ -66,98 +55,26 @@ class AdapterProxy extends Adapter {
...
+   *
+   * @returns a promise which resolves when the adapter has
+   *          finished unloading.
+   */
+  unload() {
+    if (this.deferredUnload) {
+      console.error('AdapterProxy: unload already in progress');
+      return Promise.reject();
+    }
+    this.deferredUnload = new Deferred;
```
hobinjk: nit: Should have parentheses even if they're optional

src/adapters/plugin/adapter-proxy.js
```diff
@@ -206,6 +123,10 @@ class AdapterProxy extends Adapter {
       return Promise.reject(err);
     }
     this.deferredMock = new Deferred();
+
+    // We need the actual device object when we recolve the promise
```
hobinjk: nit: resolve

src/app-instance.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * app-timestamp.js
```
hobinjk: Should this file be named app-timestamp.js then?

src/app-instance.js
```diff
@@ -0,0 +1,44 @@
...
+// the purpose of this module.
+
+'use strict';
+
+const process = require('process');
+
+class AppInstance {
+
+  constructor() {
+    this.timestamp = this.mark();
```
hobinjk: mark() doesn't return anything, so timestamp will be initialized to undefined. If this is fixed, the call to `mark()` in app.js can be removed

src/adapters/plugin/adapter-manager-proxy.js
dhylands: Yep - was planning on doing that.

src/app-instance.js
dhylands: I had originally called it app-timestamp and decided to call it app-instance. The fact that it uses a timestamp is really just an implementation detail. I'll fix the comment.

src/app-instance.js
dhylands: Doh

Issue #319, opened 2017-10-06
HTTP not being re-directed to HTTPS


benfrancis:
STR:
* Set up a new gateway with a foo.mozilla-iot.org subdomain
* Type http://foo.mozilla-iot.org into a browser tab

Expected:
* Redirects to https://foo.mozilla-iot.org

Actual:
* Blank page

hobinjk:
I've tracked this down to how we're calling Pagekite. It is currently invoked as
```
./pagekite.py --clean --frontend=tunneltoken.api.mozilla-iot.org:443 --service_on=https:tunneltoken.api.mozilla.org:localhost:4443:moziot
```
I think we need another backend definition to also forward HTTP traffic but I don't actually know how Pagekite works. @andrenatal do you know how to fix this?

andrenatal:
@hobinjk yes, let me take a look

benfrancis:
@andrenatal Do you expect a fix to be on the cloud side or the gateway side? This is the last issue blocking the 0.2.1 release, should we wait before rolling the OS 0.2.1 image and putting that out as our next public release?

andrenatal:
I think both. I was waiting for the demo to fix that to avoid breaking things. I'll fix it today.

andrenatal:
Ok, I implemented the proper pagekite redirection in the cloud (https://github.com/mozilla-iot/registration_server/pull/21) and in the gateway (https://github.com/mozilla-iot/gateway/pull/341) and now requests to port `80` via the tunnelservice are being redirected to `https`.  The patch was already deployed into production and staging. 

My only comment is that here https://github.com/mozilla-iot/gateway/blob/master/src/app.js#L206  requests to `80/8080` are being redirected to `4443` since the local http server in the gateway listens to such port, so for that reason I needed to add another docker redirect from this port to `443`.

andrenatal:
Fixed by https://github.com/mozilla-iot/gateway/pull/341

PR #341, opened 2017-10-19
Implementing redirection from http to https when accessing by the tunnel

andrenatal:


andrenatal:
Fixes #319

hobinjk:
These changes do cause gateways at 0.2.0 and below to fail to register via LE which I think is just something we need to keep in mind

andrenatal:
BTW, now we have `mozilla-iot-staging.org` running as staging. You can use it to dev/test to avoid keep creating production certs on prod, for example.

andrenatal:
@hobinjk I'll revert this PR since the nginx approached worked better and we don't need pagekite for port 80 anymore

PR #343, opened 2017-10-19
Support redirect behind forwarding

hobinjk:
By default localhost:8080 (which localhost:80 forwards to on the RPi) willl now redirect to localhost:443 instead of localhost:4443.

Note that this may require some config editing in local setups

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -16,6 +16,9 @@ module.exports = {
     https: 4443,
     http:  8080
   },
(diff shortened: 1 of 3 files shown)


dhylands:
LGTM. Perhaps a bit more description. It would be good to know how this impacts the URL

hobinjk:
Thanks! I added a few comments about the code's effect

PR #345, opened 2017-10-20
Fixes #332 - fixes GPIO adapter loading

### LINKED ISSUE #332: GPIO not fully supported in 0.2.0
I'd like to use GPIO-connected devices & sensors with Gateway and noticed a GPIO-based LED defined in the default config file but not enabled.  When I enable it, Gateway fails to start with the following error in run-app.log:

/home/pi/mozilla-iot/gateway/build/adapters/gpio/gpio-adapter.js:111
    for (var pin in gpioConfig.pins) {
                              ^
TypeError: Cannot read property 'pins' of undefined
    at new GpioAdapter (/home/pi/mozilla-iot/gateway/build/adapters/gpio/gpio-adapter.js:111:31)

GPIO-attached devices aren't discoverable directly from the Gateway UI so adding them via the command line is the only option AFAIK.

dhylands:


dhylands:
r=me since this was a trivial bug that I should have caught when introducing the Adapter-IPC code.

PR #352, opened 2017-10-24
Better LE messaging

andrenatal:


static/js/setup_subdomain.js
```diff
@@ -44,7 +44,13 @@ createDomainButton.addEventListener('click', function() {
             'Content-Type': 'application/json'
         }
     }).then(function (response) {
-        return response.text();
+        if (response.statusText.indexOf('UnavailableName') > -1){
+            throw new
+                Error('This subdomain is already being used.' +
```
hobinjk: There should be a space after the period in this string

andrenatal: fixed

PR #357, opened 2017-10-25
Use yarn for initial installation of packages

### LINKED ISSUE #309: Figure out how to use yarn to cross compile
The build script used to create the images for the Raspberry Pi uses the command:
(cd ${GATEWAY}; npm --arch=${ARCH} install --target_arch=arm)
See: https://github.com/mozilla-iot/gateway/blob/48f655c1c9d9a8f09f9824b5c1dce6e0a242922b/image/build-gateway.sh#L65
from within a docker image.

This means that we're using npm install rather than yarn which means that we don't have a reproducible image.

We need to figure out how to use yarn and get the cross compilation portion to work.

hobinjk:
The script then switches over to npm for the actual building of the
package. See https://github.com/yarnpkg/yarn/issues/2221 for
inspiration.

Fix #309

### DIFF
--- a/image/build-gateway.sh
+++ b/image/build-gateway.sh
@@ -61,8 +61,10 @@ export NVM_DIR="$HOME/.nvm"
 nvm install ${NODE_VERSION}
 nvm use ${NODE_VERSION}
 
+npm install -g yarn
+
 # Build the node modules, cross compiling any native code.
-(cd ${GATEWAY}; npm --arch=${ARCH} install --target_arch=arm)
+(cd ${GATEWAY}; yarn --ignore-scripts; npm rebuild --arch=${ARCH} --target_arch=arm)
 
 set -x
 


hobinjk:
@dhylands The main thing I'm not sure of in this PR is whether `npm install -g yarn` needs to be run as root

image/build-gateway.sh
```diff
@@ -61,8 +61,10 @@ export NVM_DIR="$HOME/.nvm"
 nvm install ${NODE_VERSION}
 nvm use ${NODE_VERSION}
 
+npm install -g yarn
+
 # Build the node modules, cross compiling any native code.
-(cd ${GATEWAY}; npm --arch=${ARCH} install --target_arch=arm)
+(cd ${GATEWAY}; yarn --ignore-scripts; npm rebuild --arch=${ARCH} --target_arch=arm)
```
dhylands: sweet. My recollection is that I have to run the rpxc command which runs this script using sudo, so an additional sudo shouldn't be required.

PR #359, opened 2017-10-26
call the setemail api to associate user's email with the subdomain

andrenatal:


hobinjk:
This gateway-side part looks good! However, I think there are some issues on
the registration_server side. It looks like there's a bug where upon setting
the email it sends out a message with a [link pointing to `/confirmemail`](https://github.com/mozilla-iot/registration_server/blob/master/server/src/email_routes.rs#L142)
instead of [`/verifyemail`](https://github.com/mozilla-iot/registration_server/blob/master/server/src/email_routes.rs#L169).

andrenatal:
@hobinjk thanks. I edited the email template on staging and production to remove the link in the confirmation email sent by the registration_server so we can land in parts to have at least the gateway patch landed for the next image that's going to be built tomorrow I think. I'll merge this then so you can just pull, point to staging, create a new subdomain and see the email coming.

andrenatal:
@hobinjk I filed a new issue to track that: https://github.com/mozilla-iot/registration_server/issues/33

PR #362, opened 2017-10-30
Point pyconf dependency at mozilla-iot fork.

mrstegeman:
The upstream version writes out invalid certbot renewal configs. Switch to the fork so we can fix that.


benfrancis:
Cool. @mrstegeman Did we send a pull request against the upstream version?

mrstegeman:
Not yet. I'm still working through a few issues, but then I can. I don't believe this specific library has been touched in 2 years, though, so who knows if they'll take patches.

PR #363, opened 2017-10-30
Write certbot configs to filesystem for later use.

mrstegeman:
These need to be in the filesystem in a proper layout to allow `certbot` to renew certificates.


src/controllers/settings_controller.js
```diff
@@ -80,9 +80,10 @@ SettingsController.post('/subscribe', function (request, response) {
     // function to automatically agree and accept the ToS
     function leAgree(opts, agreeCb) { agreeCb(null, opts.tosUrl); }
     let leStore = require('le-store-certbot').create({
-        configDir: null,
-        debug: true,
-        logsDir: null
+        configDir: '~/mozilla-iot/etc',
```
hobinjk: This differs from the definition in #364

mrstegeman: It's back in line now. This is the proper directory.

PR #364, opened 2017-10-30
Add automatic cert renewal systemd service.

mrstegeman:


tools/renew-certificates.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+gateway_dir="/home/pi/mozilla-iot/gateway"
+
+certbot renew \
+    --config-dir "$gateway_dir/etc" \
```
hobinjk: Is there configuration that we need to specify?

tools/renew-certificates.sh
```diff
@@ -0,0 +1,9 @@
...
+    --logs-dir "$gateway_dir/var/log" \
```
hobinjk: I'd prefer to not top-level `etc` and `var` directories into the gateway if we don't have to. Alternatively, they could be added to .gitignore

mrstegeman:
I'm thinking this may not actually work, since we're using DNS challenge... I need to find a way to test renewal.

tools/renew-certificates.sh
mrstegeman: certbot will try to auto-renew any certificates it finds in the given config directory. When we generate certs, those are now written out to the filesystem, including the renewal config. See #363.

mrstegeman:
Do not merge yet, still working through issues.

PR #367, opened 2017-10-30
Update yarn.lock for forked libs.

mrstegeman:


hobinjk:
It looks like the better way to do the `resolutions: {}` in package.json would be to specify the full commit hash instead of `master`. That way yarn won't fail if there's an update to one of our forked dependencies that we don't want to deploy immediately

hobinjk:
(it's definitely more of a pain to update, unfortunately)

PR #371, opened 2017-10-31
Update le-store-certbot and pyconf module pointers.

mrstegeman:


package.json
```diff
@@ -96,7 +96,7 @@
     "testEnvironment": "node"
   },
   "resolutions": {
-    "le-store-certbot": "https://github.com/mozilla-iot/le-store-certbot/tarball/master",
-    "pyconf": "https://github.com/mozilla-iot/node-config-python/tarball/master"
+    "le-store-certbot": "https://github.com/mozilla-iot/le-store-certbot/tarball/master/82b00d5d9d0e6b0b025203d11aedb34ca4803b92",
+    "pyconf": "https://github.com/mozilla-iot/node-config-python/tarball/master/254eb3beda4ec883738b322c6d1cdedd2a75c644"
```
hobinjk: I think this should just be https://github.com/mozilla-iot/node-config-python/tarball/254eb3beda4ec883738b322c6d1cdedd2a75c644 because the `/master` might become inaccurate

PR #373, opened 2017-10-31
Update dependencies

hobinjk:


hobinjk:
@dhylands The serialport upgrade might cause some spooky things to happen

dhylands:
This seems to break the ZigBee adapter. If I run on master, then everything works fine. However, if I checkout PR 373 do rm -rf node_modules followed by yarn, then when I run the gateway, I see this:
```
No Digi port found
```
So this PR will need some additional changes. I'll see if I can figure out what's going on.

hobinjk:
According to [the upgrade guide](https://github.com/node-serialport/node-serialport/blob/v6.0.4/UPGRADE_GUIDE.md) "SerialPort.list now has more consistent output across all platforms. This means the data in the objects may be different than 4x. Notably the path on OSX now returns the tty instead of the cu path and the 0x has been removed from values on linux and osx."

I'll update the PR accordingly, I missed the 0x removal part the first time around. I'll also test on my gateway with zwave

dhylands:
I also had to fixup how the zigbee parser gets hooked up. 
See: https://github.com/mozilla-iot/gateway/pull/384/commits/bea00fd0dd8e3e8200f87a3cc562b708feabec02
I tested with both ZigBee and ZWave under linux. I'll test under OSX as well.

hobinjk:
Closing in favor of #384

Issue #374, opened 2017-11-01
Gateway sdcard image will only run on RPi 3


dhylands:
If you try to boot up the sdcard image on a Raspberry Pi earlier than the Pi 3 then the gateway won't load.

It turns out that the base image needs to be built on an earlier version of the RPi, and the build-image.sh script needs to specify armv6l rather than armv7l.

Building the base image on an RPi Model B+ I was able to run the gateway on the B+ and on a PiZero.

mrstegeman:
@dhylands I believe this is fixed now, correct?

dhylands:
Yes - this was fixed by https://github.com/mozilla-iot/rpi-image-builder/commit/867ab35845f8ef50f5121e0a3c4e078c1e21f174
Also see #380 for further details.

PR #375, opened 2017-11-01
Work on more RPi's (other than the RPi3)

dhylands:


dhylands:
Fixes #374 
With this change and building the base image on an RPi 2 (Model B+) I was able to load and run the gateway on the following RPI's:
RPi 3
RPi 2 (B+) - used a WiFi dongle
PiZero (the non-wifi one) - used a WiFi dongle

PR #383, opened 2017-11-01
Enable Floorplan and Rules by default.

### LINKED ISSUE #382: Enable Rules Engine & Floorplan by default
As a user I want the rules engine and floorplan experiments to be enabled by default so I don't have to enable them manually

Leave speech UI disabled by default for now while we develop it further

mrstegeman:
Fixes #382

### DIFF
--- a/src/db.js
+++ b/src/db.js
@@ -113,15 +113,54 @@ var Database = {
     var defaultUser = config.get('authentication.defaultUser');
     if (defaultUser) {
       var passwordHash = Passwords.hashSync(defaultUser.password);
-      this.db.run('INSERT INTO users (email, password, name) VALUES (?, ?, ?)',
+      this.db.run(
+        'INSERT INTO users (email, password, name) VALUES (?, ?, ?)',
         [defaultUser.email, passwordHash, defaultUser.name],
         function(error) {
-        if (error) {
-          console.error('Failed to save default user.');
+          if (error) {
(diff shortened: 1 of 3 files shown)


hobinjk:
I think the intent is to have them enabled by default but still disable-able
@benfrancis ?

mrstegeman:
So they're still intended to be experiments for now, then?

benfrancis:
> I think the intent is to have them enabled by default but still disable-able

Correct, sorry if that wasn't clear.

mrstegeman:
This version does the trick.

src/db.js
```diff
@@ -113,15 +113,54 @@ var Database = {
...
+        if (baseKey !== '') {
+          newKey = baseKey + '.' + key;
         } else {
-          console.log('Saved default user ' + defaultUser.email);
+          newKey = key;
         }
-      });
+        
+        if (typeof obj[key] === 'object') {
+          Array.prototype.push.apply(settings,
```
hobinjk: `settings = settings.concat(generateSettings(obj[key], newKey))` or `settings.push(...generateSettings(obj[key], newKey))` would be a bit cleaner than push.apply

mrstegeman: I specifically avoided settings.concat to avoid copying the array unnecessarily. I can switch to spread syntax, though.

hobinjk:
Thanks! LGTM

PR #384, opened 2017-11-01
New serialport

dhylands:


hobinjk:
I'm testing this locally and it LGTM so far, thanks!

dhylands:
ZWave doesn't seem to be working under OSX. Investigating. ZigBee seems to work fine.

dhylands:
Seems to be related to the difference between /dev/cu.usbmodemXXX and /dev/tty.usbmodemXXX

dhylands:
ZigBee and ZWave now work properly under OSX. - so this should be good to go now.

PR #386, opened 2017-11-03
Adapter rework

mrstegeman:
Load adapters dynamically from a directory. This will allow us to do things like installing and uninstalling arbitrary packages. In doing so, adapter info and settings were moved into a package.json file in their respective directories, which are then copied into the settings database. All adapters are disabled by default and must be explicitly enabled.


src/adapters/org.mozilla-iot.adapters.example-plugin/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "api": 1,
+    "id": "org.mozilla-iot.adapters.example-plugin",
+    "name": "Example Plugin",
+    "plugin": true,
+    "config": {
+    }
```
dhylands: Presumably we can remove the config from the manifest for adapters which don't use it.

src/adapters/org.mozilla-iot.adapters.gpio/index.js
```diff
@@ -10,7 +10,7 @@
 
 var fs = require('fs');
 
-function maybeLoadGpioAdapter(adapterManager, adapterId, adapterConfig) {
+function maybeLoadGpioAdapter(adapterManager, manifest) {
```
dhylands: We probably need a mechanism for the adapter to leave a message that indicates that it couldn't be loaded and why not.

For example, the gpio adapter will fail to load on Mac OSX and on most linux laptops, since they don't actually have a gpio subsystem.

If I enable gpio on my linux laptop, then all that happens is that there is a console message printed:
```
Not starting GPIO adapter - no permissions to /sys/class/gpio/export
```

src/db.js
```diff
@@ -161,6 +161,18 @@ var Database = {
           }
         });
     }
+
+    if (process.env.NODE_ENV === 'test') {
+      // Enable mock adapter if running tests.
+      this.db.run(
+        'INSERT INTO settings (key, value) VALUES (?, ?)',
+        ['adapters.org.mozilla-iot.adapters.mock.enabled', 1],
```
dhylands: Rather than hard-coding this, I think that we should look for a particular manifest value/entry to indicate that this is a "test" plugin.

src/adapter-manager.js
```diff
@@ -324,6 +333,102 @@ class AdapterManager extends EventEmitter {
...
+    // Verify API version.
+    const apiVersion = config.get('adapterManager.api');
+    if (manifest.api !== apiVersion) {
+      console.error('API mismatch for adapter:', manifest.id,
+                    '\nExpected:', apiVersion, ' Actual:', manifest.api);
+      return;
+    }
+
+    // If this adapter is not explicitly enabled, move on.
+    if (!manifest.enabled) {
```
dhylands: This looks wrong to me, because 'enabled' doesn't come from the manifest but rather from the settings.

I think it would be clearer to just have a local enabled variable which is initialized from the manifest (if you intended that) and overridden by the setting.

src/controllers/settings_controller.js
```diff
@@ -21,6 +21,7 @@ const fs = require('fs');
 const TunnelService = require('../ssltunnel');
```
dhylands: These changes seem to be unrelated to the adapter changes. Please move to a separate PR

static/js/settings.js
```diff
@@ -114,6 +119,71 @@ var SettingsScreen = {
...
+    fetch('/settings/adapters', opts).then(function (response) {
+      return response.json();
+    }).then(function (body) {
+      if (!body) {
+        return;
+      }
+
+      const ignored = [
+        'adapters.org.mozilla-iot.adapters.example-plugin',
+        'adapters.org.mozilla-iot.adapters.mock',
```
dhylands: I'd rather not see this hard coded. Why shouldn't the example plugin show up in the list (if it present)?

src/controllers/settings_controller.js
mrstegeman: I had planned on it, but this PR disables all plugins by default, so the gateway would be put into a useless state without the enablement code in place.

dhylands: I'd probably do this in 2 stages. Have ZigBee and ZWave still enabled by default - they will fail to load if the dongles are missing anyways - and then later on change zigbee and zwave to be off by default.

I guess I was seeing a bunch of changes related to cerbot and stuff and didn't understand how that fit in with the adapter changes.

dhylands:
I thought I would bring up terminology at this point as well.

In the UI, it shows "Adapters" as being "ZigBee Adapter" and "ZWave Adapter" and these are really plugins and not adapters. There will be an adapter for each ZigBee dongle which is plugged in. So I think in Settings we should call them Plugins, and not Adapters.

benfrancis:
@dhylands I'm not sure if I understand the relationship between a "plugin", an "adapter" and an "add-on". Perhaps this is something that could be documented in said design document :) I wasn't able to figure out from the [Adapter-IPC](https://github.com/mozilla-iot/wiki/wiki/Adapter-IPC) wiki page the distinction between an "adapter" and a "plugin" because the terms aren't defined, although I see there are separate "addAdapter" and "registerPlugin" messages and that adapter messages include both a pluginId and adapterId.

The next question is whether this distinction is something that actually needs representing to the user in the user interface or whether it's just an implementation detail, which is more of a UX question. Could they not just be listed as "ZigBee Adapter 1" and "ZigBee Adapter 2"?

So far my thinking was that "adapter" (adding support for a new protocol or device type) and "extension" (modifying the UI) would be two types of "add-on". How does a "plugin" differ from the user's point of view?

benfrancis:
Note the documentation we have so far on a potential add-on system design is here https://github.com/mozilla-iot/wiki/wiki/Add-ons-System-Design

mrstegeman:
Wikis updated:
https://github.com/mozilla-iot/wiki/wiki/Add-ons-System-Design
https://github.com/mozilla-iot/wiki/wiki/Adapter-API

dhylands:
A plugin is a collection of code. Currently each plugin only implements one type of adapter, but there is no reason why a plugin couldn't implement several types of adapters.

Currently, there will be an adapter for each ZigBee or ZWave dongle. You could have multiple dongles plugged into the same gateway. A more practical example where multiple adapters would be supported by the same adapter-type might be the case where you have multiple gateways. 

I'm not sure if there would be a distinction between an add-on and a plugin, although there might be. It's intended that plugins will run in their own process, but that's might not be an appropriate use of terminology.

benfrancis:
@dhylands wrote:
> A plugin is a collection of code.

OK, not the best definition ever... :)

> Currently each plugin only implements one type of adapter, but there is no reason why a plugin couldn't implement several types of adapters.

Why would you want to do that? Why not just have one adapter per add-on?

> Currently, there will be an adapter for each ZigBee or ZWave dongle. You could have multiple dongles plugged into the same gateway. A more practical example where multiple adapters would be supported by the same adapter-type might be the case where you have multiple gateways.

I'm struggling to understand the use case for having multiple ZigBee or Z-Wave dongles plugged into the same gateway. But even if that's an edge case we want to support, can't a single software adapter add-on support multiple physical dongles? Just call them Zigbee Adapter 1 and ZigBee Adapter 2?

> I'm not sure if there would be a distinction between an add-on and a plugin, although there might be. It's intended that plugins will run in their own process, but that's might not be an appropriate use of terminology.

The user shouldn't need to care about whether an add-on is running in a separate operating system process or not. They just care that they're adding an adapter add-on that supports a new type of device or protocol.

I still don't think I understand the distinction between a plugin and an adapter. What would we be losing if it was simplified to just adapters, where an adapter is a type of add-on? Note that the UI doesn't necessarily need to use the same terminology as the API, but I'm also not clear on the distinction between a plugin and an adapter at the API level either.

dhylands:
@benfrancis wrote:
> @dhylands wrote:
> 
>     Currently each plugin only implements one type of adapter, but there is no reason why a plugin couldn't implement several types of adapters.
> 
> Why would you want to do that? Why not just have one adapter per add-on?

You might wind up wind some hardware which combined something like a ZWave contoller and some additional functionality which was outside the ZWave functionality (let's say GPIO). So you might want to present the ZWave portion as a ZWave adapter and the GPIO portion as a GPIO adapter. It might not be possible to implement these as independent adapters if there was some shared hardware/resource.

If we want the community to be able to create adapters for their own custom hardware, we can't necessarily dictate the adapter model that would be used, especially when dealing custom hardware.

>     Currently, there will be an adapter for each ZigBee or ZWave dongle. You could have multiple dongles plugged into the same gateway. A more practical example where multiple adapters would be supported by the same adapter-type might be the case where you have multiple gateways.
> 
> I'm struggling to understand the use case for having multiple ZigBee or Z-Wave dongles plugged into the same gateway. But even if that's an edge case we want to support, can't a single software adapter add-on support multiple physical dongles? Just call them Zigbee Adapter 1 and ZigBee Adapter 2? I'm not sure what you mean by "multiple gateways".

ZWave has an upper limit to the number of devices which can exist on a given mesh network (around 232 IIRC).

> I still don't think I understand the distinction between a plugin and an adapter. What would we be losing if it was simplified to just adapters, where an adapter is a type of add-on? Note that the UI doesn't necessarily need to use the same terminology as the API, but I'm also not clear on the distinction between a plugin and an adapter at the API level either.

A plugin is an instance of the Plugin class. An adapter is an instance of the Adapter class. A plugin can have one or more adapters. It is assumed that all of the adapters running within a plugin run within the same process. Right now, we're using Plugin as if it were an AdapterType (i.e. ZigBee or ZWave), but there isn't any particular reason this needs to be done (as far as the code is concerned).

A plugin is the thing that the user will "install" into their gateway. A single plugin may instantiate multiple adapters - the realistic use case for this would be talking to multiple gateways.

benfrancis:
@dhylands Sorry, I think I must be missing something because I'm still struggling to understand the need for both plugins and adapters.

> You might wind up wind some hardware which combined something like a ZWave contoller and some additional functionality which was outside the ZWave functionality (let's say GPIO). So you might want to present the ZWave portion as a ZWave adapter and the GPIO portion as a GPIO adapter. It might not be possible to implement these as independent adapters if there was some shared hardware/resource.

OK, this seems like an unlikely example but if there was a device that provided both Z-Wave and GPIO over a serial port, why not create a single custom adapter for that device? If it required custom messages over the serial port then an existing Z-Wave adapter wouldn't work anyway. 

> If we want the community to be able to create adapters for their own custom hardware, we can't necessarily dictate the adapter model that would be used, especially when dealing custom hardware.

Well we have to dictate something. But it seems likely that any Arduino type custom hardware connected directly to the gateway is going to require a custom adapter too.

> ZWave has an upper limit to the number of devices which can exist on a given mesh network (around 232 IIRC).

I can see that an industrial application (not our current target) could reach the upper limit of the number of devices a Z-Wave dongle can connect to, but even then why not just support having multiple instances of a Z-Wave adapter?

> A plugin is an instance of the Plugin class. An adapter is an instance of the Adapter class. A plugin can have one or more adapters. It is assumed that all of the adapters running within a plugin run within the same process. Right now, we're using Plugin as if it were an AdapterType (i.e. ZigBee or ZWave), but there isn't any particular reason this needs to be done (as far as the code is concerned).
>
> A plugin (or perhaps add-on) is the thing that the user will "install" into their gateway. A single plugin may instantiate multiple adapters - the realistic use case for this would be talking to multiple gateways.

I still don't know what you mean by "talking to multiple gateways" and I'm still not understanding the need for a plugin that contains multiple adapters.

Maybe some more examples will help:
* ZigBee - can support standard device types from the ZigBee HA and ZLL profiles
* Z-Wave - can support standard Z-Wave device types
* Bluetooth - can support adopted smart home profiles when they emerge
* Custom ZigBee/Z-Wave/Bluetooth/etc. adapter to support a non-standard device type or profile
* Apple HomeKit adapter - to connect to HomeKit compatible devices over WiFi or Bluetooth
* Nest Weave adapter - to connect to "works with Nest" devices (e.g. over Thread)
* WeMo adapter - to connect to WeMo devices over WiFi
* ONVIF adapter - to connect to IP cameras following the ONVIF standard
* GPIO adapter - to directly control GPIO ports (e.g. on a Raspberry Pi)
* Custom Arduino device adapter - to communicate with a custom piece of hardware over a serial port or similar

It seems to me that all of these adapters could be implemented as a single adapter running in its own process, talking to the gateway process over IPC, no need for a "plugin". The only potential problem I can think of is if you already have for example a Bluetooth adapter that communicates over a serial port and supports some standard adopted smart home profiles but then the user wants to install an additional adapter to talk to a custom device over Bluetooth, then multiple adapters might be fighting over the same serial port. Is that what you're getting at? But that would require two adapters that share the same plugin process being distributed as two separate add-ons... which sounds problematic.

Hopefully we can agree on these basic concepts soon so we can move on to packaging and installing add-ons etc. :)

dhylands:
> I still don't know what you mean by "talking to multiple gateways" and I'm still not understanding the need for a plugin that contains multiple adapters.

We've talked about this in the past. This becomes more obvious once we support more types. Let's suppose I have 3 raspberry pi's. Once has Sensor A connected to some GPIO pins, and another has sensor B connected to some GPIO pins. Both are running gateway software.

Then we have a third rpi which I'll call the main RPi. It would have an adapter that knows about the other 2 gateways and presents the URLs to the UI.

Back to plugins, I treat a plugin as basically a mechanism to allow an adapter to communicate with the gateway through IPC. We want these plugins to run in a separate process. The adapter itself has no notion of how to talk via IPC.

Maybe from the user perspective adapters and plugins are really the same. But the point I was really trying to make is that under Settings (with this PR) we have something called Adapters which are really AdapterTypes. You'll notice that the things listed under Settings->Adapters have different names from the things listed under Adapters, and that's because the things shown under Settings->Adapters are really plugins or AdapterTypes. As a user, I would find it confusing that the things shown under adapters are different than the things shown under Settings->Adapters

mrstegeman:
Just to be clear, I pulled all of the UI work out of this PR. It will be in a separate PR after this is worked out and merged.

benfrancis:
>  But the point I was really trying to make is that under Settings (with this PR) we have something called Adapters which are really AdapterTypes.

OK, that suddenly makes more sense :)

The way I would suggest representing this in the UI is that you can install a "ZigBee adapter add-on" that adds support for "ZigBee adapters" to the gateway and that add-on can support multiple physical ZigBee dongles if necessary. So in a list of add-ons you'd just see one ZigBee adapter add-on but in a list of adapters you might see multiple ZigBee adapters. Does that make sense?

BTW, I had kind of assumed that now we have the IPC layer all adapters would be ported over to running in their own process and talking IPC. So a "plugin" wouldn't be a special type of adapter, all adapters would be "plugged in".

dhylands:
> BTW, I had kind of assumed that now we have the IPC layer all adapters would be ported over to running in their own process and talking IPC. So a "plugin" wouldn't be a special type of adapter, all adapters would be "plugged in".

Once PR #408 lands, then we can update the config and make ZigBee, ZWave, etc load as plugins. Loading as a plugin or not is just a config change.

mrstegeman:
This is now up to date with our latest discussions, with the add-ons packaged as npm packages. Please review  so we can move forward with the add-on management work.

src/adapters/moziot-adapter-example/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/mozilla-iot/gateway/issues"
+  },
+  "moziot": {
+    "api": {
+      "min": 1,
+      "max": 1
+    },
+    "plugin": true
```
benfrancis: What adapters are not plugins?

dhylands: Eventually, this can be removed. I put it there so that we could have the option of having adapters which were loaded with the gateway (as opposed to being in a separate process)

src/adapter-manager.js
```diff
@@ -326,17 +338,117 @@ class AdapterManager extends EventEmitter {
...
+    }
+
+    // Get any saved settings for this adapter.
+    const key = `adapters.${manifest.name}`;
+    let savedSettings = await Settings.get(key);
+    let newSettings = Object.assign({}, manifest);
+    if (savedSettings) {
+      // Overwrite config and enablement values.
+      newSettings.moziot.enabled = savedSettings.moziot.enabled;
+      newSettings.config = Object.assign(manifest.moziot.config || {},
```
dhylands: Shouldn't that be newSettings.moziot.config rather than newSettings.config?

src/adapter-manager.js
```diff
@@ -326,17 +338,117 @@ class AdapterManager extends EventEmitter {
...
+    if (manifest.moziot.plugin) {
+      if (manifest.moziot._test) {
+        // This is a special case where we load the adapter directly
+        // into the gateway, but we use IPC comms to talk to the
+        // adapter (i.e. for testing)
+        const pluginClient = new PluginClient(manifest.name,
+                                              {verbose: false});
+        pluginClient.register().then(adapterManagerProxy => {
+          console.log('Loading adapter', manifest.name, 'as plugin');
+          const adapterLoader = dynamicRequire(path.join(adapterPath));
```
dhylands: Does this call to path.join actually do anything?

mrstegeman:
@dhylands I fixed both issues in my enable/disable PR.

Issue #379, opened 2017-11-01
Re-claim subdomain


benfrancis:
As a user I want to re-claim a subdomain I previously registered so I don't lose my subdomain if I re-install or reset my gateway

mrstegeman:
I was thinking about an architecture something like the following:

- When gateway is first set up, they are presented with a login/register form.
  - If it's a new user, they will then set up a new domain as usual.
  - If it's an existing user, they are given the option of creating a new domain (again, as usual), or they can choose one they've previously registered. Their list of previously registered domains would ideally only include "inactive" ones, i.e. gateways that haven't pinged/registered with the registration server in some time, or at least we could show the user the last time each domain was used. Either way, we'll have to handle the case where the user chooses a subdomain they're already using on another gateway.
- Secondarily, we should give the user the ability to manage their domains from the gateway's settings (or some other) UI.

benfrancis:
Your "architecture" sounds like a UI design :) Let's get a UI design from UX, or if not I'll mock one up.

My suggestion is to offer to reclaim a domain if a user enters a subdomain which is already taken. Something along the lines of... "This name is already taken. If you previously registered this subdomain and want to transfer it to this gateway then click here to reclaim it".

mrstegeman:
How does this look (an extension of `flow.md` from the registration server):

1. [gateway <-> browser] Start the server on HTTP and load the setup UI at http://localhost:8080
2. [gateway <-> cloud] User chooses a DNS name already in use (determined via `/subscribe`) and is given the option to reclaim.
3. [gateway <-> cloud] Gateway calls `/reclaim`. A random reclamation token is generated and emailed to the registered email address.
4. [gateway <-> browser] User enters the reclamation token.
5. [gateway <-> cloud] Gateway again calls `/subscribe` with the reclamation token, receives and API token.
6. [gateway <-> cloud] Run the LE DNS challenge, use `/dnsconfig`, retrieve the certificates, restart the server on HTTPS.
7. [gateway <-> browser] Create the admin account with an email address, verify email using `/setemail`.
8. [gateway <-> cloud] Periodically ping the cloud service using `/ping`.

mrstegeman:
Server side is working and merged: https://github.com/mozilla-iot/registration_server/pull/45

PR #395, opened 2017-11-06
Call /ping periodically to track domain activity.

mrstegeman:
@andrenatal Can you update the config file on the server to evict old registrations every 12 or 24 hours? Not sure what it's presently set to.

### DIFF
--- a/src/ssltunnel.js
+++ b/src/ssltunnel.js
@@ -11,6 +11,7 @@
 const fs = require('fs');
 const config = require('config');
 const Settings = require('./models/settings');
+const fetch = require('node-fetch');
 var spawnSync = require('child_process').spawn;
 
 var TunnelService = {
@@ -21,7 +22,7 @@ var TunnelService = {
 
     // method that starts the client if the box has a registered tunnel
     start: function(response, urlredirect) {
-        Settings.get('tunneltoken').then(function(result) {
+        Settings.get('tunneltoken').then((result) => {
             if (typeof result === 'object') {
                 let endpoint = result.name + '.' +
                     config.get('ssltunnel.domain');
@@ -52,6 +53,10 @@ var TunnelService = {
                 this.pagekiteProcess.on('close', (code) => {
                     console.log(`[pagekite] process exited with code ${code}`);
                 });
+
+                // Ping the registration server every hour.
+                const delay = 60 * 60 * 1000;
+                setInterval(() => this.pingRegistrationServer(), delay);
             } else {
                 console.error('tunneltoken not set');
                 if (response) {
@@ -87,7 +92,7 @@ var TunnelService = {
         return typeof tunneltoken === 'object';
     },
 
-    // method to check if user skipped the sl tunnel setup
+    // method to check if user skipped the ssl tunnel setup
     userSkipped: async function() {
         let notunnel = await Settings.get('notunnel');
         if (typeof notunnel === 'boolean' && notunnel) {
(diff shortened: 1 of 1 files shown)


benfrancis:
@mrstegeman What problem is this solving?

Tip: It would be helpful if you could file issues in GitHub (bugs, tasks and user stories for features) for any issues you find and have pull requests reference the issue they're closing so that we can track and prioritise work that needs doing.

The setTimeout in this pull request raised a red flag for me, but I don't understand what problem it's trying to solve. Why is there a need to check the local IP address every hour?

Thanks

mrstegeman:
According to the registration server API document, we are supposed to call /register periodically in order to keep the route to the gateway up to date.
https://github.com/mozilla-iot/registration_server/blob/master/api.md

Currently, we're never calling this API. On the bright side, the registration server only evicts "old" entries on startup. However, my plan is to evict them on a more regular basis. As I start to implement the domain recovery flow, it will be useful to know which domains are actively being used, to avoid stomping on active domains.

As far as the every hour part, it's entirely possible that a home router will change the gateway's DHCP reservation at some point while it's running.

andrenatal:
@mrstegeman I created an account for you in the AWS panel, so please feel free to deploy any change that might be necessary.

benfrancis:
> According to the registration server API document, we are supposed to call /register periodically in order to keep the route to the gateway up to date.
> https://github.com/mozilla-iot/registration_server/blob/master/api.md

Ugh, reading that document reminds me how un-RESTful this "REST API" is.

`The REST api follows these general rules... All the requests are GET requests.`

I don't think I'm being particularly dogmatic when I say this misses the point about what a REST API is! If someone wanted to take over responsibility for this API and design a better one, that would be great!

Anyway...

`
The server will evict old entries on a regular basis so the gateway needs to register itself often enough.`

This is vague... What does it actually do? My gateway uses DHCP and I've never noticed it stop working because the gateway doesn't call this API method. Does anyone know how it actually uses the local IP address and why it needs sending to the cloud server from the gateway? I'm afraid my Rust is a bit rusty.

If this is necessary, is there a better way to schedule regular tasks in NodeJS than a setTimeout?

mrstegeman:
The registration server keeps a record of local and public IP address mappings for each gateway. Periodically, those entries are "evicted", meaning that all the gateway and domain data still exists, but the IP mappings (which are used for responding to DNS queries and such) are wiped.

Given that we were never calling the API and things still work fine, I'm not entirely sure why it exists. I'm guessing it works fine because things are just routed through PageKite tunnels. However, I was hoping to use this info for figuring out active/inactive domains for domain recovery purposes, and given that there was already an API that provided data I wanted, I figured it would be best to go ahead and use that.

If you'd prefer I pare down the APIs instead, and only use what we actually need, I can do that. Otherwise, I can look into an alternative to setTimeout().

benfrancis:
We inherited the registration server code from a previous project so I think it's worth getting a better understanding of how it works and moulding it more to our needs if necessary. I'll see if Fabrice can help explain the original design a bit more on IRC.

In the meantime, you said you wanted to use this to identify "active/inactive domains for domain recovery purposes". Can you explain this thinking a bit more in #379?

mrstegeman:
https://github.com/mozilla-iot/gateway/issues/379#issuecomment-342478155

mrstegeman:
This goes along with the following registration server PR:
https://github.com/mozilla-iot/registration_server/pull/42

mrstegeman:
The new /ping API is now live.

src/ssltunnel.js
```diff
@@ -87,15 +92,32 @@ var TunnelService = {
...
             return true;
         }
 
         return false;
-    }
+    },
+
+    // method to ping the registration server to track active domains
+    pingRegistrationServer: function() {
+        Settings.get('tunneltoken').then((result) => {
```
benfrancis: One minor point. Is it necessary to do this I/O every hour and wear out the SD card? :P Can't we store the token in memory and update it when it actually changes (which currently as far as I know it doesn't!).

mrstegeman: Done.

Issue #366, opened 2017-10-30
Page loading is slow


hobinjk:
As a user I want the gateway's UI to be interactive within at most 5 seconds the first time I load it, and within 2 seconds on subsequent loads. Ideally this performance would be achievable on mobile.

This is an issue covering various investigations into the loading speed of the gateway. The main goal is to reduce the time until the gateway is interactive (TTI).

[Baseline](https://www.webpagetest.org/result/171030_SF_48c1dfc0a4a9b9cbe49055a357d48f3b/): 16s first, 13s repeat
[Enabling max-age caching](https://www.webpagetest.org/result/171030_DS_a065d7b9c740090d4d09318b422e0c32/): 16s first, 6s repeat
[Enabling compression and caching](https://www.webpagetest.org/result/171030_K0_45c9e74e9c027535be48ed15c7a65e2a/): 16s first, 5.5s repeat

hobinjk:
Time to first byte is consistently at 2.5s when it could be closer to 0.4-0.8s

hobinjk:
The majority of the TTFB is from the 3G slow latency

hobinjk:
[Compress+cache desktop](https://www.webpagetest.org/result/171030_3B_6dd39200bcb5f5ba3bb8d1b0a011ba42/): 6.4s first, 2.7s repeat
[Baseline desktop](https://www.webpagetest.org/result/171030_V4_8c781afcd48ff1f896d80fc2f77481ae/): 6.4s first, 5s repeat

hobinjk:
[Concat desktop](https://www.webpagetest.org/result/171030_G8_2be1c9401af1a7299cb43d2ad00032ce/): 4.7 first, 4.6 repeat
[Concat+compress+cache desktop](https://www.webpagetest.org/result/171030_T6_f8bd0220eea1cff69548f52f886a3885/): 5.1 first, 2.4 repeat

hobinjk:
[Caching /things HTML output, not concating](https://www.web pagetest.org/result/171031_HT_5da53402d2127f0ba8ce56537003f2a6/): 6.7s first, 1.7s repeat

At this point I think the main focus should be SW caching and reducing the latency of the server. The dominating factor is the 0.8s TTFB.

hobinjk:
Something that would also help but require a ton of design thought is to perform early termination of the SSL connections at the api.mozilla-iot.org server and count on pagekite to secure the api to gateway connection

benfrancis:
[...]

PR #399, opened 2017-11-07
Baseline perf and PWA improvements

hobinjk:
Based on the work in #366 this incorporates all of the changes from that except for the ServiceWorker as that needs further development to be generated automatically

[Test results running this revision](https://www.webpagetest.org/result/171107_TE_19650bccd91d2eb596c59805597783b8/)


static/app.webmanifest
```diff
@@ -8,6 +8,11 @@
...
     "sizes": "256x256",
     "type": "image/png"
+  }, {
+    "src": "images/icon-2x.png",
+    "sizes": "512x512",
+    "type": "image/png"
   }],
-  "background_color": "#5d9bc7"
+  "background_color": "#5d9bc7",
+  "theme_color": "#75AACF"
```
benfrancis: Nit: I'd suggest the theme_color is darker than background_color rather than lighter, e.g. #5288af

PR #407, opened 2017-11-09
Actions API Support

hobinjk:
- [x] Add /things/:thingId/actions HTTP routes
- [x] Add WS API support

### DIFF
--- a/src/controllers/actions_controller.js
+++ b/src/controllers/actions_controller.js
@@ -59,7 +59,7 @@ ActionsController.get('/', function(request, response) {
   if (request.params.thingId) {
     response.status(200).json(Actions.getByThing(request.params.thingId));
   } else {
-    response.status(200).json(Actions.getUnassociated());
+    response.status(200).json(Actions.getGatewayActions());
   }
 });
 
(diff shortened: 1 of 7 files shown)


src/controllers/actions_controller.js
```diff
@@ -42,7 +56,11 @@ ActionsController.post('/', function (request, response) {
  * Handle getting a list of actions.
  */
 ActionsController.get('/', function(request, response) {
-  response.status(200).json(Actions.getAll());
+  if (request.params.thingId) {
+    response.status(200).json(Actions.getByThing(request.params.thingId));
+  } else {
+    response.status(200).json(Actions.getAll());
```
benfrancis: I propose one small change. I think we should start treating the gateway as a thing in its own right, so /actions is the action queue for the gateway thing, not a list of all actions requested by all things. Possibly the simplest way to implement this would be for /actions to return all actions with no thingId, rather than filtering by thingId. Does that make sense?

hobinjk: That makes sense, I'll modify it accordingly. Thanks!

hobinjk:
@benfrancis Updated accordingly

src/models/actions.js
```diff
@@ -74,16 +74,47 @@ class Actions extends EventEmitter {
   }
 
   /**
+   * Get only the actions which are not associated with a specific thing
```
benfrancis: Nit: Can we make this clearer that "not associated with a specific thing" actually means "associated with the gateway itself"?

PR #408, opened 2017-11-09
Add code to have the gateway start/restart plugins

dhylands:


src/adapters/plugin/plugin.js
```diff
@@ -178,6 +187,24 @@ class Plugin {
...
+
+    this.process.stdout.on('data', data => {
+      console.log(data.toString());
+    });
+    this.process.stderr.on('data', data => {
+      console.log(data.toString());
+    });
+    this.process.on('close', code => {
+      console.log('Plugin:', this.pluginId, 'died, code =', code, 'restarting...');
+      this.start();
```
dhylands: This should probably be rate limited - so that if there is an error in the exec line it doesn't just respawn and consume 100% of the CPU

src/adapters/plugin/plugin.js
```diff
@@ -178,6 +187,25 @@ class Plugin {
...
+    this.process.stdout.on('data', data => {
+      console.log(data.toString());
+    });
+    this.process.stderr.on('data', data => {
+      console.log(data.toString());
+    });
+    this.process.on('close', code => {
+      console.log('Plugin:', this.pluginId, 'died, code =', code,
+                  'restarting...');
+      this.start();
```
mrstegeman: I do think it would be good to add some sort of rate limiting here, like "if this process died X times in the past minute, don't restart". Something of that nature. Maybe just go ahead and file a separate issue for that so we don't forget?

dhylands:
I filed issue #413 to do actual rate-limiting of plugin launching. I added detection of the most common reason to fail (bad exec entry in config) in this PR.

hobinjk:
If possible, I'd like to see a test of `plugin.start`

dhylands:
I put @hobinjk and @mrstegeman down to re-review.

Compared to before, this adds 3 tests for starting - a good start, a bad start, and a restart.

I also wound up reworking some of @mrstegeman 's PR by splitting the test adapters into a separate directory.

src/adapter-loader.js
```diff
@@ -11,10 +11,74 @@
...
+  if (!newSettings.moziot.hasOwnProperty('config')) {
+    newSettings.moziot.config = {};
+  }
+
+  const errorCallback = function(packageName, errorStr) {
+    console.log('Failed to load', packageName, '-', errorStr);
+  };
+
+  var pluginClient = new PluginClient(adapterId, {verbose: verbose});
+  pluginClient.register().then(adapterManagerProxy => {
```
hobinjk: Once the code reaches this point it will always return true because it doesn't do anything with this promise (either returning or awaiting it)

src/test/integration/adapters-test.js
```diff
@@ -60,4 +60,8 @@ describe('adapters/', function() {
     }
   });
 
+  it('test loading an adapter via plugin mechanism', async () => {
+
```
hobinjk: Is this meant to be blank?

src/test/integration/plugins-test.js
```diff
@@ -0,0 +1,77 @@
...
+
+  it('Test the plugin start mechanism (good exec)', async () => {
+
+    // The use of the mock adapter already tests most of the methods in the
+    // Plugin class. however it doesn't test the start method.
+
+    let plugin = new Plugin('plugin-start-test', null);
+    plugin.exec = 'node -e process.exit(42);';
+    plugin.start();
+    // Normnally, the plugin mechanism would try to restart the plugin
```
hobinjk: nit: typo

src/test/integration/plugins-test.js
```diff
@@ -0,0 +1,77 @@
...
+    // when it exits, so we set restart to false to prevent that.
+    plugin.restart = false;
+    let promise = new Promise(resolve => {
+      plugin.process.on('exit', code => {
+        console.log('Got exit code', 42);
+        resolve(code);
+      });
+    });
+    let code = await(promise);
```
hobinjk: Could be simplified as `let code = await new Promise ...`

hobinjk:
Thanks for the tests! I left some comments inline.

src/test/integration/adapters-test.js
dhylands: I was originally going to put the tests here and decided to move it. I was supposed to checkout the previous version of this file, and forgot to.

config/test.js
```diff
@@ -15,6 +15,9 @@ module.exports = {
     http: 0,
   },
   behindForwarding: false,
+  adapterManager: {
+    path: './adapters-test',
+  },
```
mrstegeman: You should probably also define the API level constant as well.

src/adapter-loader.js
```diff
@@ -47,27 +112,8 @@ if (opt.argv.length != 1) {
   process.exit(1);
 }
 var adapterId = opt.argv[0];
```
mrstegeman: Can you change this to `packageName` to keep the terms consistent?

src/adapter-manager.js
```diff
@@ -397,15 +396,8 @@ class AdapterManager extends EventEmitter {
...
-      const defaults = [
-        'moziot-adapter-zigbee',
-        'moziot-adapter-zwave',
-      ];
-
-      if ((defaults.includes(manifest.name) && !testMode) ||
-          (manifest.moziot._test && testMode)) {
-        newSettings.moziot.enabled = true;
+      if (manifest.moziot.hasOwnProperty('enabled')) {
+        newSettings.moziot.enabled = manifest.moziot.enabled;
```
mrstegeman: This branch is unnecessary, since `newSettings` is just a copy of `manifest` at this point.

src/adapters-test/moziot-adapter-settings/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "moziot-adapter-settings",
+  "version": "0.2.1",
+  "description": "Settings test adapter plugin for Mozilla IoT Gateway",
+  "main": "index.js",
+  "keywords": [
+    "moziot",
+    "adapter",
+    "mock"
```
mrstegeman: This one is not `mock`, so maybe change that keyword.

src/adapters-test/moziot-adapter-settings/settings-adapter.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * mock-adapter.js - Mock Adapter.
```
mrstegeman: `settings`, not `mock`

src/adapters/plugin/plugin.js
```diff
@@ -178,8 +188,50 @@ class Plugin {
...
+  shutdown() {
+    this.ipcSocket.close();
+  }
+
+  start() {
+    // If we need embedded spaces, then consider changing to use the npm
+    // module called splitargs
+    this.restart = true;
+    let args = this.exec.split(' ');
+    this.process = spawn(args[0], args.slice(1));
```
mrstegeman: I don't love that we're allowing add-ons to execute arbitrary commands on start-up. Is there any reason to not force them to be loaded by `adapter-loader.js`?

src/test/integration/plugins-test.js
```diff
@@ -0,0 +1,76 @@
...
+/* globals it */
+
+const Plugin = require('../../adapters/plugin/plugin');
+
+describe('plugins/', function() {
+
+  it('Test the plugin start mechanism (good exec)', async () => {
+
+    // The use of the mock adapter already tests most of the methods in the
+    // Plugin class. however it doesn't test the start method.
```
mrstegeman: However,

src/adapters/plugin/plugin.js
dhylands: adapter-loader.js is really only for node plugins. As soon as a plugin written in another language comes along you'll have the same problem.

Besides, what's to stop a plugin loaded via adapter-loader.js from calling spawn? If you really wanted to lock this down you'd need to do something like use seccomp.

config/test.js
dhylands: So does this mean that you also think that the paths in config/default.js should also be constants? Since this is a config file, this is the source of the constant.

mrstegeman: No, I was just suggesting that you define it here, so that the API level checks are actually tested.

dhylands: Sorry - I misread your comment. The way that the config files work, the contents of test.js is overlaid with default.js. So test.js only provides things which need to be different from default.js.

See: https://github.com/lorenwest/node-config/wiki/Configuration-Files

In our case, default.js corresponds to default.EXT and test.js corresponds to {deployment}.EXT

mrstegeman: Ah, I see. In that case, carry on.

PR #412, opened 2017-11-10
Use tunnel token rather than shared secret for PK tunnel.

### LINKED ISSUE #138: Make PageKite use a unique token per gateway instead of a hardcoded secret

mrstegeman:
Fixes #138


mrstegeman:
It was actually already implemented and enabled on the server side, we just weren't using it.

PR #414, opened 2017-11-13
Revert "Use email on settings for LE and move email saving to signup page"

mrstegeman:
This email input is needed for domain reclamation purposes.

This reverts commit 1c9a52374b1a9553eed98f40eedad318d336ea87.


hobinjk:
Shouldn't this keep the `config.get('ssltunnel.certemail')` in places where it makes sense and only use the provided email in `/setemail`?

PR #415, opened 2017-11-14
Template the domain on the registration page.

mrstegeman:
Stop hardcoding mozilla-iot.org.


hobinjk:
LGTM, I think @benfrancis might have some opinions about using mustache to render the page.

mrstegeman:
I'm not tied to any particular templating engine. I just went with one that worked easily with express. I'll let @benfrancis weigh in.

Alternatively, we could do a simple search and replace. It just seemed a bit hacky to me. ¯\\\_(ツ)\_/¯

mrstegeman:
https://github.com/mozilla-iot/wiki/wiki/Template-Engine-Selection

PR #416, opened 2017-11-14
Add domain reclamation UI.

### LINKED ISSUE #379: Re-claim subdomain

mrstegeman:
Fixes #379

### DIFF
--- a/src/controllers/settings_controller.js
+++ b/src/controllers/settings_controller.js
@@ -72,8 +72,23 @@ SettingsController.get('/experiments/:experimentName',
   }
 });
 
+SettingsController.post('/reclaim', async (request, response) => {
+  let subdomain = request.body.subdomain;
+
+  try {
+    await fetch(config.get('ssltunnel.registration_endpoint') +
+                '/reclaim?name=' + subdomain);
+    response.status(200).end();
+  } catch (e) {
+    console.error(e);
+    response.statusMessage = 'Error issuing certificate - ' + e;
+    response.status(400).end();
+  }
+});
+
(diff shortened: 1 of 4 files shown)


mrstegeman:
Diff is cleaner if you add `?w=1` to the URL.

src/controllers/settings_controller.js
```diff
@@ -72,8 +72,23 @@ SettingsController.get('/experiments/:experimentName',
...
+SettingsController.post('/reclaim', async (request, response) => {
+  let subdomain = request.body.subdomain;
+
+  try {
+    await fetch(config.get('ssltunnel.registration_endpoint') +
+                '/reclaim?name=' + subdomain);
+    response.status(200).end();
+  } catch (e) {
+    console.error(e);
+    response.statusMessage = 'Error issuing certificate - ' + e;
```
hobinjk: This error message sounds out-of-place

hobinjk:
@mrstegeman Under what conditions can a domain be reclaimed? I tried testing by registering a domain, verifying an email, then resetting and trying to register the same domain. The Gateway just said the domain was in use without giving me the option to send a reclamation email

mrstegeman:
Fixed the error message.

Domains can be reclaimed if they have an email address associated with them, and the same email is provided when trying to reclaim. The required server changes are only active on the staging server right now, so if you'd like to try, just switch to mozilla-iot-staging.org.

mrstegeman:
Good call, @hobinjk. I removed the call to /setemail when reclaiming.

PR #418, opened 2017-11-16
Allow add-ons to be enabled/disabled from settings.

### LINKED ISSUE #78: Enable/disable adapter
As a user I want to see a list of adapters active on the gateway so I can choose which ones to enable.

### LINKED ISSUE #422: Enable/disable add-on
As a user I want to see a list of add-ons installed on the gateway so I can choose which ones to enable.

mrstegeman:
Fixes #422

### DIFF
--- a/src/adapter-manager.js
+++ b/src/adapter-manager.js
@@ -520,7 +520,7 @@ class AdapterManager extends EventEmitter {
    * example if the user presses the button on a device which is different
    * from the one that they requested removal of.
    */
-   removeThing(thingId) {
+  removeThing(thingId) {
     var deferredRemove = new Deferred();
 
     if (this.deferredAdd) {
--- a/src/adapter-manager.js
+++ b/src/adapter-manager.js
@@ -403,7 +403,7 @@ class AdapterManager extends EventEmitter {
         'moziot-adapter-zwave',
       ];
 
-      if (defaults.includes(manifest.name) ||
+      if ((defaults.includes(manifest.name) && !testMode) ||
           (manifest.moziot._test && testMode)) {
         newSettings.moziot.enabled = true;
       } else {
--- a/src/adapter-manager.js
+++ b/src/adapter-manager.js
@@ -336,6 +336,8 @@ class AdapterManager extends EventEmitter {
    * @method loadAdapter
    *
    * Loads a single adapter.
+   *
+   * @returns A promise which is resolved when the adapter is loaded.
    */
 
   async loadAdapter(adapterId) {
@@ -348,7 +350,7 @@ class AdapterManager extends EventEmitter {
     const packageJson = path.join(adapterPath, 'package.json');
     if (!fs.lstatSync(packageJson).isFile()) {
       console.error('package.json not found for adapter:', adapterId);
-      return;
+      return Promise.reject();
     }
 
     // Read the package.json file.
@@ -358,7 +360,7 @@ class AdapterManager extends EventEmitter {
     } catch (err) {
       console.error('Failed to read package.json for adapter:', adapterId);
       console.error(err);
-      return;
+      return Promise.reject();
     }
 
     let manifest;
@@ -367,7 +369,7 @@ class AdapterManager extends EventEmitter {
     } catch (err) {
       console.error('Failed to parse package.json for adapter:', adapterId);
       console.error(err);
-      return;
+      return Promise.reject();
     }
 
     // Verify API version.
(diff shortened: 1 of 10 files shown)


src/adapter-manager.js
```diff
@@ -430,24 +432,28 @@ class AdapterManager extends EventEmitter {
...
           console.log('Loading adapter', manifest.name, 'as plugin');
-          const adapterLoader = dynamicRequire(path.join(adapterPath));
+          const adapterLoader = dynamicRequire(adapterPath);
           adapterLoader(adapterManagerProxy, newSettings, errorCallback);
-        }).catch(err => {
+        } catch (err) {
           console.error('Failed to register adapter with gateway');
           console.error(err);
-        });
+          return Promise.reject();
```
dhylands: Using await here feels awkward to me. I think that you can do something like this instead:
```
let promise = plugin.register();
promise.then(adapterManagerProxy => {
   ....other stuff ...
}).catch(err => {
   ... handle error ...
});
return pomise;
```
But I haven't been doing JS very long, so I'd defer to @benfrancis or @hobinjk

mrstegeman: It was previously `.then(...).catch(...)`, but I feel like `await` is a lot cleaner, given that the enclosing function is already async.

hobinjk: The behavior is slightly different in this case because it previously swallowed the error. Returning `Promise.reject` reports the error in a way that `pluginClient.register().catch` didn't.

While using await isn't a significant saving in this scenario, I definitely think that it's worth converting this code to have consistency with other places in this file where async/await does save complexity.

src/adapter-manager.js
```diff
@@ -430,24 +432,28 @@ class AdapterManager extends EventEmitter {
...
         // started elsewhere.
       }
     } else {
       // Load this adapter directly into the gateway.
-      const adapterLoader = dynamicRequire(path.join(adapterPath));
+      const adapterLoader = dynamicRequire(adapterPath);
       adapterLoader(this, newSettings, errorCallback);
     }
+
+    return Promise.resolve();
```
hobinjk: You don't need this return statement, the async function returns a promise which resolves to whatever was returned from the function which is already `undefined`

src/adapter-manager.js
```diff
@@ -348,7 +350,7 @@ class AdapterManager extends EventEmitter {
     const packageJson = path.join(adapterPath, 'package.json');
     if (!fs.lstatSync(packageJson).isFile()) {
       console.error('package.json not found for adapter:', adapterId);
-      return;
+      return Promise.reject();
```
hobinjk: All uses of `Promise.reject` should provide an error message explaining where the failure occurred. Since this is an async function, you can also use `throw new Error('blah')` to get a native error object and all that nice business

hobinjk:
This looks good for the most part, nice work! I left some comments about Promises and async/await. I'd also like to see tests of the new Settings routes if possible.

src/adapter-manager.js
```diff
@@ -333,53 +333,58 @@ class AdapterManager extends EventEmitter {
...
-                                  adapterId);
+                                  packageName);
     const testMode = process.env.NODE_ENV === 'test';
 
     // Skip if there's no package.json file.
     const packageJson = path.join(adapterPath, 'package.json');
     if (!fs.lstatSync(packageJson).isFile()) {
-      console.error('package.json not found for adapter:', adapterId);
-      return Promise.reject();
+      const err = `package.json not found for package: ${packageName}`;
```
dhylands: While updating some other code, I noticed that this isn't using single quotes but some weird unicode character. And if you remove the package.json file and try to execute this path, it errors out.

src/adapter-manager.js
```diff
@@ -333,53 +333,58 @@ class AdapterManager extends EventEmitter {
...
     const apiVersion = config.get('adapterManager.api');
     if (manifest.moziot.api.min > apiVersion ||
         manifest.moziot.api.max < apiVersion) {
-      console.error('API mismatch for adapter:', manifest.name);
-      console.error('Current:', apiVersion, 'Supported:',
-                    manifest.moziot.api.min, '-', manifest.moziot.api.max);
-      return Promise.reject();
+      const err = `API mismatch for package: ${manifest.name}\nCurrent: ${
+        apiVersion} Supported: ${manifest.moziot.api.min}-${
+          manifest.moziot.api.max}`;
```
dhylands: And I think that this wasn't supposed to be a multi-line string (it also suffers from the single quote problem).

mrstegeman:
@dhylands Those are backticks, indicating a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Should be fully supported by node...

dhylands:
Doh - I think that the error was caused by me accidentally changing one of them to single quotes and then trying to run it,
I didn't know about template literals - you learn something new every day

PR #419, opened 2017-11-16
Initial OAuth2 Implementation

hobinjk:
It works! The current implementation hands out non-expiring access tokens instead of supporting refresh tokens. This is one of the two mechanisms supported by IFTTT, the other of which is mortal access tokens paired with non-expiring refresh tokens.

I tried to separate this out into reasonable commits but the implementation ended up being pretty all-or-nothing.

To-do:
- [x] Figure out what is causing jest to never terminate
- [x] Cooperate with content-negotiation middleware (have to re-read OAuth spec)
- [x] Support client_secret
- [x] Review the code for compliance with the RFC
- [x] Test rejection of password and client credentials
- [x] Test scope validation
- [x] Test redirect uri mismatch in /authorize and /token
- [x] Test providing an access_token JWT (instead of an authorization_code) to /token
- [x] Test providing a garbage JWT to /token

To-do in UI:
- [x] Authorization view: Prompt user to authenticate instead of automatically approving
- [x] Tie JWT to user who authenticates
- [x] List view: List all authorizations with the option to revoke their codes and tokens


benfrancis:
Great! What extra UI does this need? In our meeting I noted down:
- Authorize a third party app
- List authorized apps
- Revoke an app's authorization

Is that everything, and are there any more details or examples you can provide?

hobinjk:
@benfrancis Yes. There should be the authorize view and the list view with the list view containing a revoke button. There are some examples of authorize views [at oauth.com](https://www.oauth.com/oauth2-servers/scope/user-interface/).

Here's a picture of GitHub's [authorized applications](https://github.com/settings/applications) view:

screen shot 2017-11-17 at 10 04 42 am

hobinjk:
Note that the code at this stage doesn't have a real clients database or verification strategy, so a hypothetical adversary running on localhost:31338 could hijack the gateway. The review request at this stage is more for a sanity check before I implement the UI portion of this.

It might be better to look through the commits instead of the complete diff.

hobinjk:
Yes, typescript is very helpful here :)

tl;dr: A client needs to be registered in src/models/oauthclients.ts, which associates a potential client with an id, a secret, a scope, and a redirect uri. If any of these parameters isn't specified at the proper time or is mismatched, the OAuth flow will abort.

The oauth flow goes something like this (eliding a few optional parameters):

- User visits client.com
- client.com redirects the user to GET gateway.local/authorize?client_id=client&scope=readwrite
- The gateway verifies that the client "client" exists and is allowed to request the scope "readwrite"
- The client's existence verified, the gateway then looks up the redirect uri associated with the client which we'll say is client.com/callback
- The user is then presented with the option to trust client.com or not
- Upon trusting client.com, the user is redirected to client.com/callback?code=secretCodeJwt
- secretCodeJwt is a JWT in the database with the role 'authorization_code'. In the future it will be set to expire after 10 minutes.
- The client can then behind-the-scenes issue a POST to gateway.local/token that looks like
```
Authorization: Basic scaryBase64String
Content-Type: application/json
// other headers go here
{
  "grant_type": "authorization_code",
  "code": secretCodeJwt,
  "client_id": "client"
}
```
- The gateway then verifies that "client" exists, secretCodeJwt is a valid authorization_code JWT associated with "client", and scaryBase64String is a proper basic authentication header of the form "client_id:client_secret"
- Once verified, the gateway replies with an access_token JWT with the permissions granted by initial /authorize request.

Writing all of this out has been very helpful to me as a way of collecting my thoughts. If anyone who ends up reviewing this thinks something is wrong with this explanation I would love to hear about it as it's very likely present in the code too (unless it's a parameter I elided).

hobinjk:
@benfrancis Do you have an SVG file for the shield icon used in the mockups?

benfrancis:
> @benfrancis Do you have an SVG file for the shield icon used in the mockups?

@hobinjk Yep, uploaded here https://drive.google.com/a/mozilla.com/file/d/1IvR8YMlxBCUFG67TwtpOJCPibyT_1ZYD/view?usp=sharing

dhylands:
If I do a GET to https://localhost:4443/oauth/authorize?client_id=hello (from Postman) then I get the following error:
```
Unhandled Rejection
TypeError: Cannot read property 'toString' of undefined
    at redirect (/home/dhylands/Dropbox/moziot/gateway/build/webpack:/src/controllers/oauth_controller.ts:103:42)
    at verifyAuthorizationRequest (/home/dhylands/Dropbox/moziot/gateway/build/webpack:/src/controllers/oauth_controller.ts:196:5)
    at OAuthController.get (/home/dhylands/Dropbox/moziot/gateway/build/webpack:/src/controllers/oauth_controller.ts:231:16)
    at Layer.handle [as handle_request] (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/layer.js:95:5)
    at /home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:335:12)
    at next (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:275:10)
    at Function.handle (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:174:3)
    at router (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:47:12)
    at Layer.handle [as handle_request] (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:317:13)
    at /home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/home/dhylands/Dropbox/moziot/gateway/node_modules/express/lib/router/index.js:335:12)
```

hobinjk:
I've uploaded the code I use to test the hello world client here: https://gist.github.com/hobinjk/763c653ba27e6cc69ec5dbbfaa137fc7

To test, visit https://localhost:4443/oauth/authorize?response_type=code&client_id=hello&scope=readwrite&state=asdf in your browser. The response_type, client_id, and scope parameters are all required and should give errors if they are anything other than the values in that query string.

read/write/thing scoping is being left to future work

dhylands:
So I ran the gateway (on my local machine), I then copied the oauth-test.js file from the gist above into the gateway directory and ran `node oauth-test.js` in a separate window, browsed to the URL mentioned above: https://localhost:4443/oauth/authorize?response_type=code&client_id=hello&scope=readwrite&state=asdf I was then presented to the authorization page, and I clicked Allow. The browser then showed 'Unauthorized' with this URL in the browser: https://localhost:4443/oauth/allow?response_type=code&client_id=hello&scope=readwrite&state=asdf&jwt=eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IjdkMGRkOGU0LWRkZjctNGQ3Zi1iYzA5LTgzNzc4NjY0NDVlNCJ9.eyJyb2xlIjoidXNlcl90b2tlbiIsImlhdCI6MTUxMTk5MTY3MH0.DJSBjlOK2WJbRzdNVJy1s6jDVsoMTQruWiX9Jes1eggjMo9Siyv_XMqXaruwvyOG6ND65F36-zX9zPAL-P-OhA

hobinjk:
You may need to log out and log back in for the latest changes to propagate to your user JWT. I added the user_token role to differentiate them from OAuth-issued tokens

dhylands:
I removed the database and ran it again, with the same results. Also tried logging out and back in, and also the same results. Right now I'm running on my Mac (I was previously running on my linux box), and I noticed that the URL doesn't have a jwt. It's currently showing https://127.0.0.1:4443/oauth/allow?response_type=code&client_id=hello&scope=readwrite&state=somethingrandom&jwt=
I'll try again on my linux box when I get home.

hobinjk:
@dhylands Interesting. Can you open up the console in the /oauth/authorize page and verify that there are no errors and `window.API.jwt` exists?

dhylands:
Under linux, the authorize page has a jwt that looks valid. Under OSX, it comes back `null`. If I browse to https://localhost:4443/ under OSX, then window.API.jwt looks valid.

hobinjk:
@dhylands That sounds very strange. I've added some checks that should theoretically guarantee the presence of a valid JWT

src/db.js
```diff
@@ -416,43 +417,67 @@ var Database = {
...
-      'INSERT INTO jsonwebtoken_to_user (keyId, user, issuedAt, publicKey) ' +
-      'VALUES (?, ?, ?, ?)',
-      [keyId, user, issuedAt, publicKey]
+      'INSERT INTO jsonwebtokens (keyId, user, issuedAt, publicKey, payload) ' +
+      'VALUES (?, ?, ?, ?, ?)',
+      [keyId, user, issuedAt, publicKey, JSON.stringify(payload)]
     );
     assert(typeof result.lastID === 'number');
     return result.lastID;
   },
```
dhylands: Presumably we need a mechanism to actually expire the old temporary tokens. I noticed that I have a bunch in my database now. Perhaps an issue should be created?

src/db.js
```diff
@@ -416,43 +417,67 @@ var Database = {
...
       keyId
     );
   },
 
+  /**
+   * Get all known JWTs of a user
+   * @param {number} userId
+   * @return {Promise<Array<Object>>}
+   */
+  getJSONWebTokensByUser: function(userId) {
```
dhylands: I didn't see this function actually being used anywhere. Presumably it will be needed later?

static/js/settings.js
```diff
@@ -382,5 +388,68 @@ var SettingsScreen = {
...
+
+    fetch('/authorizations/' + clientId, {
+      headers: API.headers(),
+      method: 'DELETE'
+    }).then(() => {
+      let authorization = revoke.parentNode;
+      authorization.parentNode.removeChild(authorization);
+    }).catch(err => {
+      console.warn('Unable to revoke', err);
+    });
```
dhylands: I noticed that when I revoke the last authorization, the screen just comes back blank rather than saying "No authorizations"

src/db.js
hobinjk: Yes, I left out a "DROP TABLE IF EXISTS jsonwebtoken_to_user" because it would make flipping between this and master difficult, but it should be added in the future. I filed #448 to keep track

src/db.js
hobinjk: It's used in https://github.com/mozilla-iot/gateway/pull/419/files#diff-f1ef7a0af26e5388f461170438e6b4adR20 and https://github.com/mozilla-iot/gateway/pull/419/files#diff-f1ef7a0af26e5388f461170438e6b4adR40

static/js/settings.js
hobinjk: Good catch! I'll fix this behavior

PR #424, opened 2017-11-20
Change adapter to add-on where appropriate.

mrstegeman:
This should help clear up a lot of the terminology confusion we've been dealing with lately.


src/addon-manager.js
```diff
@@ -445,38 +444,37 @@ class AdapterManager extends EventEmitter {
...
+      const addonLoader = dynamicRequire(addonPath);
+      addonLoader(this, newSettings, errorCallback);
     }
   }
 
   /**
-   * @method loadAdapters
-   * Loads all of the configured adapters from the adapters directory.
+   * @method loadAddons
+   * Loads all of the configured add-ons from the adapters directory.
```
dhylands: nit: Probably need to change the word 'adapters' here.

src/addons/moziot-adapter-gpio/gpio-adapter.js
```diff
@@ -105,18 +105,18 @@ class GpioDevice extends Device {
 }
 
 class GpioAdapter extends Adapter {
```
dhylands: There are some references to adapterManager in addons-test/moziot-adapter-settings/settings-adapter.js that should be changed to addonManager.

src/addon-manager.js
```diff
@@ -386,8 +385,8 @@ class AdapterManager extends EventEmitter {
       return Promise.reject(err);
     }
 
-    // Get any saved settings for this adapter.
-    const key = `adapters.${manifest.name}`;
+    // Get any saved settings for this add-on.
+    const key = `addons.${manifest.name}`;
```
dhylands: Does this need some type of settings migration for people who have the previous version installed?

Maybe not an issue since we haven't done a release since adapters.xxx was introduced.

mrstegeman: Yeah, that was my thought. Since we never released, just roll with it.

PR #425, opened 2017-11-20
Add-on cleanups.

mrstegeman:
* GPIO and example adapters are now in their own repos.
* Stop using moziot terminology.
* Only show add-ons that are present in the file system on the UI.


src/controllers/settings_controller.js
```diff
@@ -245,7 +245,19 @@ SettingsController.get('/addons', async (request, response) => {
     if (result === undefined) {
       response.status(404).json([]);
     } else {
-      response.status(200).json(result);
+      let presentAddons = [];
+      for (const setting of result) {
+        const packageName = setting.key.substr(setting.key.indexOf('.') + 1);
```
dhylands: I think it would be worthwhile adding a comment that you're stripping the leading 'addons.' from the beginning of the key.

PR #429, opened 2017-11-21
Add UI for add-on discovery.

### LINKED ISSUE #195: Discover add-ons
As a user I want to find out what adapters are available for the gateway

This may end up being as simple as a tag on npmjs.com or a web page with a directory of adapters.

mrstegeman:
Fixes #195

### DIFF
--- a/src/controllers/addons_controller.js
+++ b/src/controllers/addons_controller.js
@@ -1,0 +1,25 @@
+const PromiseRouter = require('express-promise-router');
+const fetch = require('node-fetch');
+const AddonManager = require('../addon-manager');
+
+const AddonsController = PromiseRouter();
+
+AddonsController.get('/discover', async (request, response) => {
+  let addons;
+  try {
+    const res = await fetch('https://raw.githubusercontent.com/mozilla-iot/' +
+                            'addon-list/master/list.json');
+    addons = await res.json();
+  } catch (e) {
+    response.status(400).send(e);
+    return;
+  }
+
+  let discovered = addons.map((a) => {
+    a.installed = AddonManager.isAddonAvailable(a.name);
+    return a;
+  });
+  response.status(200).json(discovered);
+});
+
+module.exports = AddonsController;
--- a/src/controllers/addons_controller.js
+++ b/src/controllers/addons_controller.js
@@ -1,10 +1,86 @@
 const PromiseRouter = require('express-promise-router');
 const fetch = require('node-fetch');
 const AddonManager = require('../addon-manager');
+const Settings = require('../models/settings');
 
 const AddonsController = PromiseRouter();
 
-AddonsController.get('/discover', async (request, response) => {
+AddonsController.get('/available', async (request, response) => {
+  Settings.getAddonSettings().then(function(result) {
+    if (result === undefined) {
+      response.status(404).json([]);
+    } else {
+      let availableAddons = [];
+      for (const setting of result) {
+        // Remove the leading 'addons.' from the settings key to get the
+        // package name.
+        const packageName = setting.key.substr(setting.key.indexOf('.') + 1);
+        if (packageName.length <= 0) {
+          continue;
+        }
+
+        if (AddonManager.isAddonAvailable(packageName)) {
+          availableAddons.push(setting);
+        }
+      }
+
+      response.status(200).json(availableAddons);
+    }
+  }).catch(function(e) {
+    console.error('Failed to get add-on settings.');
+    console.error(e);
+    response.status(400).send(e);
+  });
+});
+
+AddonsController.put('/toggle/:addonName', async (request, response) => {
+  var addonName = request.params.addonName;
+
+  if (!request.body || request.body['enabled'] === undefined) {
+    response.status(400).send('Enabled property not defined');
+    return;
+  }
+
+  var enabled = request.body['enabled'];
+
+  const key = `addons.${addonName}`;
+
+  let current;
+  try {
+    current = await Settings.get(key);
+  } catch (e) {
+    console.error('Failed to get current settings for add-on ' + addonName);
+    console.error(e);
+    response.status(400).send(e);
+    return;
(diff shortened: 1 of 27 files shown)


mrstegeman:
@hobinjk I've got a bugfix PR coming to fix that issue and some other CSS stuff.

static/js/settings.js
```diff
@@ -296,6 +326,89 @@ var SettingsScreen = {
...
+        const titleDiv = document.createElement('div');
+        titleDiv.innerText = addon.display_name;
+        titleDiv.className = 'addon-settings-name';
+
+        const descriptionDiv = document.createElement('div');
+        descriptionDiv.innerText = addon.description;
+        descriptionDiv.className = 'addon-settings-description';
+
+        const controlDiv = document.createElement('div');
+        controlDiv.className = 'addon-settings-controls';
```
benfrancis: Nit: Can we come up with more semantic markup that just divs? At least span for the text labels.

static/js/settings.js
```diff
@@ -296,6 +326,89 @@ var SettingsScreen = {
...
+          actionElement.className = 'addon-discovery-settings-added';
+          actionElement.innerText = 'Added';
+        } else {
+          actionElement = document.createElement('button');
+          actionElement.id = `addon-install-${addon.name}`;
+          actionElement.className = 'text-button addon-discovery-settings-add';
+          actionElement.innerText = 'Add';
+        }
+
+        /* TODO
```
benfrancis: I'd suggesting leaving out this commented code for now, maybe even remove the install button given it doesn't do anything!

static/js/settings.js
```diff
@@ -296,6 +326,89 @@ var SettingsScreen = {
...
+    }).then(function (body) {
+      if (!body) {
+        return;
+      }
+
+      const addonList = document.getElementById('discovered-addons-list');
+      addonList.innerHTML = '';
+
+      for (const addon of body) {
+        const li = document.createElement('li');
```
benfrancis: There's nothing fundamentally wrong with this approach, but just note that elsewhere I've abstracted components like this out into a separate object with a view() method to generate an HTML string and a render() method which adds it to the DOM (e.g. thing.js and new-thing.js) - a little bit like React. You then create each item by instantiating an object. It's not particularly sophisticated but it might scale better than putting everything in one giant for loop.

I see that when you moved adapters into settings you removed the Adapter object in adapter.js so maybe we should have a discussion about what approach we want to use for things like this across the front end?

We could even replace the current approach with something more sophisticated like Web Components.

benfrancis:
Just playing with the UI, I noticed a weird glitch when entering the add-ons screen in settings. I see a white bar at the bottom of the screen and then after the CSS transition the screen drops vertically by the same number of pixels and falls into place. Not sure what's causing it, but maybe something to do with the overflow:auto change to the CSS?

src/controllers/addons_controller.js
```diff
@@ -0,0 +1,25 @@
+const PromiseRouter = require('express-promise-router');
+const fetch = require('node-fetch');
+const AddonManager = require('../addon-manager');
+
+const AddonsController = PromiseRouter();
+
+AddonsController.get('/discover', async (request, response) => {
```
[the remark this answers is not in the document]
mrstegeman: Sure, I can change the path.

If we do it on the back-end, we can compare the list against what's already installed.

src/controllers/addons_controller.js
```diff
@@ -0,0 +1,25 @@
...
+  let addons;
+  try {
+    const res = await fetch('https://raw.githubusercontent.com/mozilla-iot/' +
+                            'addon-list/master/list.json');
```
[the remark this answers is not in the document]
mrstegeman: Agreed, will do.

static/images/add-on.svg
```diff
@@ -0,0 +1,69 @@
...
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
```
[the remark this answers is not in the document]
benfrancis: Updated icon https://drive.google.com/a/mozilla.com/file/d/1iJgqfzQG4jkuAIDz0SZvxlgyYTDoFpHS/view?usp=sharing

static/js/settings.js
mrstegeman: I'll push up a PR to enable this immediately after this is merged.

static/js/settings.js
mrstegeman: Yep, I'll switch them to spans.

static/js/settings.js
mrstegeman: I'll split it back out into a view object.

mrstegeman:
@benfrancis fixed the transition as well.

mrstegeman:
@benfrancis What would your path suggestion be for installing and uninstalling? I think right now I have /addons/install and /addons/uninstall, both PUT requests. I just want to get that figured out so we can be consistent without me having to change them again.

benfrancis:
@mrstegeman Hmm, I guess installing an add-on on the gateway is effectively adding a new add-on resource. So personally I would go for `POST /addons` with the data needed to "add" the new add-on and `DELETE /addons/:addonName` to remove it.

mrstegeman:
@benfrancis think I've got all the path name issues resolved.

mrstegeman:
@benfrancis done

PR #430, opened 2017-11-21
Fixes #397 - allow scanChannels to be overridden via config

### LINKED ISSUE #397: ZigBee channel issues with GE Modules
The GE modules only seem to pair with the dongle when the dongle is using channel 15 or 20 (which are the 2 channels "between" the 3 Wifi frequencies. See: https://support.metageek.com/hc/en-us/article_attachments/115017048148/ZigBee_Channels.pdf

I think that we should have the dongle restrict itself to those 2 channels, since those will in general reduce the amount of interference.

This would correspond to an SC setting of 0x0210. I think that this should be configurable, so that if we need to we can go back to the old behaviour of hainvg SC set to 0xFFFF

dhylands:
This allows the scanChannels to be overridden to force a
particular ZigBee operating channel to be used.

### DIFF
--- a/src/addons/zigbee-adapter/package.json
+++ b/src/addons/zigbee-adapter/package.json
@@ -25,6 +25,9 @@
     },
     "enabled": true,
     "plugin": true,
-    "exec": "node ./src/addon-loader.js zigbee-adapter"
+    "exec": "node ./src/addon-loader.js zigbee-adapter",
+    "config": {
+      "scanChannels": "0x1ffe"
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


benfrancis:
@dhylands Just to confirm, does this leave the default as it already was?

I'd still be interested to know what the spec has to say about choosing channels for compliance testing. I know ZigBee does some clever jumping around to avoid interference from WiFi channels in the same frequency band, but I have no idea how it works or what makes a device "compliant" with the standard.

benfrancis:
Oh, looks like you already answered my question :) https://github.com/mozilla-iot/gateway/issues/397#issuecomment-346175150

Issue #81, opened 2017-05-23
Settings - Edit user


benfrancis:
As a user I want to edit the name, email address or password of a user so that their details are up to date

Issue #80, opened 2017-05-23
Settings - Remove user


benfrancis:
As a user I want to remove a user from the gateway so that they no longer have access

PR #436, opened 2017-11-29
Add APIs to edit and delete users.

mrstegeman:
References #80
References #81

### DIFF
--- a/src/controllers/users_controller.js
+++ b/src/controllers/users_controller.js
@@ -76,7 +76,7 @@ UsersController.put('/:userId', async (request, response) => {
   const user = await Users.getUserById(request.params.userId);
 
   if (!user) {
-    response.sendStatus(500);
+    response.sendStatus(404);
     return;
   }
 
--- a/src/controllers/users_controller.js
+++ b/src/controllers/users_controller.js
@@ -46,16 +46,16 @@ UsersController.get('/info', auth, async (request, response) => {
  * Create a user
  */
 UsersController.post('/', async (request, response) => {
-  let body = request.body;
+  const body = request.body;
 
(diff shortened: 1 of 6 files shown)


src/controllers/users_controller.js
```diff
@@ -69,4 +69,48 @@ UsersController.post('/', async (request, response) => {
...
 });
 
+/**
+ * Edit a user
+ */
+UsersController.put('/:userId', async (request, response) => {
+  const user = await Users.getUserById(request.params.userId);
+
+  if (!user) {
+    response.sendStatus(500);
```
hobinjk: To me this feels more like a 404 user not found than an internal server error.

mrstegeman: Fixed.

src/test/integration/users-test.js
```diff
@@ -50,3 +55,50 @@ it('logs in as a user', async () => {
...
+it('edits an invalid user', async () => {
+  const jwt = await createUser(server, TEST_USER);
+  const rsp = await pFinal(
+    editUser(server, jwt, Object.assign(TEST_USER_UPDATE_1, {id: 0})));
+  expect(rsp.response.status).toEqual(404);
+});
+
+it('edits a user', async () => {
+  const jwt = await createUser(server, TEST_USER);
+  const user = await Users.getUser(TEST_USER.email);
```
hobinjk: I'd prefer for this to use the existing `userInfo(server, jwt)` function to get the user description rather than retrieving it directly from the model.

mrstegeman: Forgot I added the ID to the info returned by the API. Done.

PR #437, opened 2017-11-29
Add support for dimmable zigbee lights

dhylands:
- added support for dimmable zigbee lights (i.e. ZigBee devices which support the genLevelCtrl cluster)
- modified the adapter loader to not relaunch an addon if it dies due to things like syntax error.
- the plugin class now prefixes any logged lines using the adapter-id (with the string '-adapter' removed).
- added a config parameter to the zigbee adapter which will cause a discover message to be sent to each cluster and the received attributes are then logged. This was used to generate https://github.com/mozilla-iot/wiki/wiki/ZigBee-Attributes
- modified the 3rd argument of the Property class from being just a type to being a property description, so the external adapter repositories will need to be modified and merged around the same time as this change is merged.

- the propertyChanged IPC message was also modified (and the wiki needs to be updated). The original message looks like:
```
{
  messageType: 'propertyChanged',
  data: {
    pluginId: 'pluginId-string',
    adapterId: 'adapterId-string',
    deviceId: 'device-id',
    propertyName: 'name-of-property',
    propertyValue: updated-value,
  },
}
```
The new version looks like:
```
{
  messageType: 'propertyChanged',
  data: {
    pluginId: 'pluginId-string',
    adapterId: 'adapterId-string',
    deviceId: 'device-id',
    property: {
      name: 'property-name',
      value: 'property-value',
      ... remaining fields from property.asDict() ...
    },
  },
}
```
The change to the properyChanged notification allows the property description information to filter up to the PropertyProxy and also allows protocol specific fields to be exposed (this is especially useful when debugging).
This change will affect @hobinjk and his rust implementation of the hue adapter.

### DIFF
--- a/src/addons/plugin/device-proxy.js
+++ b/src/addons/plugin/device-proxy.js
@@ -30,7 +30,8 @@ class DeviceProxy extends Device {
     // Copy over any extra device fields which might be useful for debugging.
     this.deviceDict = {};
     for (let field in deviceDict) {
-      if (field in ['id', 'name', 'type', 'properties', 'actions', 'events']) {
+      if (['id', 'name', 'type', 'properties', 'actions', 'events']
+            .includes(field)) {
         continue;
       }
       this.deviceDict[field] = deviceDict[field];
--- a/src/addons/plugin/device-proxy.js
+++ b/src/addons/plugin/device-proxy.js
@@ -27,7 +27,20 @@ class DeviceProxy extends Device {
       this.properties.set(propertyName, propertyProxy);
     }
 
+    // Copy over any extra device fields which might be useful for debugging.
+    this.deviceDict = {};
+    for (let field in deviceDict) {
+      if (field in ['id', 'name', 'type', 'properties', 'actions', 'events']) {
+        continue;
+      }
+      this.deviceDict[field] = deviceDict[field];
+    }
+
     //TODO: Add support for actions once we know what they look like.
+  }
+
+  asDict() {
+    return Object.assign({}, this.deviceDict, super.asDict());
   }
 }
 
--- a/src/addons/plugin/property-proxy.js
+++ b/src/addons/plugin/property-proxy.js
@@ -15,18 +15,16 @@ const Property = require('../property');
 
 class PropertyProxy extends Property {
   constructor(device, propertyName, propertyDict) {
-    super(device, propertyName, propertyDict.type);
-
-    if (propertyDict.description) {
-      this.description = propertyDict.description;
-    }
-    if (propertyDict.unit) {
-      this.description = propertyDict.unit;
-    }
+    super(device, propertyName, propertyDict);
 
     this.value = propertyDict.value;
 
     this.propertyChangedPromises = [];
+    this.propertyDict = {};
+  }
+
+  asDict() {
+    return Object.assign({}, this.propertyDict, super.asDict());
   }
 
   /**
@@ -45,11 +43,12 @@ class PropertyProxy extends Property {
    * Called whenever a property changed notification is received
    * from the adapter.
    */
-  doPropertyChanged(value) {
-    this.setCachedValue(value);
+  doPropertyChanged(propertyDict) {
+    this.propertyDict = Object.assign({}, propertyDict);
+    this.setCachedValue(propertyDict.value);
     while (this.propertyChangedPromises.length > 0) {
       var deferredChange = this.propertyChangedPromises.pop();
-      deferredChange.resolve(value);
+      deferredChange.resolve(propertyDict.value);
     }
   }
 
--- a/src/addons/zigbee-adapter/zb-adapter.js
+++ b/src/addons/zigbee-adapter/zb-adapter.js
@@ -1,6 +1,6 @@
 /**
  *
- * ZigBeeAdapter - Adapter which manages ZigBee devices.
+ * ZigbeeAdapter - Adapter which manages Zigbee devices.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -10,7 +10,7 @@
 'use strict';
 
 var Adapter = require('../adapter');
-var ZigBeeNode = require('./zb-node');
+var ZigbeeNode = require('./zb-node');
 var SerialPort = require('serialport');
 var xbeeApi = require('xbee-api');
 var at = require('./zb-at');
@@ -72,9 +72,9 @@ function FUNC(ths, func, args) {
   return new Command(EXEC_FUNC, [ths, func, args]);
 }
 
-class ZigBeeAdapter extends Adapter {
+class ZigbeeAdapter extends Adapter {
   constructor(addonManager, manifest, port) {
-    // The ZigBee adapter supports multiple dongles and
+    // The Zigbee adapter supports multiple dongles and
     // will create an adapter object for each dongle.
     // We don't know the actual adapter id until we
     // retrieve the serial number from the dongle. So we
@@ -149,7 +149,7 @@ class ZigBeeAdapter extends Adapter {
         console.error('SerialPort open err =', err);
       }
 
-      // Hook up the ZigBee raw parser.
+      // Hook up the Zigbee raw parser.
       this.serialport.on('data', chunk => {
         this.xb.parseRaw(chunk);
       });
@@ -189,7 +189,7 @@ class ZigBeeAdapter extends Adapter {
     // Use this opertunity to create the node for the Coordinator.
 
     var coordinator = this.nodes[this.serialNumber] =
-      new ZigBeeNode(this, this.serialNumber, this.networkAddr16);
+      new ZigbeeNode(this, this.serialNumber, this.networkAddr16);
 
     // Go find out what devices are on the network.
     this.queueCommands(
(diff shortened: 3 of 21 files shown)


benfrancis:
Cool. It looks like this exposes a ZigBee dimmable light as a multiLevelSwitch.

How does the classification work? Does it currently just classify anything which supports the Level Control cluster as a multiLevelSwitch?

I note that the ZigBee Home Automation profile (v1.2) defines device descriptions (under section 5.7 Device Descriptions) for...

Generic:
* Level Control Switch (Identify, On/Off & Level Control clusters)
* Level Controllable Output (On/Off, Level Control, Scenes, Groups)

Lighting:
* Dimmable Light (On/Off, Level Control, Scenes, Groups)
* Color Dimmable Light (On/Off, Level Control, Color Control, Scenes, Groups)
* Dimmer Switch (On/Off, Level Control, Identify)

(I still don't understand the difference between "server side" and "client side" clusters in the spec...)

Presumably all of these would currently be mapped onto a multiLevelSwitch device type? I think that's right, I'm just wondering whether in future we also need a dimmableLight type to go with the colorDimmableLight type? I'm honestly not sure what the difference is between the two, many of these types seem the same to me!

I think you've said before that actual devices don't expose these "device descriptions" as outlined in the specification, just the clusters they support. Is that correct? Is there any technical reason to support separate multiLevelSwitch and dimmableLight web thing types, or just that they could be represented by a different icon in the UI?

DavidKarlas:
About `(I still don't understand the difference between "server side" and "client side" clusters in the spec...)` I assume you are talking about 2.2.2 from http://www.zigbee.org/wp-content/uploads/2014/10/07-5123-06-zigbee-cluster-library-specification.pdf
Zigbee is really meant(specified) as mesh network but mostly used as centralised around gateway/hub...

The way things work in zigbee each device has list of `inputClusters` and `outputClusters`. For example LightBulb will have `CLUSTER_ID_GENONOFF_HEX` in list of `inputClusters` but SmartButton will have same cluster `CLUSTER_ID_GENONOFF_HEX` in `outputClusters`...

What controller can do... It can Bind `outputCluster(CLUSTER_ID_GENONOFF_HEX)` from SmartButton to `inputCluster(CLUSTER_ID_GENONOFF_HEX)` of LightBulb. And Zigbee network will be 100% independent from Hub/Gateway after this binding is done even if raspberry dies it will continue to work, this Bindings tables is stored on XBee controller(or any other) itself in persistent storage so even after power loss when button is pressed on SmartButton it will try to find LightBulb on network and send it command...

But you probably don't want to do it this way instead you will want to send Bind to SmartButton `outputCluster` and tell it to send message to controller(XBee aka. Gateway) when button is pressed and then Gateway Rules will call LightBulb `inputClusters` to change that value...

Fun thought... It's possible to create high availability mesh network with 2 XBees and 2 RaspberryPis, but you would need to Bind SmartButtons to send to both XBees and then 2 Gateways would need logic to determine who is master and who is in HotStandby

src/addons/plugin/device-proxy.js
```diff
@@ -27,8 +27,21 @@ class DeviceProxy extends Device {
       this.properties.set(propertyName, propertyProxy);
     }
 
+    // Copy over any extra device fields which might be useful for debugging.
+    this.deviceDict = {};
+    for (let field in deviceDict) {
+      if (field in ['id', 'name', 'type', 'properties', 'actions', 'events']) {
```
mrstegeman: I believe this should be:
`if (['id', 'name', 'type', 'properties', 'actions', 'events'].includes(field))`

src/addons/plugin/device-proxy.js
```diff
@@ -27,8 +27,21 @@ class DeviceProxy extends Device {
...
+        continue;
+      }
+      this.deviceDict[field] = deviceDict[field];
+    }
+
     //TODO: Add support for actions once we know what they look like.
   }
+
+  asDict() {
+    return Object.assign({}, this.deviceDict, super.asDict());
```
mrstegeman: This is going to overlay the base class properties on top of this object's properties. Was that intended?

src/addons/plugin/property-proxy.js
```diff
@@ -15,18 +15,16 @@ const Property = require('../property');
...
+    super(device, propertyName, propertyDict);
 
     this.value = propertyDict.value;
 
     this.propertyChangedPromises = [];
+    this.propertyDict = {};
+  }
+
+  asDict() {
+    return Object.assign({}, this.propertyDict, super.asDict());
```
mrstegeman: This is going to overlay the base class properties on top of this object's properties. Was that intended?

src/addons/zigbee-adapter/zb-classifier.js
```diff
@@ -10,22 +10,73 @@
...
+      CLUSTER_ID_GENONOFF,            // clusterId
+      'onOff',                        // attr
+      'setOnOffValue',                // setAttrFromValue
+      'parseOnOffAttr'                // parseValueFromAttr
+    );
+  }
+
+  addProperty(node, name, descr, profileId, endpoint, clusterId,
+              attr, setAttrFromValue, parseValueFromAttr) {
+    let property =  new ZigBeeProperty(node, name, descr, profileId,
```
mrstegeman: nit: Could we change all instances of ZigBee to Zigbee? Zigbee is the proper name.

src/addons/zigbee-adapter/zb-property.js
```diff
@@ -30,12 +58,114 @@ class ZigBeeProperty extends Property {
...
+   *    attrData contains the actual data.
+   *
+   * The above fields can be examined symbolically using the zcl-id module:
+   *    zclId.attr('genLevelCtrl', 0).key == 'currentLevel'
+   *    zclId.status(0).key == 'success'
+   *    zclId.dataType(32).key == 'uint8'
+   *
+   * @returns an array containing 2 entries. The first entry is the
+   *          property value, and the second entry is a printable version
+   * suitable for logging.
```
mrstegeman: nit: comment indentation is inconsistent

src/addons/zigbee-adapter/zb-classifier.js
benfrancis: Huh, you're right. I must have been getting this wrong for years! Very commonly capitalized ZigBee, but the Zigbee Alliance website doesn't capitalize the B. Wonder if they changed it at some point...

dhylands:
> I still don't understand the difference between "server side" and "client side" clusters in the spec...

I think of it kind of like X-Windows. An on-off switch is typically a server, and the gateway thru the dongle would be a client. However, if you had a wall switch that was controlling a light when, then the wall switch would be acting as a client.

If you look at the clusters/attributes on https://github.com/mozilla-iot/wiki/wiki/ZigBee-Attributes then the genOnOff input cluster on endpoint 1 (of the GE Dimmer Lamp Module) is acting like a server and that's what the gateway uses to control it.

I believe that the genOnOff output cluster on endpoint 2 is acting as a client and could be bound to another device. I'm not sure if this is just tied to the switch, or if it's tied to the genOnOff input cluster.

dhylands:
> How does the classification work? Does it currently just classify anything which supports the Level Control cluster as a multiLevelSwitch?

Yes. Any device which supports genLevelCtrl is required to also support genOnOff.

That means that the device is capable of dimming. It may be that if the user has a non-dimmable device plugged in that they might want to disable the dimming capability, but that would be a purely UI side thing.

dhylands:
> Presumably all of these would currently be mapped onto a multiLevelSwitch device type?

For the time being - yes. If we add support for color dimmable light, then it would probably be some new type with one or more additional properties. It's also possible that we could advertise multiple types, or that we advertise these as the same type and the UI uses the properties to distinguish the capabilities.

src/addons/plugin/device-proxy.js
dhylands: You are correct. I thought I tested this, but apparently I didn't.

src/addons/plugin/device-proxy.js
dhylands: I went back and forth on this. I think that having the properties as a separate dict is a better solution, but it requires a bunch more changes. Let's defer this to a separate issue.

dhylands: Although, the extra stuff that's being merged is the same fields that the derived class has in its own definition. So there should never be a collision (unless there is already a collision in the derived class). The Proxy classes just flatten it out into a single object instead of being spread out via inheritance.

dhylands:
> Is there any technical reason to support separate multiLevelSwitch and dimmableLight web thing types, or just that they could be represented by a different icon in the UI?

I'm going to say that depends on the UI. If the device advertises itself as a light, then its a light (probably something like a ZLL light using the ZHA protocol). A dimmable switch, on the other hand, might be a light, or it might be a fan, or it might be something else.

src/addons/zigbee-adapter/zb-adapter.js
```diff
@@ -261,7 +261,7 @@ class ZigBeeAdapter extends Adapter {
       // We're going to change something, so we might as well set the link
       // key, since it's write only and we can't tell if its been set before.
       configCommands.push(this.AT(AT_CMD.LINK_KEY,
-                                  {linkKey: 'ZigBeeAlliance09'}));
+                                  {linkKey: 'ZigbeeAlliance09'}));
```
mrstegeman: I think this one maybe should have stayed `ZigBeeAlliance09`

src/addons/plugin/property-proxy.js
dhylands: I don't think it does what you said. Is overlays, the contents of this.propertyDict with the contents of the base classes asDict method, neither of which includes this object's properties.

The same applies to your previous comment about deviceDict.

src/addons/zigbee-adapter/zb-adapter.js
dhylands: Argh. You are correct. (hangs head in shame).

PR #453, opened 2017-12-07
Add missing unit test and remove dead code.

mrstegeman:


src/db.js
```diff
@@ -340,17 +340,6 @@ var Database = {
     }
   },
 
-  /**
```
hobinjk: This feels like code we'll need very soon for resetting add-on settings or permanently deleting an add-on. I'd prefer for this not to be deleted and a follow-up issue filed for adding a tested DELETE /settings/:key route

PR #455, opened 2017-12-08
Move certs into ssl folder.

mrstegeman:
Resurrecting #143.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,10 @@ $ yarn
  The HTTPS server looks for privatekey.pem and certificate.pem. You can use a real certificate or generate a self-signed one by following the steps below.
 
  ```
- $ openssl genrsa -out privatekey.pem 2048
- $ openssl req -new -sha256 -key privatekey.pem -out csr.pem
- $ openssl x509 -req -in csr.pem -signkey privatekey.pem -out certificate.pem
+ $ mkdir -p ssl
+ $ openssl genrsa -out ssl/privatekey.pem 2048
+ $ openssl req -new -sha256 -key ssl/privatekey.pem -out ssl/csr.pem
+ $ openssl x509 -req -in ssl/csr.pem -signkey ssl/privatekey.pem -out ssl/certificate.pem
 ```
 
  Start the web server:
--- a/src/app.js
+++ b/src/app.js
@@ -56,6 +56,23 @@ if (fs.existsSync('notunnel')) {
   });
 }
 
+// Move certificates, if necessary. If this throws an error, let it bubble up.
+if (!fs.existsSync('ssl')) {
+  fs.mkdirSync('ssl');
+}
+
+if (fs.existsSync('privatekey.pem')) {
+  fs.renameSync('privatekey.pem', path.join('ssl', 'privatekey.pem'));
+}
+
+if (fs.existsSync('certificate.pem')) {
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -119,9 +119,9 @@ $ yarn
  The HTTPS server looks for privatekey.pem and certificate.pem. You can use a real certificate or generate a self-signed one by following the steps below.
 
  ```
- $ openssl genrsa -out privatekey.pem 2048
- $ openssl req -new -sha256 -key privatekey.pem -out csr.pem
- $ openssl x509 -req -in csr.pem -signkey privatekey.pem -out certificate.pem
+ $ openssl genrsa -out ssl/privatekey.pem 2048
```
hobinjk: Potentially prefix this with `mkdir ssl`

src/app.js
```diff
@@ -53,6 +53,19 @@ if (fs.existsSync('notunnel')) {
   });
 }
 
+// Move certificates, if necessary. If this throws an error, let it bubble up.
+if (fs.existsSync('privatekey.pem')) {
```
hobinjk: These need the `ssl` directory to already exist

tools/deploy-certificates.sh
```diff
@@ -6,9 +6,9 @@ gateway_dir="/home/pi/mozilla-iot/gateway"
 umask 077
 
 # Copy in the new certs.
-cp "$RENEWED_LINEAGE/chain.pem" "$gateway_dir/chain.pem"
-cp "$RENEWED_LINEAGE/cert.pem" "$gateway_dir/certificate.pem"
-cp "$RENEWED_LINEAGE/privkey.pem" "$gateway_dir/privatekey.pem"
+cp "$RENEWED_LINEAGE/chain.pem" "$gateway_dir/ssl/chain.pem"
```
hobinjk: This will likely also need a mkdir ssl based on whether ssl exists

tools/make-self-signed-cert.sh
```diff
@@ -2,6 +2,6 @@
 
 set -x
 
-openssl genrsa -out privatekey.pem 2048
-openssl req -new -sha256 -key privatekey.pem -out csr.pem -subj '/CN=www.toizom.com/O=MozillaIoT Gateway/C=US'
-openssl x509 -req -in csr.pem -signkey privatekey.pem -out certificate.pem
+openssl genrsa -out ssl/privatekey.pem 2048
```
hobinjk: `mkdir -p ssl` before

tools/manual-renew-certs.sh
```diff
@@ -121,11 +121,11 @@ rm -f "${pagekite_pidfile}"
 
 echo "Copying in new certificates."
 cp "${moziot_dir}/etc/live/${domain}/cert.pem" \
```
hobinjk: Potentially `mkdir -p ssl` before

mrstegeman:
This is actually going to become rather important if we decide to build a docker image. Having the certs in a separate directory will allow for data persistence. We'll also need to move the settings database, but that can be done separately.

dhylands:
Presumably then the database and log files should go into separate directories as well?

We should file a bug to handle this in regards to updates as well. i.e. An update will need to move the certs from their current directory to the new directory.

mrstegeman:
@dhylands this PR handles moving the certs. And yes, logs and such should go elsewhere.

PR #457, opened 2017-12-14
Add code to read discovered attributes

dhylands:
This PR adds the following features:

1 - A timeout mechanism. If a response isn't received from a ZigBee command, then the command will be reissued a few times before failing.
2 - If enabled, when doing a discover attributes, all of the attributes will also be read, and the value will be printed alongside the attribute information. This is especially useful when adding support (i.e. new classifiers) for new device types.
3 - Adds some support for printing out the internal zigbee command queue (typically only used for debugging).

### DIFF
--- a/src/addons/zigbee-adapter/zb-adapter.js
+++ b/src/addons/zigbee-adapter/zb-adapter.js
@@ -1134,7 +1134,7 @@ class ZigbeeAdapter extends Adapter {
   }
 
   discoverAttributesTimeout(frame) {
-    // Some the attributes fail to read during discover. I suspect that
+    // Some of the attributes fail to read during discover. I suspect that
     // this is because something related to the attribute hasn't been
     // initialized. Just report what we know.
     if (frame.zcl && frame.zcl.cmd == 'read') {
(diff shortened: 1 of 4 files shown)


src/addons/zigbee-adapter/zb-adapter.js
```diff
@@ -1075,21 +1112,48 @@ class ZigbeeAdapter extends Adapter {
...
   }
 
+  doneDiscoverAttributes(node) {
+    console.log('***** Discovery done for node:', node.id, '*****');
+    this.waitFrameTimeoutFunc = null;
+    this.discoveringAttributes = false;
+  }
+
+  discoverAttributesTimeout(frame) {
+    // Some the attributes fail to read during discover. I suspect that
```
mrstegeman: Some of*

src/addons/zigbee-adapter/zb-adapter.js
```diff
@@ -1193,6 +1262,15 @@ class ZigbeeAdapter extends Adapter {
     console.log(str);
   }
 
+  dumpCommands() {
+    console.log('Commands (' + this.cmdQueue.length + ')');
+    for (let idx in this.cmdQueue) {
```
mrstegeman: for...of, not for...in

PR #459, opened 2017-12-19
fixing final tarball filename ENV variable

artynet:
as described in the subject, this commit fixes the final tarball name. If not corrected, the process will not end up as it should.


artynet:
ok @dhylands, it was useful to me anyway since I tweaked  to cross-compile the modules for OpenWrt/LEDE. Maybe @andrenatal mentioned that to you.

Best, Arturo

PR #461, opened 2017-12-20
Verify files list in add-ons, and optional checksums.

mrstegeman:
In case something goes wrong during add-on installation, we can
use the files list and an optional SHA256SUMS file to verify the
add-on's expected contents.

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -22,6 +22,7 @@ const fs = require('fs');
 const path = require('path');
 const rimraf = require('rimraf');
 const tar = require('tar');
+const crypto = require('crypto');
 
 // Use webpack provided require for dynamic includes from the bundle  .
 const dynamicRequire = (() => {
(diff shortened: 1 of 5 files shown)


src/addon-manager.js
```diff
@@ -377,6 +400,51 @@ class AddonManager extends EventEmitter {
...
+      fname = path.join(addonPath, fname);
+      if (!fs.existsSync(fname)) {
+        const err = `Package ${manifest.name} missing file: ${fname}`;
+        console.error(err);
+        return Promise.reject(err);
+      }
+    }
+
+    // If a SHA256SUMS file is present, verify it.
+    const sumsFile = path.join(addonPath, 'SHA256SUMS');
```
dhylands: I think that we should provide a comment which shows the format of the SHA256SUMS file. It looks like the file format matches what sha256sum --check would expect, so it would be sufficient to just say that, and perhaps give an example of how to create the SHA256SUM file, something like:
```
cd some-addon-directory
rm -f SHA256SUMS
sha256sum * > SHA256SUMS
```
The rm is needed to ensure that the SHA256SUM file itself doesn't show up in the file. And then you can check using:
```
sha256sum --check SHA256SUMS
```

mrstegeman: That's exactly what I did in the example and GPIO add-ons, but I can add a comment here.

PR #467, opened 2017-12-21
Add timestamps to log lines

dhylands:


src/log-timestamps.js
```diff
@@ -0,0 +1,43 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+'use strict';
+
+const FUNCS = ['log', 'info', 'debug', 'error', 'warn'];
+
+function logWithTimestamp(realFunc, args) {
+  let currTime = new Date();
+  let prefix = currTime.getFullYear() + '-' +
```
mrstegeman: I **REALLY** wish there was a proper, non-module sprintf for JS.

PR #469, opened 2018-01-04
Improve logging timestamps

dhylands:
These changes cause the reported filename and line number to be
correct when running under jest.

### DIFF
--- a/src/log-timestamps.js
+++ b/src/log-timestamps.js
@@ -28,7 +28,7 @@ if (!console.constructor.hooked) {
   console.constructor.hooked = true;
 
   // BufferedConsole is used (under jest) when running multiple tests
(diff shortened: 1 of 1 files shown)


src/log-timestamps.js
```diff
@@ -11,33 +11,89 @@
...
-for (let func of FUNCS) {
-  let realFunc = console[func];
-  console[func] = function() {
-    logWithTimestamp(realFunc, arguments);
-  };
+if (!console.constructor.hooked) {
+  console.constructor.hooked = true;
+
+  // BufferedConsole is used (under jest) when running multiple tests
+  // CustomeConsole is used (under jest) when running a single test
```
dhylands: I see a typo here (s/b CustomConsole)

PR #470, opened 2018-01-05
Fix up plugin unloading.

dhylands:
These changes allow a plugin to be unloaded and immediately
reloaded as part of a test suite.

Essentially it makes sure that the promise returned by
Adapter.unload doesn't get resolved until the plugin has actually
closed the socket handles in PluginClient.unload(). Without these
changes, trying to reload a plugin during the 1/2 second delay
would cause a conflict. The 1/2 second delay is required during
normal non-test cases to ensure that the gateway has read the
plugin unloaded message before the socket is closed.

### DIFF
--- a/src/addons/plugin/addon-manager-proxy.js
+++ b/src/addons/plugin/addon-manager-proxy.js
@@ -116,11 +116,11 @@ class AddonManagerProxy extends EventEmitter {
           //       of the file because at the top of the file, otherwise we
           //       have circular requires and the addonManager object won't
           //       have been created yet.
-          let addonManager = require('../../addon-manager');
+          const addonManager = require('../../addon-manager');
           // NOTE: The call to getPlugin will only succeed when doing
           //       inproc IPC, since getPlugin reaches into server-side
           //       data structures, and we're running on the client.
-          let plugin = addonManager.getPlugin(this.pluginClient.pluginId);
+          const plugin = addonManager.getPlugin(this.pluginClient.pluginId);
           if (plugin && plugin.unloadedRcvdPromise) {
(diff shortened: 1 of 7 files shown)


src/addons/plugin/addon-manager-proxy.js
```diff
@@ -98,15 +99,49 @@ class AddonManagerProxy extends EventEmitter {
...
+        if (config.get('ipc.protocol') === 'inproc') {
+          // When we're testing, we run in the same process and we need
+          // to close the sockets before the adapter.unload promise is
+          // resolved. So we hook into the plugin unloadedRcvdPromise.
+
+          // NOTE: We need to put this require here rather than at the top
+          //       of the file because at the top of the file, otherwise we
+          //       have circular requires and the addonManager object won't
+          //       have been created yet.
+          let addonManager = require('../../addon-manager');
```
mrstegeman: nit: `let addonManager` and `let plugin` below should both use `const`.

src/addons/plugin/plugin.js
```diff
@@ -74,10 +76,26 @@ class Plugin {
       case Constants.PLUGIN_UNLOADED:
         this.shutdown();
         this.pluginServer.unregisterPlugin(msg.data.pluginId);
-        deferredUnload = this.deferredUnload;
-        if (deferredUnload) {
-          this.deferredUnload = null;
-          deferredUnload.resolve();
+        if (this.unloadedRcvdPromise) {
+          let socketsClosedPromise = new Deferred();
```
mrstegeman: This could use `const` as well.

PR #471, opened 2018-01-05
Added some debugging to plugin IPC

dhylands:
As part of chasing down the bugs exposed by #470 I added
the following debug code to the IPC class. Since this
only adds overhead for bringing up and down ipc sockets
it will have negligible effect on performance and I'd like
to keep it.

### DIFF
--- a/src/addons/plugin/ipc.js
+++ b/src/addons/plugin/ipc.js
@@ -4,13 +4,16 @@ const config = require('config');
 const fs = require('fs');
 const nanomsg = require('nanomsg');
 
+require('string.prototype.padstart').shim();
+require('string.prototype.padend').shim();
+
 var appInstance = require('../../app-instance');
 
 const DEBUG = false;
 const DEBUG_MSG = false;
 
-var boundAddrs = new Set();
-var connectedAddrs = new Set();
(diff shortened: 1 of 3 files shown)


src/addons/plugin/ipc.js
```diff
@@ -7,14 +7,22 @@ const nanomsg = require('nanomsg');
 var appInstance = require('../../app-instance');
 
 const DEBUG = false;
+const DEBUG_MSG = false;
+
+var boundAddrs = new Set();
```
mrstegeman: `boundAddrs` and `connectedAddrs` should be `const`.

src/addons/plugin/ipc.js
```diff
@@ -34,26 +42,87 @@ class IpcSocket {
     }
     this.ipcAddr = this.protocol + '://' + this.ipcFile;
 
+    this.logPrefix = 'IpcSocket' + ('   ' + this.socketId).slice(-3) + ': ' +
```
mrstegeman: Instead of doing all this manual padding everywhere, you can use `padStart`:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

PR #477, opened 2018-01-08
Fix addon-loader to work properly

dhylands:
Plugins are loaded from the gateway/build tree. Normally __dirname
reflects this, but since the addon-loader is launched separately
from the gateway, it may differ. These changes cause the addon loader
to be told explicitly which directory to load a plugin from.


dhylands:
I'm planning on it. gpio happens to work because its not tagged as a plugin. example-adapter has never worked as a plugin.

PR #479, opened 2018-01-08
Add code to validate the package.json file included with an addon.

dhylands:


src/addon-manager.js
```diff
@@ -369,6 +368,58 @@ class AddonManager extends EventEmitter {
...
+    }
+  }
+
+  /**
+   * @method validateManifest
+   *
+   * Verifies that the manifest looks valid. We only need to validate
+   * fields that we actually use.
+   */
+
```
mrstegeman: nit: extra newline unnecessary here

src/addon-manager.js
```diff
@@ -369,6 +368,58 @@ class AddonManager extends EventEmitter {
...
+  validateManifest(manifest) {
+    let manifestTemplate = {
+      name: '',
```
mrstegeman: We should probably require `version` as well, in preparation for handling add-on updates. I will take care of adding support for the `files` array.

src/addon-manager.js
```diff
@@ -462,6 +513,15 @@ class AddonManager extends EventEmitter {
       }
     }
 
+    // Verify that important fields exist in the manifest
+    let err = this.validateManifest(manifest);
+    if (err) {
+      console.error(
+        `Error found in manifest for ${packageName}\n${err}`);
```
mrstegeman: nit: pull the error string into a constant instead of typing it twice

src/test/integration/addons-test.js
```diff
@@ -9,22 +9,62 @@ const {
...
+const jsonfile = require('jsonfile');
+const path = require('path');
+const pFinal = require('../promise-final');
+
+const testManifestFilename = path.join(__dirname, '../..',
+                                       config.get('addonManager.path'),
+                                       'test-adapter', 'package.json');
+
+const testManifest = {
+  name: 'test-adapter',
```
mrstegeman: Add `version`.

src/test/integration/addons-test.js
```diff
@@ -9,22 +9,62 @@ const {
...
+      min: 1,
+      max: 1
+    },
+    enabled: true,
+  },
+};
+
+function copyManifest(manifest) {
+  // This essentially does a deep copy.
+  return JSON.parse(JSON.stringify(manifest));
```
mrstegeman: This would be more efficient:
`return Object.assign({}, manifest);`

src/test/integration/addons-test.js
```diff
@@ -193,4 +235,45 @@ describe('addons', function() {
...
+    let manifest = copyManifest(testManifest);
+    delete manifest.moziot.api.min;
+    expect(await loadSettingsAdapterWithManifest(manifest)).toBeTruthy();
+  });
+
+  it('Fail package.json with non-numeric moziot.api.min', async () => {
+    let manifest = copyManifest(testManifest);
+    manifest.moziot.api.min = 'abc';
+    expect(await loadSettingsAdapterWithManifest(manifest)).toBeTruthy();
+  });
```
mrstegeman: Add a test for `version`.

src/test/integration/addons-test.js
dhylands: Unfortunately, that only does a shallow copy and modifications made to nested objects in the returned manifest cause changes to testManifest. Using the JSON trick is a cheap deep copy. Object.assign is NOT a deep copy.

mrstegeman: Ah, good point.

src/addon-manager.js
dhylands: That's probably because I added the console.error intending to remove it before I checked it in. I'll double check.

PR #480, opened 2018-01-09
Validate files array in package.json.

mrstegeman:


src/addon-manager.js
```diff
@@ -383,10 +383,16 @@ class AddonManager extends EventEmitter {
...
         }
         if (typeof(objectVal) === 'object') {
-          let err = this.validateManifestObject(prefix + key + '.',
-                                                objectVal, templateVal);
-          if (err) {
-            return err;
+          if (Array.isArray(objectVal)) {
+            if (templateVal.length > 0 && objectVal.length === 0) {
+              return `Expecting ${prefix}${key} to be a non-empty array.`;
+            }
```
dhylands: Should this also validate the type of the array entries? Currently, it would accept files: [1, 2, 3]
We can probably make the simplifying assumption that each element of the array should be of the same type.
We probably don't need to support empty arrays.

mrstegeman: Sure, we can do that.

PR #481, opened 2018-01-09
Add required API field to addons list.

mrstegeman:
This allows us to check each discovered add-on against the running
API level.

### DIFF
--- a/src/test/integration/addons-test.js
+++ b/src/test/integration/addons-test.js
@@ -188,6 +188,9 @@ describe('addons', function() {
     expect(list[0]).toHaveProperty('description');
     expect(list[0]).toHaveProperty('version');
     expect(list[0]).toHaveProperty('url');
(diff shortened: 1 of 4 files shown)


src/test/integration/addons-test.js
```diff
@@ -188,6 +188,9 @@ describe('addons', function() {
     expect(list[0]).toHaveProperty('description');
     expect(list[0]).toHaveProperty('version');
     expect(list[0]).toHaveProperty('url');
+    expect(list[0]).toHaveProperty('api');
```
dhylands: Shouldn't this be moziot.api ?

mrstegeman: No, that's the `package.json` format. The addons list format is different.

dhylands: ok - this is just failing then because the other change wasn't merged yet

PR #484, opened 2018-01-10
Kill plugin subprocess on disable/uninstall.

mrstegeman:
Without this, the plugin processes were never being killed off (except when the gateway shut down), so on every disable-enable cycle, or uninstall-install cycle, multiple processes would end up running for the same add-on.

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -752,6 +752,7 @@ class AddonManager extends EventEmitter {
       return Promise.resolve();
     }
 
+    const plugin = this.getPlugin(packageName);
     let adapters = this.getAdaptersByPackageName(packageName);
     let unloadPromises = [];
     for (const a of adapters) {
@@ -760,11 +761,13 @@ class AddonManager extends EventEmitter {
       this.adapters.delete(a.id);
     }
 
+    // Give the process 5 seconds to exit before killing it.
     const cleanup = () => {
-      const plugin = this.getPlugin(packageName);
-      if (plugin) {
-        plugin.kill();
-      }
+      setTimeout(() => {
+        if (plugin) {
+          plugin.kill();
+        }
+      }, 5000);
     };
 
     return Promise.all(unloadPromises).then(() => cleanup(), () => cleanup());
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -760,12 +760,14 @@ class AddonManager extends EventEmitter {
       this.adapters.delete(a.id);
     }
 
-    const plugin = this.getPlugin(packageName);
-    if (plugin) {
-      plugin.kill();
-    }
+    const cleanup = () => {
+      const plugin = this.getPlugin(packageName);
+      if (plugin) {
+        plugin.kill();
+      }
+    };
 
-    return Promise.all(unloadPromises);
+    return Promise.all(unloadPromises).then(() => cleanup(), () => cleanup());
   }
 
   /**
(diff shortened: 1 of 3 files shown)


mrstegeman:
The adapters are all unloaded first, though I wasn't waiting for the promises.

The problem is that unloading the adapter never shuts down the plugin process.

dhylands:
#487 makes it unload the adapter when it should. I don't mind having it try to kill the process if the normal unload fails.

src/addon-manager.js
```diff
@@ -760,12 +760,14 @@ class AddonManager extends EventEmitter {
...
-    }
+    const cleanup = () => {
+      const plugin = this.getPlugin(packageName);
+      if (plugin) {
+        plugin.kill();
+      }
+    };
 
-    return Promise.all(unloadPromises);
+    return Promise.all(unloadPromises).then(() => cleanup(), () => cleanup());
```
dhylands: I still we think a delay here. The client side of the plugin waits for 1/2 second after sending the plugin unloaded message before closing the IPC connections (because the client doesn't know that the server received the message and closing will cause the message to disappear).

src/addon-manager.js
```diff
@@ -760,12 +760,14 @@ class AddonManager extends EventEmitter {
...
     }
 
-    const plugin = this.getPlugin(packageName);
-    if (plugin) {
-      plugin.kill();
-    }
+    const cleanup = () => {
+      const plugin = this.getPlugin(packageName);
+      if (plugin) {
+        plugin.kill();
```
dhylands: I think we should also log something here, like plugin didn't shutdown cleanly, killing

src/addon-manager.js
dhylands: I realized while I was shoveling the driveway that we also reuse the Plugin object.
https://github.com/mozilla-iot/gateway/blob/master/src/addons/plugin/plugin-server.js#L99

We should probably create a new one and replace the old one. That way if the old one still has timers etc they'll still be valid and we won't wind up killing the new process we just launched

src/addon-manager.js
```diff
@@ -760,11 +761,13 @@ class AddonManager extends EventEmitter {
...
     const cleanup = () => {
-      const plugin = this.getPlugin(packageName);
-      if (plugin) {
-        plugin.kill();
-      }
+      setTimeout(() => {
+        if (plugin) {
+          plugin.kill();
+        }
+      }, 5000);
```
dhylands: The thing about reusing the plugin object was that if the user re-enables the plugin before the 5 seconds expires, then we'll wind up killing the new process.

If cleanup did something like this:
```
    const cleanup = () => {
      let process = plugin.process;
      setTimeout(() => {
        process.kill();
      }, 5000);
    };
```
then it wouldn't matter. Otherwise this looks good.

src/addons/plugin/plugin.js
```diff
@@ -39,16 +39,21 @@ class Plugin {
     this.ipcSocket.bind();
     this.exec = '';
     this.execPath = '.';
-    this.process = null;
+
+    // Make this a nested object such that if the Plugin object is reused,
+    // i.e. the plugin is disabled and quickly re-enabled, the gateway process
+    // can maintain a proper reference to the process object.
+    this.process = {p: null};
```
dhylands: This seems to be causing a test failure

dhylands: The plugin-test.js is looking for plugin.process.on

mrstegeman: Fixed.

PR #486, opened 2018-01-10
Make plugins read their config from the database.

mrstegeman:


mrstegeman:
It's already difficult, because the user has to regenerate the SHA256SUMS file if they modify package.json. This PR is just making the add-ons honor the architecture of the add-ons system, that way, when we get around to adding a configuration UI, the add-ons do what they're supposed to.

dhylands:
Actually, we could just write a script that allows the config to be shown or changed.

mrstegeman:
I'll add one to this PR shortly.

tools/config-editor.py
```diff
@@ -0,0 +1,110 @@
...
+            sys.exit(1)
+    else:
+        print(config_str)
+
+    c.close()
+    conn.close()
+
+
+if __name__ == '__main__':
+    main()
```
dhylands: We should probably test to make sure that the json is valid and relaunch the editor if it isn't.

If they exit the editor without editing the file then we shouldn't update or relaunch.

dhylands: But I'm also happy to make such improvements later. I approved the PR earlier and still think its good.

mrstegeman: Line 83. The new JSON is loaded and parsed. ValueError is thrown if the JSON fails to parse.

PR #492, opened 2018-01-12
Color Light (onOffColorLight) Thing Support

### LINKED ISSUE #485: On/off Color Light UI
As a user, I want to see the detail view of my color-supporting bulb and control its color and on-off state from one page. Note that I'm using the term "HSBBulb" instead of "Colo(u)redDimmableBulb" because I'm avoiding reliving the events of 1776.

Current WIP:
screen shot 2018-01-10 at 4 42 42 pm

hobinjk:
Close #485 

Todo:
- [x] Remove d3
- [x] Get an icon and name in "Add Thing" dialog


hobinjk:
I changed the way the bulb works to be white if on and black if off which looks better with colors but worse with pure white. I think what I'll end up doing is set the stroke based on intensity and the fill based on on/off

static/js/things.js
```diff
@@ -67,6 +67,11 @@ var ThingsScreen = {
               // eslint-disable-next-line no-unused-vars
               var newOnOffSwitch = new OnOffSwitch(description);
               break;
+            case 'colorLight':
```
benfrancis: This isn't one of the [Web Thing Types](https://iot.mozilla.org/wot/#dimmablecolorlight-type) defined so far. Do you think we need an onOffColorLight type as well?

static/js/color-light.js
```diff
@@ -0,0 +1,265 @@
...
+    this.element.addEventListener('click', this.handleClick.bind(this));
+  }
+  return this;
+}
+
+ColorLight.prototype = Object.create(OnOffSwitch.prototype);
+
+ColorLight.prototype.iconView = function() {
+  let color = this.properties.color;
+  return `<div class="color-light" style="background: ${color};">
```
benfrancis: I feel like there's a better way of doing this rather than hard coding the SVG markup in JavaScript. Could the icon be a background image or an embedded SVG file for example (it's a bit tricky to tell without being able to inspect the live HTML)?

static/js/thing-detail-layout.js
```diff
@@ -0,0 +1,36 @@
+(function() {
+
+function ThingDetailLayout(elements) {
```
benfrancis: nit: Could this just be ThingDetail rather than ThingDetailLayout?

static/js/thing-detail-layout.js
```diff
@@ -0,0 +1,36 @@
...
+
+  let svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+  svg.classList.add('thing-detail-layout-links');
+  svg.setAttribute('width', scale * 2);
+  svg.setAttribute('height', scale * 2);
+
+  for (let i = 0; i < elements.length; i++) {
+    let angle = i/elements.length * 2 * Math.PI;
+    let x = scale * Math.cos(angle);
+    let y = scale * Math.sin(angle);
```
benfrancis: Nice!

benfrancis:
> I changed the way the bulb works to be white if on and black if off which looks better with colors but worse with pure white. I think what I'll end up doing is set the stroke based on intensity and the fill based on on/off

I'm not sure about this, but I'm not sure what you mean by intensity (brightness?) or which bit you're suggesting to be black? Examples might help. My concerns would be consistency with other things in the things screen and whether full black looks too severe for the visual design.

static/js/color-light.js
hobinjk: IIRC the icon has to be in the HTML for its stroke and fill to be modified by classes/js. The main things that I'm pretty sure will not work are making it a background image or using ``.

static/js/thing-detail-layout.js
hobinjk: As a function this lays out a collection of a thing's detail circles. Admittedly the only reason this exists is that it was necessary to use d3, although I like the idea of being able to modularly change the layout from wide to tall (or whatever ends up being necessary)

hobinjk:
Intensity is a close analog to brightness from https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color

hobinjk:
WebComponents would definitely be the best way of accomplishing this type of functionality and would allow all of the Thing details to be modular and let the SVG icon be in a better place, i.e. not in a multiline javascript string.

benfrancis:
OK, I've been able to test this now thanks to the virtual things adapter.

- Given this is an on/off light let's have ON displayed when on and OFF displayed when off, as with on/off switches.
- I could maybe live with the white/grey for on/off bulb icon, but what concerns me about this modification to the design is that when the color is set to white and the bulb is off, it has a white background like the "on" state of on/off switches. This seems like a common case which is going to look inconsistent and confusing at a glance. What issue with the design are you trying to work around? Maybe there's another solution.
- Scaling on mobile - when viewing the expanded view on a smartphone the properties disappear off the screen. Me may be able to fix this by shortening the connecting lines but I suspect we may need to move the position of the circles up or down slightly to make to make them fit. I've attached the design I had for the expanded smart plug UI (the maths may get trickier when we have more than two properties of a Thing though...). I've been trying to get the UI to work down to a resolution of 320x480, though you might argue that's a bit extreme.
smart_plug_mobile
- Slide 24 of the spec says "Tapping on a Thing shows you an expanded view of that Thing (except for very simple things with only one property like the on/off switch which can be interacted with directly)." Given the onOffColorLight has more than one property, that means tapping it should take you to the expanded view. I'm not entirely happy with this design because it's inconsistent and means an on/off requires an extra step, but currently the only way to reach the expanded view is through the floorplan which seems bad. Can you change this to link to the expanded view and maybe we can discuss an improvement on this afterwards? (e.g. I thought about requiring a long press to get to the expanded view, but that's less discoverable.).
- There seems to be a lot of logging of property changes going on in console.log, is that left over from debugging?
- I'll upload some icons for the SVG view. Ideally I'd prefer the design and interactions of things in the Things view and the Floorplan view to be the same, and have nice animations when things expand out. It's just tricky to get the mix of SVG and HTML right and position things properly.

dhylands:
I normally just use the "squash and merge" button.

benfrancis:
I just noticed a slight alignment problem in the Things view, onOffColorLight and multiLevelSwitch appearing slightly higher than onOffSwitch for some reason.

screen shot 2018-01-18 at 20 13 04

hobinjk:
Updated to address review comments. I've reverted to the spec since the On/off label solves the main problem I was thinking about.

benfrancis:
* font-weight bold

hobinjk:
Review comments addressed, thanks! As far as the intermittent error goes, I think there's enough refactoring of that code path in the dimmableColorLight PR that it shouldn't continue happening.

hobinjk:
Oh wait no, there's a bug with color on the Things screen

PR #497, opened 2018-01-17
Add ZWave smartplug support

dhylands:


src/addons/zwave-adapter/zwave-classifier.js
```diff
@@ -10,64 +10,158 @@
...
+    const levelValueId =
+      node.findValueId(COMMAND_CLASS_SWITCH_MULTILEVEL, 1, 0);
+    if (levelValueId) {
+      node.type = Constants.THING_TYPE_MULTI_LEVEL_SWITCH;
+      this.addProperty(
+        node,                   // node
+        'level',                // name
+        {                       // property decscription
+          type: 'number',
+          unit: '%',
```
mrstegeman: Unit should be `'percent'` according to spec.

src/addons/zwave-adapter/zwave-classifier.js
```diff
@@ -10,64 +10,158 @@
...
+
+    let powerValueId = node.findValueId(COMMAND_CLASS_METER, 1, 8);
+    if (powerValueId) {
+      node.type = Constants.THING_TYPE_SMART_PLUG;
+      this.addProperty(
+        node,                   // node
+        'instantaneousPower',   // name
+        {                       // property decscription
+          type: 'number',
+          unit: 'watts',
```
mrstegeman: Unit should be `'watt'` according to spec.

src/addons/zwave-adapter/zwave-classifier.js
```diff
@@ -10,64 +10,158 @@
...
+
+    let voltageValueId = node.findValueId(COMMAND_CLASS_METER, 1, 16);
+    if (voltageValueId) {
+      node.type = Constants.THING_TYPE_SMART_PLUG;
+      this.addProperty(
+        node,                   // node
+        'voltage',              // name
+        {                       // property decscription
+          type: 'number',
+          unit: 'volts',
```
mrstegeman: Unit should be `'volt'` according to spec.

src/addons/zwave-adapter/zwave-classifier.js
```diff
@@ -10,64 +10,158 @@
...
+
+    let currentValueId = node.findValueId(COMMAND_CLASS_METER, 1, 20);
+    if (currentValueId) {
+      node.type = THING_TYPE_SMART_PLUG;
+      this.addProperty(
+        node,                   // node
+        'current',              // name
+        {                       // property decscription
+          type: 'number',
+          unit: 'amps',
```
mrstegeman: Unit should be `'ampere'` according to spec.

mrstegeman:
https://benfrancis.github.io/wot/#web-thing-types

src/addons/zwave-adapter/zwave-classifier.js
dhylands: The spec changed since I wrote that. It used to reference an ISO document which used %, but I'll change it since the updated spec now show 'percent'. Weird that the spec uses the singular version of the others since its almost never 1.

benfrancis:
@dhylands The idea (proposed by EVRYTHNG) is to use SI units in their full word form (as opposed to an abbreviated form or symbol). From what I could tell these are the SI strings for these units, but I couldn't find a very good reference for this so I may have made mistakes.

Does anyone know of a good place to look up standard SI units and their representation?

mrstegeman:
Here are a couple:

https://en.wikipedia.org/wiki/SI_base_unit
https://en.wikipedia.org/wiki/SI_electromagnetism_units

benfrancis:
> Here are a couple:
> 
> https://en.wikipedia.org/wiki/SI_base_unit
> https://en.wikipedia.org/wiki/SI_electromagnetism_units

Wikipedia is one of the sources I was using, no mention of "percent" though.

mrstegeman:
"percent" isn't really a unit, but for spec purposes, it makes sense to use the full word, given that that's what we use for everything else, i.e. "volt", "watt", etc.

dhylands:
I don't remember exactly which document I referred to, but these are similar:
http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication330e2008.pdf (page 44)
http://unitsofmeasure.org/ucum.html#para-29
and refers to several dimensionless units, including percent, along with the symbol to use.

PR #500, opened 2018-01-17
Fix crash when sending data to closed websocket.

mrstegeman:
I hit a case where a websocket error event was thrown without a
subsequent close event. Handle that.


mrstegeman:
@hobinjk Done.

hobinjk:
Thanks!

PR #501, opened 2018-01-17
Multi Level Switch UI

hobinjk:
This also incorporates #492.

Note that both need SVG icons/outlines for their svgView. @benfrancis Can you generate them from your existing assets? If not, I can make them from what I have on hand.

### DIFF
--- a/static/js/level-detail.js
+++ b/static/js/level-detail.js
@@ -1,0 +1,49 @@
+/**
+ * LevelDetail
+ *
+ * A bubble showing the level of a thing
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+function LevelDetail(thing) {
+  this.thing = thing;
+}
+
+LevelDetail.prototype.attach = function() {
+  this.level = this.thing.element.querySelector('.level-input');
+  this.level.addEventListener('input', () => {
+    this.thing.updateLevel(this.level.value);
+  });
+  this.level.addEventListener('change', this.set.bind(this));
+}
+
+LevelDetail.prototype.view = function() {
+  let level = this.thing.properties.level;
+
+  return `<div class="thing-detail-container">
+    <div class="thing-detail">
+      <div class="thing-detail-contents">
+        <form class="level">
+          <input type="range" min="0" max="100" value="${level}"
+                 class="level-input"/>
+        </form>
+      </div>
+    </div>
+    <div class="thing-detail-label">Level</div>
+  </div>`;
+};
+
+LevelDetail.prototype.update = function() {
+  if (this.thing.properties.level == this.level.value) {
+    return;
+  }
+  this.level.value = this.thing.properties.level;
+};
+
+LevelDetail.prototype.set = function() {
+  this.thing.setLevel(this.level.value);
+};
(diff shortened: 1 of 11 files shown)


static/js/level-detail.js
```diff
@@ -0,0 +1,49 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+function LevelDetail(thing) {
+  this.thing = thing;
+}
+
+LevelDetail.prototype.attach = function() {
```
benfrancis: What do you think about creating a ThingProperty base class and having LevelProperty extend ThingProperty? That's the approach I've been taking with the SmartPlug UI.

static/js/on-off-detail.js
```diff
@@ -0,0 +1,49 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+'use strict';
+
+function OnOffDetail(thing) {
+  this.thing = thing;
+}
+
+OnOffDetail.prototype.attach = function() {
```
benfrancis: Similarly, could this be OnOffProperty?

static/js/thing.js
```diff
@@ -32,6 +32,7 @@ var Thing = function(description, format) {
   // Parse base URL of Thing
   if (description.href) {
     this.href = new URL(description.href, App.ORIGIN);
+    this.id = this.href.pathname.split('/').pop();
```
benfrancis: Just a warning that this may not work once we allow adding things by URL... but we can cross that bridge when we come to it.

static/js/multi-level-switch.js
```diff
@@ -0,0 +1,181 @@
...
+    this.showOn();
+  } else {
+    this.showOff();
+  }
+};
+
+MultiLevelSwitch.prototype.updateLevel = function(level) {
+  level = parseFloat(level);
+  this.properties.level = level;
+  // TODO update ui thingy
```
benfrancis: ?

static/js/multi-level-switch.js
```diff
@@ -0,0 +1,181 @@
...
+  let blank = '#5d9bc7';
+  let bar = 'white';
```
benfrancis: These constants should maybe live somewhere else?

benfrancis:
- The percentage label should be have font-weight bold.

hobinjk:
Updated for review comments, thanks!

static/js/multi-level-switch.js
```diff
@@ -12,6 +12,11 @@
 
 /* globals LevelDetail, OnOffDetail, OnOffSwitch, Thing, ThingDetailLayout */
 
+const MULTI_LEVEL_SWITCH_OFF_BAR = 'white';
+const MULTI_LEVEL_SWITCH_OFF_BLANK = '#89b6d6';
+const MULTI_LEVEL_SWITCH_ON_BAR = '#5d9bc7';
+const MULTI_LEVEL_SWITCH_ON_BLANK = 'white';
```
benfrancis: It would better if these were CSS classes that are set on the element rather than putting the color codes in JavaScript.

benfrancis: Unless there's a particular reason that makes that very difficult..?

hobinjk: These are used in the definition of a linear gradient. I could try redesigning the level display to be a div inside of a div which could then be properly styled using classes instead

benfrancis: Oh I see, clever. I don't mind if you'd prefer to leave it this way.

PR #502, opened 2018-01-17
Remove Zigbee and Z-Wave adapters from tree.

mrstegeman:


image/add-gateway.sh
```diff
@@ -145,6 +146,19 @@ main() {
     sudo mkdir -p ${MOZILLA_IOT_DIR}
     sudo tar xf ${GATEWAY_TARBALL} -C ${MOZILLA_IOT_DIR}
 
+    # Install default add-ons
+    sudo mkdir -p "${ADDONS_DIR}"
+    tempdir=$(mktemp -d)
+    zigbee_url="https://github.com/mozilla-iot/zigbee-adapter/releases/download/v0.3.0-pre1/zigbee-adapter-0.3.0-pre1.tgz"
```
dhylands: I think that this is fine for now. Please create an issue to have this do something a bit more intelligent like use the most recent from the addons list or something. Otherwise we'll have to remember to edit this file every time we do a release.

mrstegeman: Done.

dhylands:
Looks good

Issue #447, opened 2017-12-05
Investigate using a ServiceWorker for caching and/or offline support


hobinjk:
Picking up from #366 

ServiceWorker caching is good but requires tooling to generate the pre-cached list of files. Offline support through a SW would also be nice and might need specific code support.

hobinjk:
I figured out that I don't actually need to pre-cache. Right now I instead work off a whitelist of known static directories with updating of the cache happening automatically in the background. There are still a few bugs but this does speed things up from the baseline of no caching.

NB: WebPageTest considers the background caching to be part of the page's loading process and gives an artificially high loading time. The time I'm currently using is First Interactive but Start Render also looks reasonable.

Baseline: [18.2, 0.8](https://www.webpagetest.org/result/180122_GW_caa6a5ee55d730beb70d579595fa9b02/)
SW: [17.7, 4.0](https://www.webpagetest.org/result/180122_R8_f6f40c2ce3857927841bd1887dd07423/)

hobinjk:
SW with caching enabled: [16.6, 1.5](https://www.webpagetest.org/result/180123_NR_589a9c1ea7400934aabd0050db8286f9/)

PR #503, opened 2018-01-18
Fix issues with caching static files.

### LINKED ISSUE #454: Too much is cached?
Here's the scenario. My gateway is plugged in via ethernet. I had previously connected to my gateway using gateway.local and gotten back a list of things (I hadn't added any so it was empty).

Then I rebooted in a way that caused the wifi setup to become active. When I browse to http://gateway.local from my desktop using the ethernet interface, I still get back an empty things screen, even though the gateway software isn't even running. If I browsed instead directly to the IP address using http://192.168.1.79/ then I get back the wifi setup screen, which I was expecting. I was also able to browse to http://gateway.local:8080 and get the wifi setup screen. Pressing reload on the empty things screen just gives an error, since it's already been redirected to https://gateway.local/things and the setup code isn't listening to https requests.

Looking at the the network inspector (using developer tools), it shows that all of the results were retrieved using cached data. the final few fetches (like from /things/) showed that zero bytes were transferred, but the UI didn't give any indication that something was amiss. Presumably we should at least have / setup so that it doesn't cache it.

mrstegeman:
The gateway was setting a max age of 14d for static resources.
However, it also sets a weak ETag header by default, which is
better for caching purposes when we're changing files, so we'll
now just rely on the ETag.

I tested this and verified that resources were properly cached
until modified.

Fixes #454

### DIFF
--- a/src/controllers/root_controller.js
+++ b/src/controllers/root_controller.js
@@ -23,7 +23,7 @@ RootController.get('/', TunnelSetup.isTunnelSet,
   function(request, response) {
     response.sendFile('index.html', {
       root: Constants.VIEWS_PATH,
-      maxAge: '14d'
+      maxAge: 0,
     });
   }
 );
--- a/src/router.js
+++ b/src/router.js
@@ -33,8 +33,9 @@ var Router = {
     // Compress all responses larger than 1kb
     app.use(compression());
 
-    // First look for a static file
-    app.use(express.static(Constants.STATIC_PATH, {maxAge: '14d'}));
+    // First look for a static file. Turn off max-age and let the default ETag
+    // handle caching.
+    app.use(express.static(Constants.STATIC_PATH, {maxAge: 0}));
 
     // Content negotiation middleware
     app.use(function(request, response, next) {


hobinjk:
The contributing factor to the 15 second loading time isn't throughput, it's latency. The ETag header prevents the file from being sent but still costs a full round trip. Using maxAge prevents the need for the round trip but means that you have to use a hard refresh to signal that you don't want to use the cache.

In terms of real-world performance:
[WebPageTest of master, 15s -> 3.7s repeat](https://www.webpagetest.org/result/180118_HF_095a6a536221460f46a9a1e7a08979d8/)
[WebPageTest of this PR, 15s -> 14.1s repeat](https://www.webpagetest.org/result/180118_5C_3fc1c8794b2f4e66b678f2703455f7ac/)

In general, I have yet to have any issues with overzealous caching so I don't really see where #454 is coming from. I'll put a comment describing my workflow in that issue in case it helps.

mrstegeman:
Yikes! @dhylands and I have both seen this issue several times, though, especially when modifying some of the static JS files. It probably won't be a huge problem for our users (other than after upgrades, for up to 2 weeks), but it certainly makes development a pain.

hobinjk:
A ServiceWorker would be able to properly fix this, serving from the cache initially and updating in the background, e.g. https://serviceworke.rs/strategy-cache-and-update.html

#447 exists to track writing a SW generator and is currently up for grabs if you'd like to work on that.

dhylands:
Serving JS from the cache when its mismatched can cause crashes. I've seen this once on my RPi where I had a previous version installed, and then updated the software, but had some of the JS coming from the cache and making REST requests which no longer existed rather than using the new code which would have used the new REST requests.

PR #518, opened 2018-01-19
Add a back button to expanded thing UI.

### LINKED ISSUE #512: Expanded Thing UI - Back Button
The menu button in the Expanded Thing UI should be turned into a back button like in the Add Thing UI, as per the mockups.

The one tricky thing is that there are multiple entry points to this view so the back button may need to navigate back in browser history rather than a particular URL.

mrstegeman:
Fixes #512

### DIFF
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -207,7 +207,7 @@ h1, h2, h3, h4, h5, h6 {
 #back-button:hover,
 #add-button:active,
 #create-rule-button:active,
-#discover-addons-button:active, {
+#discover-addons-button:active {
   opacity: 1;
 }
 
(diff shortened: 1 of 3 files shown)


static/css/app.css
```diff
@@ -185,15 +185,26 @@ h1, h2, h3, h4, h5, h6 {
...
 }
 
-#menu-button::-moz-focus-inner {
+#menu-button::-moz-focus-inner,
+#back-button::-moz-focus-inner {
   border: 0;
 }
 
 #menu-button:hover,
+#back-button:hover,
```
benfrancis: Nit: This doesn't seem to be working for some reason.

mrstegeman: Fixed. There was a trailing comma further down.

PR #520, opened 2018-01-19
Add thing name to expanded UI.

### LINKED ISSUE #513: Expanded Thing UI - Thing Name
Show the name of a Thing in the Expanded Thing UI, as per the mockups. Otherwise there's no way to tell what thing you're controlling from its expanded view.

mrstegeman:
In addition:
* Fix some scrolling issues.
* De-duplicate code for simple thing views.
* Small bug fixes.

Fixes #513

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -193,3 +193,25 @@
   font-weight: bold;
 }
 
+#thing-title.hidden {
+  display: none;
+}
+
+#thing-title {
+  width: 100%;
+  text-align: center;
+  position: fixed;
+  top: 1.8rem;
+}
+
+#thing-title-icon {
+  height: 4rem;
+  position: relative;
+  top: 1.3rem;
+  padding-right: 1rem;
+}
+
+#thing-title-name {
+  font-size: 1.6rem;
+  color: #fff;
+}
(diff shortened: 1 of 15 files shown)


static/css/thing.css
```diff
@@ -179,3 +200,16 @@
   font-weight: bold;
 }
 
+.thing-title {
+  width: 100%;
+  text-align: center;
+  position: fixed;
+  top: 20px;
```
hobinjk: Should be in rem

static/css/thing.css
```diff
@@ -179,3 +200,16 @@
...
+}
+
+.thing-title-icon {
+  height: 40px;
```
hobinjk: These should all be in rem, I also think that 40px is slightly off for vertical alignment. For example, 
screen shot 2018-01-19 at 6 59 11 pm

hobinjk:
Same goes for the onOffSwitch

benfrancis:
I'm a bit confused by this diff because not all of the code seems related to this issue, but I agree with @hobinjk, why do you need to add extra HTML to the view of every thing type? Why can't there be one thing-name element in the ThingsScreen, whose value is set in showThing(), with an icon URL set based on a CSS class?

hobinjk:
I agree that adding the element to ThingsScreen is a much nicer solution

mrstegeman:
@benfrancis It was Friday and my brain was overcomplicating things.

PR #527, opened 2018-01-23
Implement generic unknown thing UI.

### LINKED ISSUE #511: Unknown Thing UI
As a user I would like to be able to add a thing to the gateway, view and set its properties, and request actions, even if it doesn't conform to one of the standard thing types.

Currently the UI only lets you interact with things which conform to a standard Thing type. As people create custom things users should be able to add those things to the gateway and interact with them via a generic UI.

Mockups in progress...

mrstegeman:
Fixes #511

### DIFF
--- a/static/js/unknown-thing.js
+++ b/static/js/unknown-thing.js
@@ -1,0 +1,353 @@
+/**
+ * Unknown thing.
+ *
+ * UI element representing an unknown thing type.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+
+/* globals Thing, ThingDetailLayout */
+
+/**
+ * UnknownThing Constructor (extends Thing).
+ *
+ * @param Object description Thing description object.
+ */
+var UnknownThing = function(description, format) {
+  this.displayedProperties = {};
+  if (description.properties) {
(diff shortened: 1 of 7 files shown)


static/js/unknown-thing.js
```diff
@@ -0,0 +1,354 @@
...
+  }
+};
+
+/**
+ * Update the display for the provided property.
+ * @param {string} name - name of the property
+ * @param {*} value - value of the property
+ */
+UnknownThing.prototype.updateProperty = function(name, value) {
+  console.log(name);
```
hobinjk: nit: unnecessary console.log

static/js/unknown-thing.js
```diff
@@ -0,0 +1,354 @@
...
+/* globals Thing, ThingDetailLayout */
+
+/**
+ * UnknownThing Constructor (extends Thing).
+ *
+ * @param Object description Thing description object.
+ */
+var UnknownThing = function(description, format) {
+  this.displayedProperties = {};
+  if (description.properties) {
```
hobinjk: This should also include if (format === 'htmlDetail') or the updateProperty method later on might call update() on an unattached detail

mrstegeman: All of the `update()` methods already handle that case.

mrstegeman: Doing that also ends up breaking rendering of the HTML views, as the code has to look up if there are additional properties or not to justify a full expanded view.

hobinjk: Ah, sounds good. Thanks for the explanation!

PR #528, opened 2018-01-23
Add basic service-worker caching non-JSON requests

### LINKED ISSUE #447: Investigate using a ServiceWorker for caching and/or offline support

hobinjk:
Fix #447

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -97,6 +97,12 @@ var App = {
   }
 };
 
+if (navigator.serviceWorker) {
+  navigator.serviceWorker.register('/service-worker.js', {
+    scope: '/'
+  });
+}
+
 /**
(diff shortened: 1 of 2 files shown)


hobinjk:
If anything goes wrong with this, an update to service-worker.js just requires the tab or browser to be closed and opened to take effect (for some reason a soft refresh doesn't do anything)

hobinjk:
I simplified the logic of the serviceworker and pushed an updated version which should be working. The failure mode of a serviceworker is to silently let the browser do its thing which caused me a bit of pain debugging.

PR #529, opened 2018-01-24
Implement UI for remaining smart plug details.

mrstegeman:
In addition, add some constants that the Z-Wave and Zigbee adapters
were looking for.


benfrancis:
I actually intentionally left out the other properties from the UI for simplicity, particularly the level property which Dave noted is advertised by some devices even though it doesn't actually work.

static/js/level-detail.js
```diff
@@ -22,7 +22,7 @@ LevelDetail.prototype.attach = function() {
 }
 
 LevelDetail.prototype.view = function() {
-  let level = this.thing.properties.level;
+  const level = this.thing.properties.level;
```
benfrancis: How is this a constant?

mrstegeman: `level` is never redefined.

mrstegeman:
@benfrancis The properties are all conditionally shown. If they're not in the property description, they won't be shown. I can confirm that my Z-Wave plug did not dim properly. I think the proper solution would be to not report `level` if it's not functional.

benfrancis:
@mrstegeman OK, I think Dave was saying that the problem is the Z-Wave device reports that it has a multilevel switch but it actually just acts as a binary switch (or whatever the Z-Wave terms are), but there's no way for the adapter to know that it doesn't work.

mrstegeman:
@benfrancis Maybe we just disable the `level` property on Z-Wave devices until we get that figured out?

dhylands:
The ZW096 module is a relay device and as such can't dim. The ZW099 module is an actual dimmer and does dim.

dhylands:
This is a case where the adapter really needs to say, this device doesn't behave this way, so I'll adjust the properties. The ZW096 is a smartPlug with no level (the device says it supports the level command and it it does in fact respond to the level command, but they don't cause any changes to the physical device). The ZW096 is a smartPlug where the level should work properly.

What we really need is something the linux kernel calls 'quirks' (or special cases) and the adapter would just not add the level property for a ZW096

mrstegeman:
Z-Wave issues addressed here: https://github.com/mozilla-iot/zwave-adapter/pull/1

PR #534, opened 2018-01-25
Add support for time-based triggers; Refactor Rules UI

hobinjk:


hobinjk:
There's a weird locale thing where the time of day is based on a 24 hour clock but input type="time" for en-US expects you to input it as AM/PM

src/rules-engine/triggers/TimeTrigger.js
```diff
@@ -0,0 +1,70 @@
...
+    // TODO(hobinjk): support sunrise and sunset
+    let parts = this.time.split(':');
+    let hours = parseInt(parts[0], 10);
+    let minutes = parseInt(parts[1], 10);
+
+    let nextTime = new Date();
+    nextTime.setHours(hours);
+    nextTime.setMinutes(minutes);
+    nextTime.setSeconds(0);
+    nextTime.setMilliseconds(0);
```
dhylands: This is probably fine for this release, but we should figure out what to do if the wall time changes (like at daylight savings time changes and/or the user just changes the time).

I only looked at TimeTrigger.js since the other files appear to be mostly UI stuff.

hobinjk: Good point! As far as I can tell from playing around in the console with Date this will always trigger at a time when the wall clock is correct unless the time doesn't exist. In the spring forward case where there can be a nonexistent time it will silently advance an hour. In the fall back case it looks like it goes for the first instance of the time.

hobinjk:
@benfrancis Done! Thanks

They're #562 and #563

PR #536, opened 2018-01-26
Introducing communication with the offline intent parser server

andrenatal:
To test this, one needs to install and run: `https://github.com/mozilla-iot/intent-parser`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ addons:
 
 before_install:
   - ./tools/compile-openzwave.sh
+  - git clone https://github.com/mozilla-iot/intent-parser
+  - sudo pip install -e git+https://github.com/mycroftai/adapt#egg=adapt-parser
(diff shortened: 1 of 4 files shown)


andrenatal:
Tests here should be fixed too: https://github.com/mozilla-iot/gateway/blob/master/src/test/integration/commands-test.js

andrenatal:
Ok, all tests are fixed now. 

To test this, just install and start the [server](https://github.com/mozilla-iot/intent-parser) in a different terminal and start the gateway. All the rest is transparent 

There's a lot of room for improvement, but this patch fulfill what the old parser was doing plus the  automatic and dynamic training based on the current set of things. So this means that now the intent parser takes into account the current name of things, in opposite to the hard coded set we had before. Also the speed of having the intent determined offline is easily noticed. And, as it is all open source, we can adapt to our needs.

I suggest to first have this code reviewed, and then we decide about how to ship the intent server along the gateway image.

Issue #553, opened 2018-01-26
Things overlap buttons with a large number of things


benfrancis:
It would be nice if we handled a large number of things on the things screen a bit better.

Currently if you have a very large number of things they overlap with the menu button, add button and wordmark. It's not too bad because the buttons are on top and the wordmark is underneath so all the functionality still works. it just looks bad.

480x800:
![screen shot 2018-01-26 at 19 20 59](https://user-images.githubusercontent.com/552417/35456783-9a44082c-02ce-11e8-866e-dc59c1a18446.png)

320x480:
![screen shot 2018-01-26 at 19 24 53](https://user-images.githubusercontent.com/552417/35456792-a2884ad4-02ce-11e8-90b5-71246def506f.png)

I'm not sure of the best design to solve this. One option would be having a single row of things scrolling horizontally, but that's a bit wasteful on a large screen. Will play around with some designs.

hobinjk:
I personally think that the best way of solving this problem is to add a proper title bar which would include the hamburger button and the Mozilla IoT branding

Issue #537, opened 2018-01-26
Expanded thing UI won't fit on small screens


benfrancis:
The smart plug expanded thing UI won't fit on the screen, even at the common resolution of 1024x768.

I'd really like it to scale down to 800x480 (landscape) for demos with the Raspberry Pi 7" touchscreen, and ideally as low as 320x480 for smartphones (though this might not be practical for complex thing types).

320x480:
![screen shot 2018-01-24 at 16 17 43](https://user-images.githubusercontent.com/552417/35454702-0206a008-02c7-11e8-8aa7-0bbd29c1d42e.png)

320x480:
![screen shot 2018-01-24 at 16 19 08](https://user-images.githubusercontent.com/552417/35454712-0830682e-02c7-11e8-9ef5-5aa89f384930.png)

800x480:
![screen shot 2018-01-24 at 16 19 37](https://user-images.githubusercontent.com/552417/35454716-0bcd047e-02c7-11e8-907f-307bba1b311c.png)

800x480:
![screen shot 2018-01-26 at 15 17 56](https://user-images.githubusercontent.com/552417/35454719-0fb517a2-02c7-11e8-86bd-1b81909b01d9.png)

1024x768:
![screen shot 2018-01-26 at 15 26 18](https://user-images.githubusercontent.com/552417/35454723-1281cd0e-02c7-11e8-9074-b3809eec9f40.png)

PR #554, opened 2018-01-26
Position details on an ellipsis

hobinjk:
This adds pretty gigantic gutters into the things view since everything was overlapping (see #553).

Since those could be a bit iffy, I can take those changes out and keep only the ones relevant to #537 where I arrange the thing details on a conservatively-scaled ellipsis

### DIFF
--- a/static/js/thing-detail-layout.js
+++ b/static/js/thing-detail-layout.js
@@ -17,31 +17,41 @@ function ThingDetailLayout(elements) {
 }
 
 ThingDetailLayout.prototype.doLayout = function() {
-  let scale = 300;
+  let xScale = 300;
+  let yScale = 300;
+
   let angleStart = 0;
 
-  if (window.innerWidth < 680) {
-    scale = Math.max(130, window.innerWidth / 2 - 30);
-    angleStart = -0.7;
+  let circlePadding = 70;
+
+  let limitedXScale = window.innerWidth / 2 - circlePadding;
+  if (limitedXScale < xScale) {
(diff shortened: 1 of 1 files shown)


hobinjk:
(screenshots forthcoming)

hobinjk:
320x480 things view:
screen shot 2018-01-26 at 14 46 36

800x480 things view:
screen shot 2018-01-26 at 14 46 43

320x480 thing detail view, still needs some work:

screen shot 2018-01-26 at 14 49 10

800x480 detail view:

screen shot 2018-01-26 at 14 48 12

hobinjk:
Now with better support for the tiny:
screen shot 2018-01-26 at 16 16 53

benfrancis:
Very clever solution with the ellipsis.

An unfortunate side effect of the things screen fix is that I get a single column of things on my Pixel smartphone (and I imagine most smartphones?).
screenshot_20180129-170150

There's also a slightly weird effect where you can get a column of width 1 and then a column of width 2:
screen shot 2018-01-29 at 17 01 32

The thing detail view fix definitely seems worth landing as it's a big improvement.

I wonder whether there's another solution for the things view, could you maybe split that out into a separate pull request? I'm tempted to remove the Mozilla IoT wordmark because it keeps getting in the way. Another option would be to have margins at the top and bottom and get things to scroll in the box in between, but I imagine that will look pretty bad. A single row scrolling horizontally might actually be preferable to a single column.

hobinjk:
@benfrancis This is now only the detail positioning. That 1 and then 2 column behavior is really weird

PR #556, opened 2018-01-26
Allow user to skip tunnel setup.

### LINKED ISSUE #159: Fix domain setup skip button
Make skip button use plan HTTP by default rather than assume .pem files exist when they might not

mrstegeman:
Fixes #159

### DIFF
--- a/src/views/tunnel_setup.mustache
+++ b/src/views/tunnel_setup.mustache
@@ -23,8 +23,10 @@
       <div id="error-setup" class="error hidden">
         An error occured when setting up the subdomain
       </div><br>
-      <button id="skip-subdomain-button" type="button">Skip</button>
       <button id="create-domain-button" type="button" disabled>Create</button>
+      <div>
+        <a id="skip-subdomain-anchor" href="#">Skip</a>
+      </div>
     </form>
     <script src="/js/setup_subdomain.js"></script>
   </body>
--- a/src/views/tunnel_setup.mustache
+++ b/src/views/tunnel_setup.mustache
@@ -23,8 +23,7 @@
       <div id="error-setup" class="error hidden">
         An error occured when setting up the subdomain
       </div><br>
-      <!-- https://github.com/mozilla-iot/gateway/issues/159 -->
-      <!--<button id="skip-subdomain-button" type="button">Skip</button>-->
+      <button id="skip-subdomain-button" type="button">Skip</button>
       <button id="create-domain-button" type="button" disabled>Create</button>
     </form>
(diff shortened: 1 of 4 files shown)


hobinjk:
The checking of skip-tunnel to see if the page is HTTP or HTTPS should be unnecessary in the frontend since you can look at window.location. In general this looks like it might work but I haven't tested the full PR

mrstegeman:
@hobinjk you're right, fixed.

static/js/setup_subdomain.js
```diff
@@ -167,24 +166,24 @@ reclamationToken.addEventListener('keydown', submitOnEnter);
...
+      'Accept': 'application/json',
+    },
   }).then(function (response) {
     if (response.ok) {
-      displayMessage('Redirecting..', 'message');
+      displayMessage('Redirecting...', 'message');
       setTimeout(function() {
-        window.location.replace('http://' + window.location.hostname +
-                                ':8080');
+        window.location.replace(`http://${window.location.hostname}:8080`);
```
benfrancis: On the Pi, won't this be port 80 rather than port 8080?

src/views/tunnel_setup.mustache
```diff
@@ -23,8 +23,7 @@
       <div id="error-setup" class="error hidden">
         An error occured when setting up the subdomain
       </div><br>
-      <!-- https://github.com/mozilla-iot/gateway/issues/159 -->
-      <!--<button id="skip-subdomain-button" type="button">Skip</button>-->
+      <button id="skip-subdomain-button" type="button">Skip</button>
```
benfrancis: On the mockups this is a small link underneath the next button, rather than a button, to try to encourage people to choose a subdomain rather than skip it. This is because 1) It's more secure than plain HTTP 2) It will allow remote access through a firewall and 3) it will allow us to count the gateway in our analytics.

I suggest making it a small link underneath the button. An alternative might be to add "(recommended)" after "Choose a secure web address for your gateway", but I suspect the former will be more effective.

benfrancis:
@andrenatal Does this look OK to you? I can't remember how it was broken in the first place, just that it left the gateway in a broken state.

hobinjk:
Functionality LGTM testing on my Pi.

I think the Skip link could stand to be a smaller and greyed out a bit. Ideally it would require a confirmation step warning the user that they will be unable to access their Gateway remotely.

benfrancis:
Merging this to get it into the image for further testing. I'd like to see how it works as part of the Pi image.

PR #559, opened 2018-01-29
Setup the image to resize the filesystem on boot

dhylands:


image/add-gateway.sh
```diff
@@ -166,6 +179,43 @@ main() {
...
+# Default-Start: 3
+# Default-Stop:
+# Short-Description: Resize the root filesystem to fill partition
+# Description:
+### END INIT INFO
+. /lib/lsb/init-functions
+case "$1" in
+  start)
+    log_daemon_msg "Starting resize2fs_once"
+    ROOT_DEV=$(findmnt / -o source -n) &&
```
mrstegeman: Don't you need trailing line continuation characters `\` on the next several lines after `&&`?

PR #568, opened 2018-01-30
Cache-bust floorplan image on load in addition to on upload

### LINKED ISSUE #567: floor plan uploaded image lost if activate a thing before leaving page
I upload a floor plan, sprinkle my things on it, then click the check mark to exit edit mode.
Next I click on one of the Things, which opens a new page to turn it on/off, and then I click the back arrow. My floor plan image has disappeared. (But the things I have positioned on it are still in the same place.)

If instead, after I finish edit mode I go to a different page, such as the Things page, to turn something on/off, and then go back to the floor plan page, the image is still there.

hobinjk:
Fix #567

### DIFF
--- a/static/js/floorplan.js
+++ b/static/js/floorplan.js
@@ -42,6 +42,11 @@ var FloorplanScreen = {
     this.uploadButton.addEventListener('click', this.requestFile.bind(this));
     this.uploadForm.addEventListener('submit', this.blackHole);
     this.fileInput.addEventListener('change', this.upload.bind(this));
+
+    // Add a timestamp to the background image to force image reload
+    var timestamp = Date.now();
+    this.floorplan.setAttribute('style',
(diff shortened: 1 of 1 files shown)


hobinjk:
This is definitely a bit hacky. The thing I'm working around is that the original url is that the initial load puts`/uploads/floorplan.svg` into the cache. The upload then puts `/uploads/floorplan.svg?timestamp=13` into the cache. If you then load the floorplan again it goes back to having the original url of `/uploads/floorplan.svg` which is in the cache with the old data.

Alternative ways of fixing this are blacklisting the /uploads directory in the SW or adding support for recognizing a special ?cache-bust=true parameter to the service worker.

benfrancis:
If you can come up with a cleaner fix in time for the code freeze then that would be preferable, but if not then land this one (I probably won't be able to review now before code freeze today).

hobinjk:
Superseded by #598

PR #569, opened 2018-01-31
Add make-prep.sh script

dhylands:
This automates a bunch of the steps required to make a base image.

### DIFF
--- a/image/make-prep.sh
+++ b/image/make-prep.sh
@@ -12,6 +12,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.*
 
 SCRIPT_NAME=$(basename $0)
+SCRIPT_DIR=$(dirname $0)
 VERBOSE=0
 ENABLE_CONSOLE=1
 ENABLE_WIFI=0
@@ -19,6 +20,8 @@ WIFI_SSID=
 WIFI_PASSWORD=
 ENABLE_SSH=0
 HOSTNAME=gateway
+PRINT_SUMMARY=0
+DD_DEV=
 
 ###########################################################################
 #
@@ -35,8 +38,11 @@ where OPTION can be one of:
   --ssid          Specify the SSID for Wifi access
   --password      Specify the password for wifi access
   --hostname      Specify the hostname
+  --dd DEV        Issue a dd command to copy the image to an sdcard
+  --summary       Print summary of changed files
   -h, --help      Print this help
   -v, --verbose   Turn on some verbose reporting
+  -x              Does a 'set -x'
 END
 }
 
@@ -50,8 +56,7 @@ copy_prep_files() {
   HOME_PI="${ROOT_MOUNTPOINT}/home/pi"
 
   sudo rm -f "${HOME_PI}"/prepare-base*.sh
-  sudo wget --directory-prefix="${HOME_PI}" "https://raw.githubusercontent.com/mozilla-iot/gateway/master/image/prepare-base.sh"
-  sudo wget --directory-prefix="${HOME_PI}" "https://raw.githubusercontent.com/mozilla-iot/gateway/master/image/prepare-base-root.sh"
+  sudo cp "${SCRIPT_DIR}"/prepare-base*.sh "${HOME_PI}"
   sudo chmod +x "${HOME_PI}"/prepare-base*.sh
 }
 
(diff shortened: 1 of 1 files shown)


image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+#
+# Copies the prepare-base.sh and prepare-base-root.sh files to the image
+#
+copy_prep_files() {
+  echo "Copying prepare-base scripts"
+
+  HOME_PI="${ROOT_MOUNTPOINT}/home/pi"
+
+  sudo rm -f "${HOME_PI}"/prepare-base*.sh
+  sudo wget --directory-prefix="${HOME_PI}" "https://raw.githubusercontent.com/mozilla-iot/gateway/master/image/prepare-base.sh"
```
mrstegeman: Is it necessary to wget these scripts? Why not just copy them from the gateway repo (that this script lives in)?

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+}
+
+###########################################################################
+#
+# Enables the serial console
+#
+enable_serial_console() {
+
+  CONFIG=${BOOT_MOUNTPOINT}/config.txt
+  if grep -q "enable_uart=1" ${CONFIG} ; then
```
mrstegeman: This logic is backwards. This should be `if ! grep ...`.

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+}
+
+###########################################################################
+#
+# Enables Wifi
+#
+enable_wifi() {
+  echo "Enabling Wifi"
+  WPA_CONF=${ROOT_MOUNTPOINT}/wpa_supplicant/wpa_supplicant.conf
+  sudo sh -c "cat >> '${CONFIG}'" <<END
```
mrstegeman: This should be `WPA_CONF`, not `CONFIG`.

Also, what happens if this specific SSID was somehow already enabled? Would the 2 conflict and cause issues?

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+
+###########################################################################
+#
+# Update hostname
+#
+update_hostname() {
+  HOSTS_FILENAME="${ROOT_MOUNTPOINT}/etc/hosts"
+  HOSTNAME_FILENAME="${ROOT_MOUNTPOINT}/etc/hostname"
+  sudo sh -c "echo '${HOSTNAME}' > '${HOSTNAME_FILENAME}'"
+  sudo sed -i "${HOSTS_FILENAME}" -e "s/raspberrypi/${HOSTNAME}/"
```
mrstegeman: Might want to add the `g` flag to the regex, in case you have something like `127.0.0.1 raspberrypi raspberrypi.local`.

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+    enable_ssh
+  fi
+
+  update_hostname
+  copy_prep_files
+  sync
+
+  cleanup
+
+  # Because I always forget, print out the commands to upload to aws
```
mrstegeman: This comment doesn't seem to match what is being echoed.

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+  mkdir -p ${BOOT_MOUNTPOINT}
+  if sudo mount ${BOOT_DEV} ${BOOT_MOUNTPOINT}; then
+    echo "Boot partition mounted successfully"
+  else
+    echo "Failed to mount boot partition"
+    exit 1
+  fi
+  BOOT_MOUNTED=1
+
+  # Mount the root parition of the image
```
mrstegeman: partition

image/make-prep.sh
```diff
@@ -0,0 +1,325 @@
...
+      fi
+    done
+    echo "Loop mounts created successfully"
+  else
+    echo "Failed to create loop mounts"
+    exit 1
+  fi
+  LOOP_MOUNT_CREATED=1
+
+  # Mount the boot parition of the image
```
mrstegeman: partition

dhylands:
Thanks - I addressed all of the review comments and actually tested all of the features this time.

I also added a --dd option to have it write the final image out to an sdcard, and a --summary option which prints out the contents of the adjusted files.

PR #584, opened 2018-01-31
Things List Detail Link and on/off interaction

### LINKED ISSUE #570: Proposed refinements to Things View
Now that all the web thing types are all implemented, I'm finding the things view quite frustrating to use because:
1. There's inconsistency where some web thing types have an expanded view and others don't, some are directly tappable and some are hyperlinks.
2. Turning smart plugs on and off now requires two taps instead of one, which is really inconvenient.

The current performance over the SSL tunnel makes this particularly frustrating to use.

I've being playing around with some alternative designs and this is the best solution I've come up with so far:

things-expandable-off

things-expandable-on

In this design:
- All things which have an on/off property can be toggled on/off with a single tap
- Things which have an expanded view have an expand icon on them you can click to get to the expanded view

Other options I considered were:

- Long press/right click to get to expanded view (not very discoverable)
- All things are hyperlinks to an expanded view (introduces an extra step for all on/off switches)
- All things are hyperlinks but things with on/off have a mini on/off button attached to them (looked strange, particularly for on/off switches)

hobinjk:
Close #570

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -28,11 +28,11 @@
   z-index: 10;
   float: right;
   background-color: #89b6d6;
-  background-size: 3.2rem;
+  background-size: 2.4rem;
   background-position: center;
   background-repeat: no-repeat;
-  background-image: url('../images/thing-details-link.svg');
-  border: 2px solid white;
+  background-image: url('../images/things-icon.png');
+  border: 1px solid white;
   border-radius: 2rem;
   width: 4rem;
   height: 4rem;
@@ -78,16 +78,6 @@
   height: 3.6rem;
   overflow: hidden;
   margin: 0.5rem 0;
-}
-
-/* Hint at interactive things */
-.color-light,
-.dimmable-light,
-.on-off-light > .thing-icon,
-.smart-plug > .thing-icon,
-.on-off-switch > .thing-icon,
-.level-bar-container {
-  cursor: pointer;
 }
 
 /* On/Off Switch */
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -80,6 +80,16 @@
   margin: 0.5rem 0;
 }
 
+/* Hint at interactive things */
+.color-light,
+.dimmable-light,
+.on-off-light > .thing-icon,
+.smart-plug > .thing-icon,
+.on-off-switch > .thing-icon,
+.level-bar-container {
+  cursor: pointer;
+}
+
 /* On/Off Switch */
 
 .on-off-switch .thing-icon {
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -31,7 +31,7 @@
   background-size: 2.4rem;
   background-position: center;
   background-repeat: no-repeat;
-  background-image: url('../images/things-icon.png');
+  background-image: url('../images/thing.svg');
   border: 1px solid white;
   border-radius: 2rem;
   width: 4rem;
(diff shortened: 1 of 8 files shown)


static/css/thing.css
```diff
@@ -21,6 +21,23 @@
...
 }
 
+.thing-details-link {
+  box-sizing: border-box;
+  display: block;
+  position: relative;
+  z-index: 10;
+  float: right;
+  background-color: #89b6d6;
+  background-size: 3.2rem;
```
benfrancis: Can you make this 2.4rem instead of 3.2rem? The bubble is smaller than the mockups (which it does actually need to be if it's going to line up with the right of the icon), so I think the icon needs to shrink as well. Hopefully this is a big enough touch target.

static/css/thing.css
```diff
@@ -63,6 +80,16 @@
...
 }
 
+/* Hint at interactive things */
+.color-light,
+.dimmable-light,
+.on-off-light > .thing-icon,
+.smart-plug > .thing-icon,
+.on-off-switch > .thing-icon,
+.level-bar-container {
+  cursor: pointer;
```
benfrancis: I actually think only the "expand thing" bubble should have the pointer cursor because it's a hyperlink that navigates somewhere, whereas tapping the thing icon behaves more like a button. The pointer is usually for hyperlinks and I wouldn't expect a hyperlink to toggle some state and stay on the same page.

static/css/thing.css
```diff
@@ -277,12 +308,12 @@
   background-image: url('../images/smart-plug-off.svg');
 }
 
-.on .smart-plug .thing-icon, .smart-plug-container.on .thing-icon {
+.smart-plug.on .thing-icon, .smart-plug-container.on .thing-icon {
```
benfrancis: cool, this is a more efficient selector. There was some reason this wouldn't work before.

static/images/thing-details-link.svg
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
benfrancis: You probably could have just used the PNG from the "Things" item in the main menu. Let me export an SVG that matches that a bit better...

static/css/thing.css
```diff
@@ -21,6 +21,23 @@
...
+  display: block;
+  position: relative;
+  z-index: 10;
+  float: right;
+  background-color: #89b6d6;
+  background-size: 3.2rem;
+  background-position: center;
+  background-repeat: no-repeat;
+  background-image: url('../images/thing-details-link.svg');
+  border: 2px solid white;
```
benfrancis: I think 1px might actually look better for such a small circle, even though it's inconsistent with the thing icon.

static/css/thing.css
hobinjk: Oh right, that is how buttons work :+1:

static/images/thing-details-link.svg
hobinjk: I've switched to the PNG, thanks for pointing out that it exists!

hobinjk:
I've made the suggested changes, thanks for the review!

benfrancis:
Here's an SVG icon that might scale down better https://drive.google.com/file/d/1xs5eGJ6yYUuepw0VoNb8L_DGLWgb7n_l/view?usp=sharing

Issue #571, opened 2018-01-31
Content Negotiation does not set Vary: Accept


hobinjk:
Found when testing in chrome: the API fetch request for `/things` would hit the disk cache and return the HTML response since the Gateway didn't tell it that the Accept header mattered.

PR #586, opened 2018-01-31
Use xlink:href instead of href

### LINKED ISSUE #579: Floorplan hyperlinks don't work in Safari
- Floorplan background image doesn't appear
- Hyperlinks on thing icons don't work

hobinjk:
Fix #579


hobinjk:
Note that things on Safari are rather broken without #571 also fixed

hobinjk:
*shakes fist

PR #591, opened 2018-02-01
Add sensor icons - closes #578

### LINKED ISSUE #578: Binary Sensor & Multilevel Sensor missing SVG icons for floorplan
In Firefox they're just blank, but you get a default image in Chrome & Safari
screen shot 2018-01-31 at 19 03 36

benfrancis:
An SVG icon for sensors (not very exciting), and some tweaks to the sensor icons to better match the UX spec.


hobinjk:
The Multi-level sensor is using the binary sensor's svg on the floorplan. Is this the intended icon?

benfrancis:
Yeah they're the same. I couldn't think of a better neutral state.

I actually expected the multi-level sensor to have a bar chart style like the multi-level switch but it's just a percentage in a circle. That's probably fine, I'm not actually sure we're going to keep that Web Thing type because I'm not sure if it's actually useful with just a percentage as a value.

hobinjk:
Sounds good to me

PR #595, opened 2018-02-01
Flex thing title based on back and speech buttons

### LINKED ISSUE #594: Thing name overlaps back button in thing detail view
I think we need to either wrap or probably truncate the text, whichever looks best.

screen shot 2018-02-01 at 21 04 18

hobinjk:
Fix #594


hobinjk:
Yeah, I had to make the layout very divvy to center the title based on the presence of the speech button. I've reduced the padding to 0.5rem on the icon and it looks better. Thanks for the review!

benfrancis:
Divvy. Good word.

PR #596, opened 2018-02-02
Hardcode SVG text

hobinjk:
This is necessary to work on computers without Open Sans installed since svg loaded as an image (or background-image) is prohibited from accessing the fonts of the embedding page.


hobinjk:
Yes, I just took the original and converted the text to a path. It's the kind of thing that's a bit of a pain but it didn't take too long

PR #598, opened 2018-02-02
Manually reload cache of floorplan image

### LINKED ISSUE #567: floor plan uploaded image lost if activate a thing before leaving page

hobinjk:
Close #567

### DIFF
--- a/static/js/floorplan.js
+++ b/static/js/floorplan.js
@@ -185,20 +185,29 @@ var FloorplanScreen = {
     var formData = new FormData();
     formData.append('file', file);
     this.uploadButton.classList.add('loading');
+    let headers = {
+      'Authorization': `Bearer ${window.API.jwt}`,
+    };
+
     fetch('/uploads', {
       method: 'POST',
       body: formData,
-      headers: {
-        'Authorization': `Bearer ${window.API.jwt}`,
-      }
+      headers: headers
     }).then((response) => {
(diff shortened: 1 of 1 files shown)


static/js/floorplan.js
```diff
@@ -185,20 +185,29 @@ var FloorplanScreen = {
...
+        fetch('/uploads/floorplan.svg', {
+          method: 'GET',
+          headers: headers,
+          // Make sure we update the cache with the new floorplan
+          cache: 'reload'
+        }).then(() => {
+          // Add a timestamp to the background image to force image reload
+          var timestamp = Date.now();
+          this.floorplan.setAttribute('style',
+            `background-image: url("/uploads/floorplan.svg?t=${timestamp}")`);
```
mrstegeman: Is the timestamp still necessary since you're clearing the cache?

hobinjk: Yes, the image still needs to be reloaded on the current page and just setting the style attribute without the additional query parameter does nothing

PR #603, opened 2018-02-05
Add intent-parser to image

hobinjk:
Run intent-parser in background doing nothing (for now) to make sure
that we can eventually OTA a switch over to it.


hobinjk:
So I made this PR but as I was writing out the commit message I realized that a better way to ship the intent parser would be as an always-enabled packaged python add-on to the gateway like the TP-link adapter

PR #627, opened 2018-02-07
[minor] promote use of LTS

Swaagie:
First off, thanks for this initiative. Wonderful!

Coming from a Node.js background I would advice to always be on latest LTS branch. Some additional testing is required before this can be merged. I'm curious why `node@7` usage was documented or if it simply was a matter of not being updated yet?

Using latest LTS makes sense from a security, stability and support perspective.

Let me know what you think, would love to get more involved in the project. Still working hard to get a complete IoT setup at home.


hobinjk:
@Swaagie Thanks for the contribution! Moving to the LTS is definitely something that we want to do. There are some other references to 7.10.1 in `cross-compile.sh`, `install.sh`, `image/prepare-base.sh`, `image/prepare-base-root.sh`, and `image/build-gateway.sh`.

I'm okay with merging this as the first step of a longer migration process if you don't want to have to hunt down all the minor things that have to change.

Swaagie:
Thanks for the references I'll scrutinize further to change those files as well.

I can see why the `package-lock.json` shouldn't be part of this PR. `npm@5` by default will create a package-lock.json though and It is quite similar to `yarn.lock`, or would you prefer to have a separate PR bringing that in (if at all)?

Will update shortly.

hobinjk:
At some point in the future I think we should switch to package-lock.json over yarn.lock. That's a larger conversation that would require some discussion and the involvement of the entire team.

I do want to avoid having both present so that there's no doubt what the "official" lock file is.

Swaagie:
Word, we've had the same discussion on other projects. Will remove it

PR #628, opened 2018-02-07
Log to console on failure to parse add-ons list.

### LINKED ISSUE #619: No error or warning messages if list.json isn't valid
If there is an error in the addon-list list.json file then there are no error or warning messages printed. The UI winds up behaving weirdly (in my case a blank screen)

mrstegeman:
There are no issues with the UI, there's just nothing to display. The back button still works fine.

Fixes #619

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -378,7 +378,7 @@ var SettingsScreen = {
           addon.installed = this.installedAddons.has(addon.name);
           new DiscoveredAddon(addon);
         }
-      });
+      }).catch((e) => console.error(`Failed to parse add-ons list: ${e}`));
     });
   },
 


static/js/settings.js
```diff
@@ -378,7 +378,7 @@ var SettingsScreen = {
           addon.installed = this.installedAddons.has(addon.name);
           new DiscoveredAddon(addon);
         }
-      });
+      }).catch((e) => console.error(`Failed to parse add-ons list: ${e}`));
```
dhylands: Hmm. This will log to the web console right? I was thinking that we should have something log to the node console. Or is this a shared file?

Since the URL can't change at runtime, maybe we can do a validity check on the server side at startup time?

mrstegeman: Yes, it will log to the web console. As far as I know, we don't currently have an easy was to send errors back to node for logging. Maybe we should think about doing that.

I don't think checking at startup will do much good. The addons list can change at any time, and the gateway can be running for an indefinite period of time, so a one-off check doesn't really help.

dhylands: ok - then this is as good as we can do for now.

Maybe an even better solution is to do a travis job that runs jsonlint on each PR

PR #629, opened 2018-02-07
Fix UI flash before login.

### LINKED ISSUE #208: UI flash before login page appears
When a user isn't logged in and they load the web app they briefly see an empty devices view before the login page appears.

This is due to the JavaScript redirect (previously we had a server side redirect).

@lightsofapollo Is there a cleaner way of doing this?

mrstegeman:
We still see the lighter blue color, but none of the elements.

Fixes #208

### DIFF
--- a/src/views/index.html
+++ b/src/views/index.html
@@ -70,7 +70,7 @@
     <script type="text/javascript" src="/js/rule-screen.js"></script>
 
   </head>
-  <body>
+  <body class="hidden">
     <img id="wordmark" alt="Mozilla IoT" src="/images/moziot-wordmark.png" />
 
     <!-- Things View -->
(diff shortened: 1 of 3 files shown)


mrstegeman:
What if we combine the two approaches? We have no guarantee of when the code will run, and the UI layout will have probably started (and potentially finished) either way.

hobinjk:
Oh, good point, I didn't remember that isLoggedIn gets called before verifyJWT.

My main concern is that by defaulting the body to hidden we're requiring the API request in verifyJWT to complete before any UI is shown to the user.

That being said, while testing this out on reasonably average 4G I didn't notice that bad of a downgrade so this LGTM.

PR #642, opened 2018-02-08
Add adapterId value to setProperty to match Adapter IPC spec

hobinjk:
https://github.com/mozilla-iot/wiki/wiki/Adapter-IPC#setproperty

### DIFF
--- a/src/addons/plugin/adapter-proxy.js
+++ b/src/addons/plugin/adapter-proxy.js
@@ -57,6 +57,7 @@ class AdapterProxy extends Adapter {
   }
 
   sendMsg(methodType, data) {
+    data.adapterId = this.id;
     return this.plugin.sendMsg(methodType, data);
   }
 


src/addons/plugin/property-proxy.js
```diff
@@ -67,6 +67,7 @@ class PropertyProxy extends Property {
       this.device.adapter.sendMsg(
         Constants.SET_PROPERTY, {
           deviceId: this.device.id,
+          adapterId: this.device.adapter.id,
```
dhylands: I think that this should really go in the AdapterProxy.sendMsg method, similar to how the Plugin.sendMsg injects the pluginId: https://github.com/mozilla-iot/gateway/blob/a5c90fbbe10594ea1cb6fbdda62c9afc4ac6d762/src/addons/plugin/plugin.js#L216

hobinjk:
@dhylands Updated accordingly, thanks for the feedback!

src/addons/plugin/adapter-proxy.js
```diff
@@ -57,6 +57,7 @@ class AdapterProxy extends Adapter {
   }
 
   sendMsg(methodType, data) {
+    data.adapterId = this.id;
     return this.plugin.sendMsg(methodType, data);
   }
```
dhylands: LGTM

PR #644, opened 2018-02-08
Allow user to update add-ons.

### LINKED ISSUE #478: Allow user to update add-ons
As a user, I want to be able to update my add-ons to pull in new features, bug fixes, etc.

### LINKED ISSUE #643: Add-on list should show the version number of each installed add-on
Currently, you see the list of installed addons, and a description, but no version.

mrstegeman:
Fixes #478
Fixes #643

### DIFF
--- a/src/controllers/addons_controller.js
+++ b/src/controllers/addons_controller.js
@@ -10,6 +10,54 @@ const Settings = require('../models/settings');
 
 const AddonsController = PromiseRouter();
 
+/**
+ * Install an add-on.
+ *
+ * @param {String} name The package name
+ * @param {String} url The package URL
+ * @returns A Promise that resolves when the add-on is installed.
+ */
+async function installAddon(name, url) {
+  const tempPath = fs.mkdtempSync(`${os.tmpdir()}${path.sep}`);
+  const destPath = path.join(tempPath, `${name}.tar.gz`);
+
+  console.log(`Fetching add-on ${url} as ${destPath}`);
+
+  try {
+    const res = await fetch(url);
+    const dest = fs.createWriteStream(destPath);
+    await promisePipe(res.body, dest);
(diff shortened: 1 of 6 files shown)


dhylands:
And being really nit-picky here, the text for Updating and Updated seems to be crammed right up against the button. It could use a tiny bit of space between the text and the button.

dhylands:
The spacing looks much nicer, but I'm still seeing the old version number and not the version number of the newly installed  addon.
wrong-version

mrstegeman:
Did you hard refresh? I tested multiple times. The spacing is still wrong in that picture, so I'm guessing things were still cached.

dhylands:
Something is wonky with my files. My repo says its at the same hash, but the contents of the file doesn't match what's here.

dhylands:
It looks fine now. I think I had some combination of Dropbox mangling things and caching issues.

dhylands:
Incidentally, installing an update wipes out the git repository if you happen to have your git repository stored in the build tree (which I typically do for development).

mrstegeman:
Correct, it does a full uninstall + install. This is probably preferred, that way each install/update starts with clean source.

dhylands:
Yeah - I think its mostly a developer be-aware. The only thing I could think of improving is to have the update fail if a .git directory is present. Another option would be to copy the .git directory from the currently installed directory to the newly installed one (i.e. using something like the ncp module)

benfrancis:
We need a UX spec for this.

mrstegeman:
@benfrancis I added the ui-needed label and some screenshots to #478 this morning.

benfrancis:
@mrstegeman Sorry, I didn't realise this was a pull request for the same issue, was just trying to mark everything that needs UI design work :)

PR #648, opened 2018-02-09
Ignore cors and other related requests

hobinjk:


hobinjk:
This would cause an issue if using a custom non-HTTPS value for the addon list because it's a mixed content CORS fetch request. The serviceworker would attempt to cache it then be very confused

hobinjk:
This fixes #683

hobinjk:
Incorporated in #684

PR #662, opened 2018-02-12
Make tsc ignore addons.

mrstegeman:


hobinjk:
@mrstegeman You can ignore all of `build` if you want to since the only files that go there are ones that tsc is done processing.

PR #663, opened 2018-02-12
Relocate persistent data files into a user profile.

mrstegeman:
These need to be moved out of the gateway tree for data persistence
when using Docker. This will also allow users to easily switch
profiles, if desired.

Z-Wave logs and such will be handled at the add-on level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,13 +118,15 @@ $ yarn
 
  Add SSL certificate:
 
- If you don't plan on using Mozilla's provided tunneling service to set up a `*.mozilla-iot.org` domain, you can use your own SSL certificate. The HTTPS server looks for `privatekey.pem` and `certificate.pem`. You can use a real certificate or generate a self-signed one by following the steps below.
+ If you don't plan on using Mozilla's provided tunneling service to set up a `*.mozilla-iot.org` domain, you can use your own SSL certificate. The HTTPS server looks for `privatekey.pem` and `certificate.pem` in the `ssl` sub-directory of the `userProfile` directory specified in your config. You can use a real certificate or generate a self-signed one by following the steps below.
 
  ```
- $ mkdir -p ssl
- $ openssl genrsa -out ssl/privatekey.pem 2048
- $ openssl req -new -sha256 -key ssl/privatekey.pem -out ssl/csr.pem
- $ openssl x509 -req -in ssl/csr.pem -signkey ssl/privatekey.pem -out ssl/certificate.pem
+ $ MOZIOT_HOME="${MOZIOT_HOME:=${HOME}/.mozilla-iot}"
+ $ SSL_DIR="${MOZIOT_HOME}/ssl"
+ $ [ ! -d "${SSL_DIR}" ] && mkdir -p "${SSL_DIR}"
+ $ openssl genrsa -out "${SSL_DIR}/privatekey.pem" 2048
+ $ openssl req -new -sha256 -key "${SSL_DIR}/privatekey.pem" -out "${SSL_DIR}/csr.pem"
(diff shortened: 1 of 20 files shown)


tools/make-self-signed-cert.sh
```diff
@@ -2,7 +2,8 @@
 
 set -x
 
-mkdir -p ssl
-openssl genrsa -out ssl/privatekey.pem 2048
-openssl req -new -sha256 -key ssl/privatekey.pem -out ssl/csr.pem -subj '/CN=www.toizom.com/O=MozillaIoT Gateway/C=US'
-openssl x509 -req -in ssl/csr.pem -signkey ssl/privatekey.pem -out ssl/certificate.pem
+export SSL_DIR="${HOME}/mozilla-iot/data/ssl"
```
dhylands: Does this need to be exported?

dhylands: And there really needs to be a way to override this. My mozilla-iot directory isn't in my HOME directory, it's in my Dropbox folder.

src/test/run-tests.sh
```diff
@@ -2,7 +2,7 @@
 
 SCRIPTDIR="$(dirname ""$0"")"
 
-if [ ! -f "ssl/certificate.pem" ]; then
+if [ ! -f "$HOME/mozilla-iot/data/ssl/certificate.pem" ]; then
```
dhylands: We should probably have an environment variable like MOZIOT_HOME default to ${HOME}/mozilla-iot but allow it to be override. All of the bash scripts should probably use something like:
```
: ${MOZIOT_HOME:="${HOME}/mozilla-iot"}
```
or the slightly more verbose, but perhaps more easily understood
```
MOZIOT_HOME="${MOZIOT_HOME:=${HOME}/mozilla-iot}"
```
and then use ${MOZIOT_HOME} everywhere

src/app.js
```diff
@@ -34,6 +35,17 @@ require('./log-timestamps');
...
+// Relocate the database before opening it, if necessary.
+const dbPath = config.get('database.filename');
+const dbDir = path.dirname(dbPath);
+if (!fs.existsSync(dbPath)) {
+  mkdirp.sync(dbDir);
+}
+
+if (fs.existsSync('db.sqlite3')) {
+  fs.renameSync('db.sqlite3', dbPath);
+}
```
dhylands: I think that this deserves a comment to explain why its here (i.e. migration)

benfrancis:
Can we have a quick discussion about this in our meeting today after planning the 0.4 release and before landing? It seems like a modestly large architectural change that warrants a group discussion.

mrstegeman:
@benfrancis Sure thing.

mrstegeman:
@benfrancis are we ready to move forward with this, or are there still outstanding issues?

config/default.js
```diff
@@ -55,7 +56,7 @@ module.exports = {
     domain: 'mozilla-iot.org',
     pagekite_cmd: './pagekite.py',
     port: 443,
-    certemail: 'certificate@mozilla-iot.org'
+    certemail: 'certificate@mozilla-iot.org',
   },
-  bcryptRounds: 2
+  bcryptRounds: 2,
```
benfrancis: What's with all the extra commas?

README.md
```diff
@@ -181,6 +183,7 @@ $ jest src/test/{test-name}.js
   * **`router.js`** - Routes app URLs to controllers
   * **`ssltunnel.js`** - Utilities to determine state of tunnel and manage the PageKite process
   * **`tunnel_setup.js`** - Express middleware to determine if the tunnel is set up
+  * **`user-profile.js`** - Manages persistent user data
```
benfrancis: I'm not sure it's really a user profile if it's shared by all users, but to be honest I can't think of a better name.

config/default.js
mrstegeman: It's not JSON, so the trailing commas make maintenance easier.

PR #687, opened 2018-02-18
execute logic in check-user.js in a domcontentloaded event handler …

attl8d:
I encountered this issue on chrome, where the document.body would be null when the code executed, rendering all the content of the app hidden.
I did not encounter this in firefox.

### DIFF
--- a/static/js/check-user.js
+++ b/static/js/check-user.js
@@ -37,5 +37,5 @@
     });
   }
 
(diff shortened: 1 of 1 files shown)


hobinjk:
Thanks for the PR! I fixed this in #684 but forgot to mention that in the PR title. Sorry for the confusion!

PR #699, opened 2018-02-21
Enable HTTP Strict Transport Security (HSTS).

mrstegeman:


benfrancis:
@mrstegeman Just curious, what happens when you're using the gateway using a local IP address or .local domain? Do things still work as before?

And I assume this only applies to the gateway's own origin? Would it prevent you doing a fetch to a native web things via its local IP address if it is only using plain HTTP?

hobinjk:
@benfrancis In the ip/.local case I've been getting "Strict-Transport-Security: The connection to the site is untrustworthy, so the specified header was ignored." which seems alright

PR #715, opened 2018-02-27
Implement EventTrigger and ActionEffect

hobinjk:


hobinjk:
My plan for the UI is to add events and actions into the existing property drop down so it will end up looking like
- On
- Off
- Action "toggle"
as an effect and
- On
- Off
- Event "motion"
as a trigger

hobinjk:
screen shot 2018-02-28 at 1 35 56 pm

screen shot 2018-02-28 at 13 48 46

src/rules-engine/Property.js
```diff
@@ -40,8 +39,8 @@ class Property extends EventEmitter {
     this.name = parts[parts.length - 1];
 
     this.onMessage = this.onMessage.bind(this);
-    this.ws = null;
-    this.id = Math.floor(Math.random() * 100000);
+    let thingHref = this.href.split('/properties')[0];
```
benfrancis: Is there some way we can avoid hard coding this string? We might not be able to assume this URL structure for native web things.

src/rules-engine/ThingConnection.js
```diff
@@ -0,0 +1,60 @@
...
+    this.href = href;
+    this.messageHandler = messageHandler;
+    this.onMessage = this.onMessage.bind(this);
+    this.ws = null;
+  }
+
+  async start() {
+    const jwt = await Settings.get('RulesEngine.jwt');
+    const gateway = await Settings.get('RulesEngine.gateway');
+    const wsHref = gateway.replace(/^http/, 'ws') + this.href + '?jwt=' + jwt;
```
benfrancis: This is probably OK for now, but note that our current spec defines a links member of the Thing Description which includes the WebSocket endpoint. We may not always be able to assume that the WebSocket URL is the same as the Thing URL. But I realise we don't currently implement the links member and it may be subject to change in the Interest Group.

src/rules-engine/effects/ActionEffect.js
```diff
@@ -0,0 +1,81 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+const assert = require('assert');
+const fetch = require('node-fetch');
+const Effect = require('./Effect');
+const Settings = require('../../models/settings');
+
+/**
+ * An Effect which creates an action
```
benfrancis: Heh, this would have been confusing if Effect was also still called Action! Good job we renamed it.

src/rules-engine/ThingConnection.js
```diff
@@ -0,0 +1,60 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+const e2p = require('event-to-promise');
+const Settings = require('../models/settings');
+const WebSocket = require('ws');
+
+class ThingConnection {
```
benfrancis: Please add a short description of what a ThingConnection does for documentation.

src/rules-engine/Property.js
hobinjk: Yes, I'll file a follow-up issue since it will also involve migrating any existing rules.

src/rules-engine/effects/ActionEffect.js
hobinjk: Yep, our acting to rename had a very positive effect

src/rules-engine/ThingConnection.js
hobinjk: I can definitely change this once we implement the links member

PR #720, opened 2018-02-28
Move add-ons into user profile directory.

### LINKED ISSUE #716: Move add-ons out of source/build tree and into user profile
Installed add-ons need to be moved into the user profile such that they are persistent with Docker containers, etc.

mrstegeman:
This requires a bit of hackery and symlinking to make existing
calls to `require()` work properly. This can be resolved in the
future by creating a proper gateway-addon package for Node.

Fixes #716

### DIFF
--- a/src/test/run-tests.sh
+++ b/src/test/run-tests.sh
@@ -3,7 +3,9 @@
 SCRIPTDIR="$(dirname ""$0"")"
 
 MOZIOT_HOME="${MOZIOT_HOME:=${HOME}/.mozilla-iot}"
+MOZIOT_HOME="${MOZIOT_HOME}/test"
 if [ ! -f "${MOZIOT_HOME}/ssl/certificate.pem" ]; then
+  export MOZIOT_HOME
   ${SCRIPTDIR}/../../tools/make-self-signed-cert.sh
 fi
 
--- a/src/user-profile.js
+++ b/src/user-profile.js
@@ -25,8 +25,15 @@ const Profile = {
     this.baseDir = config.get('profileDir');
     this.configDir = path.join(this.baseDir, 'config');
     this.sslDir = path.join(this.baseDir, 'ssl');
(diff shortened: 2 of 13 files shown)


src/test/run-tests.sh
```diff
@@ -3,7 +3,9 @@
 SCRIPTDIR="$(dirname ""$0"")"
 
 MOZIOT_HOME="${MOZIOT_HOME:=${HOME}/.mozilla-iot}"
+MOZIOT_HOME="${MOZIOT_HOME}/test"
 if [ ! -f "${MOZIOT_HOME}/ssl/certificate.pem" ]; then
```
dhylands: Having the export inside the if feels wrong.

src/user-profile.js
```diff
@@ -140,6 +153,30 @@ const Profile = {
...
+            fs.renameSync(oldFname, newFname);
+          } else if (fname.endsWith('.js') && lstat.isFile()) {
+            // Symlink dependencies.
+            if (!fs.lstatSync(newFname).isSymbolicLink()) {
+              fs.symlinkSync(oldFname, newFname);
+            }
+          }
+        }
+      }
+    }
```
dhylands: Rather than create symlinks, would it make sense to move the entire tree using something like there move command from the fs-extra package.

src/test/run-tests.sh
mrstegeman: I agree, but it also seemed wrong to export unnecessarily.

src/user-profile.js
mrstegeman: I wanted to keep the js files in the build tree such that when they're updated in the source tree, we don't have to md5 or anything to figure out if we need to copy again. It also didn't feel right putting source that is imported by both the gateway and add-ons in the user profile.

Ideally, this will all go away in the future after we make a node package for the add-ons to import.

PR #729, opened 2018-03-02
Auto-update add-ons on startup and daily.

mrstegeman:


hobinjk:
Awesome! This will be great to have

PR #736, opened 2018-03-05
Sanitize HTML and URLs.

### LINKED ISSUE #733: Security issue: unsanitized html
This is a pervasive security issue but one which doesn't matter that much for now because we don't ever add Things to the gateway without user input. That being said, this definitely needs to eventually be fixed. One example attack is to create a thing description containing a property with `"name": "\"><script>alert('pwned');</script>"`.

When Add Thing by URL exists the ease of performing this attack increases greatly, requiring the attacker to only be on the same network instead of coercing the victim into installing an add-on containing exploit code. It's still not really that big of a concern even then but bears looking into before it becomes an actual threat

mrstegeman:
Fixes #733

### DIFF
--- a/static/js/speech.js
+++ b/static/js/speech.js
@@ -53,7 +53,7 @@ var Speech = {
             x.className = 'show';
             setTimeout(function(){ x.className =
                 x.className.replace('show', ''); }, 3000);
-            new Audio(`/audio/${audio}.mp3`).play();
+            new Audio(`/audio/${encodeURIComponent(audio)}.mp3`).play();
           }
           document.getElementById('stm-levels').classList.add('hidden');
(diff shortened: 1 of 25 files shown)


static/js/speech.js
```diff
@@ -72,7 +72,7 @@ var Speech = {
           var opts = {
             method: 'POST',
             cache: 'default',
-            body: `{"text":"${results[0].text}"}`,
+            body: `{"text":"${results[0].text.replace(/"/g, '\\"')}"}`,
```
hobinjk: This would be a bit more general as
```
body: JSON.stringify({text: results[0].text})
```

mrstegeman: Yep. Fixed.

Issue #645, opened 2018-02-08
View Logs


dhylands:
We should provide an easy way for the end-user to download log files from the gateway. The run-app.log and OZW_Log.txt files are 2 that immediately come to mind.

benfrancis:
As a user I would like to be able to share log files from my gateway so that someone can help debug my problems.

mrstegeman:
Do we want this accessible from the UI or just a URL that we can tell people to browse to?

benfrancis:
I don't really think this should be exposed through the UI. It's more of a developer feature than an end user feature. I suppose it could be part of a developer mode you enable like on Android. Having a special URL like "/logs" might be OK but we should be careful about security.

Swaagie:
This would be a nice to have feature indeed. In addition, having a route on the Web Thing API could be nice to allow external analyzers to fetch and parse logs?

mrstegeman:
This is possible now, we just need to decide if we want to put this in the UI or not, and where. `/logs/*` is authenticated, so the JWT must be included.

benfrancis:
This should go in a new developer settings section, coming soon...

benfrancis:
UX specification: https://docs.google.com/presentation/d/1rMeg6hep8JQflqwbiNIbhSoK9StCeVZ2npphmQHKda4/edit#slide=id.g3c095fbf0e_6_1

UI mockup (just a button in developer settings):
![developer-settings](https://user-images.githubusercontent.com/552417/40841583-e492de12-65a2-11e8-8ab3-7bb2fd6a3b24.png)

PR #737, opened 2018-03-06
Add a route to allow user to download logs.

mrstegeman:
References #645

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   },
   "homepage": "https://iot.mozilla.org",
   "dependencies": {
+    "archiver": "^2.1.1",
     "asn1.js": "^5.0.0",
     "bcryptjs": "^2.4.3",
     "body-parser": "^1.17.1",
@@ -75,6 +76,7 @@
     "eslint": "^4.7.0",
     "eslint-plugin-html": "^4.0.0",
     "jest": "^21.2.1",
+    "jszip": "^3.1.5",
     "nock": "^9.0.18",
     "node-rest-client": "^3.1.0",
     "nyc": "^11.0.2",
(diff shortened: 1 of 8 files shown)


benfrancis:
A couple of thoughts:
* What should be the relationship between /debug and /logs? We already have a debug controller which is useful for development but I think has to be enabled with a command line argument. Is this more of a user-facing thing?
* Is creating a .zip file the best approach? I wonder if might be nicer to just view the logs in my browser. If there are multiple log files the /logs could be an index with links to /logs/foo and /logs/bar.

mrstegeman:
1. `/debug` is only allowed if node's `debug` option is set. `/logs` should probably always be available.
2. That's probably doable, but will be a larger amount of effort for what is really a developer feature. Also, you told me you didn't think it should be exposed through the UI :-) (https://github.com/mozilla-iot/gateway/issues/645#issuecomment-367669503). There are indeed multiple log files. Right now, there's `run-app.log` plus the Z-Wave logs, but there could eventually be a lot more.

Maybe for 0.4, we can put a link somewhere in the UI that will allow the user to at least download the zip file. After 0.4, we could consider adding a log-reader UI.

benfrancis:
I didn't mean a log-reader UI necessarily, just that loading /logs in the browser would provide a plain text response which the browser can render, rather than having to download and unzip a file and open the file you wanted.

Is that not possible because of a missing JWT or something?

mrstegeman:
Correct, you'd be missing the JWT. I believe we have support for parsing the JWT out of a URL param, but we have to get it there somehow initially, i.e. via a link from the UI.

mrstegeman:
@benfrancis I will add some additional routes to do a directory listing and allow the user to view an individual file in the browser. Then we can decide what we actually want to make available from the UI (and where).

PR #738, opened 2018-03-06
Remove managed devices when an add-on is disabled.

mrstegeman:
This prevents unowned devices from showing up in the "Add Things"
screen.

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -756,7 +756,7 @@ class AddonManager extends EventEmitter {
     }
 
     const adapters = this.getAdaptersByPackageName(packageName);
-    const adapterNames = adapters.map((a) => a.id);
+    const adapterIds = adapters.map((a) => a.id);
     const unloadPromises = [];
     for (const a of adapters) {
       console.log('Unloading', a.name);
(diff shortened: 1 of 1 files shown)


src/addon-manager.js
```diff
@@ -755,8 +755,9 @@ class AddonManager extends EventEmitter {
       pluginProcess = plugin.process;
     }
 
-    let adapters = this.getAdaptersByPackageName(packageName);
-    let unloadPromises = [];
+    const adapters = this.getAdaptersByPackageName(packageName);
+    const adapterNames = adapters.map((a) => a.id);
```
hobinjk: nit: should be called adapterIds

src/addon-manager.js
```diff
@@ -770,6 +771,13 @@ class AddonManager extends EventEmitter {
           console.log(`Killing ${packageName} plugin.`);
           pluginProcess.p.kill();
         }
+
+        // Remove devices owned by this add-on.
+        for (const deviceId of Object.keys(this.devices)) {
+          if (adapterNames.includes(this.devices[deviceId].adapter.id)) {
+            delete this.devices[deviceId];
```
hobinjk: I don't think anything relies on the event, but this should probably call handleDeviceRemoved to perform the delete or emit a THING_REMOVED event.

PR #742, opened 2018-03-07
Add support for bulbs with changeable color temp.

### LINKED ISSUE #685: Add support for bulbs with color-temperature adjustment
Examples are IKEA Tradfri and several TP-Link bulbs.

mrstegeman:
Fixes #685

### DIFF
--- a/src/views/index.html
+++ b/src/views/index.html
@@ -34,6 +34,7 @@
     <script type="text/javascript" src="/js/thing-detail-layout.js"></script>
     <script type="text/javascript" src="/js/thing.js"></script>
     <script type="text/javascript" src="/js/color-detail.js"></script>
+    <script type="text/javascript" src="/js/color-temperature-detail.js"></script>
     <script type="text/javascript" src="/js/label-detail.js"></script>
     <script type="text/javascript" src="/js/level-detail.js"></script>
     <script type="text/javascript" src="/js/on-off-detail.js"></script>
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -212,11 +212,13 @@
   stroke-width: 2px;
 }
 
+.color-temperature,
 .level {
   width: 8rem;
   height: 8rem;
 }
 
+.color-temperature-input,
 .level-input {
   width: 8rem;
   margin: 0;
--- a/static/js/color-light.js
+++ b/static/js/color-light.js
@@ -26,9 +26,7 @@ function ColorLight(description, format) {
 
     if (description.properties.hasOwnProperty('color')) {
       this.details.color = new ColorDetail(this);
-    }
-
-    if (description.properties.hasOwnProperty('colorTemperature')) {
+    } else if (description.properties.hasOwnProperty('colorTemperature')) {
       const prop = description.properties.colorTemperature;
       this.details.colorTemperature =
         new ColorTemperatureDetail(this, prop.min, prop.max);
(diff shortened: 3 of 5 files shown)


mrstegeman:
Dimmable, only color temp:
screen shot 2018-03-07 at 10 16 58

Dimmable, color + color temp:
screen shot 2018-03-07 at 10 18 15

In the latter case, the icon color is determined by whichever property was set last.

benfrancis:
> In the latter case, the icon color is determined by whichever property was set last.

This sounds a bit odd, but maybe I don't understand. Presumably setting the colorTemperature property also changes the color property and vice versa? Can the color in the UI not just always be whatever color is? Or is that actually what you meant?

mrstegeman:
I tried to replicate what the TP-Link app was doing, and what their API reports. When you change the color temperature, the reported color does not change. However, when you change the color, their API sets the temperature to 0. Their app's UI shows whichever was set last as the current color.

For devices that don't behave this way, i.e. they change their color property when the temperature is modified, the UI will still reflect that.

mrstegeman:
Doing it this way made the icon match what my bulb was actually doing.

mrstegeman:
@benfrancis I believe most bulbs with this capability have separate LED arrays for temperature-shifting and for color, which leads to the disconnect.

hobinjk:
Philips Hue bulbs return in their API which property was last used to set the color of the bulb ("colormode", taking values "ct" (color temperature), "xy" (xy coordinates), or "hs" (hue and saturation)). I don't know about the others

mrstegeman:
@benfrancis Correct. When you set the color temperature, the color LEDs turn off. When you set the color, the temperature LEDs turn off.

In this PR, I have some code to convert color temperature to RGB. It's not perfect, but it's a decent approximation.

mrstegeman:
@benfrancis we could add another property that says which mode is currently active, color or color temperature. Maybe that would be better than relying upon magic values.

hobinjk:
Is there a significant difference between a light displaying white in color mode versus color temperature mode? At least for the philips hue lights there is no perceptible difference which makes me think that there's no need to support both colorTemperature and color

mrstegeman:
I guess the benefit is that is can be tricky to find "soft white" or "cool white" from a color picker, especially like the one provided by Firefox on Android.

benfrancis:
We had a chat. Our conclusion was that eventually we might want to implement our own colour picker which has temperature and brightness features and just have the colour picker.

In the meantime, we will use the hack that when setting color the light goes into color mode and colorTemperature goes to 0. When setting colorTemperature color goes to #ffffff. Level and color are linked, but level and colorTemperature are not. That was the least worst hack we could come up with.

benfrancis:
At the API level though, this is now making me wonder whether there should even be a colorTemperature property if it's really more of a UI thing and colorTemperature is just a subset of color...

hobinjk:
Ideally a bulb would receive a color and if the bulb contains a special low-gamut color-temperature array that can represent the color it uses that array and if it doesn't it uses the RGB array.

The way Philips Hue works is https://plus.google.com/photos/107696725527584609973/album/5806291983792940817/5806292012199918114 which suggests that it doesn't have a separate array.

benfrancis:
I guess if the hardware only supports color temperature then a colorTemperature property at the API level is necessary to restrict the set of values the user can input. If it supports both color temperature and color then it's really just a UI thing and the colorTemperature property is kind of redundant, as is a separate LED array if the main color array can represent all the colors the color temperature array can represent!

benfrancis:
Arguably this would be simpler at the API level if we actually had separate colorLight and colorTemperatureLight web thing types and any light which supports all RGB colors is a colorLight with just a single color property. Adding color temperature features to a color light is up to the UI.

mrstegeman:
@hobinjk I'm not entirely opposed to that, but it could be confusing to a user. If a user select magenta, and their color-temp-only-bulb just shows some flavor of "white", that would be extremely frustrating and confusing.

@benfrancis I agree that it would be simpler at the API level to have separate types for bulbs with only color temperature, and also that it makes sense to simplify the UI if a device has both. However, I think both are dependent on a new color picker with temperature-based (and other) presets.

benfrancis:
OK, let's go ahead as we previously agreed to get this working.

I might hold off on landing the colorTemperature property in the Web Thing API specification while we incubate this a bit more though, I don't want to design the API based on temporary UI constraints.

mrstegeman:
@benfrancis @hobinjk this should be good now, as only one of the color properties is allowed.

PR #744, opened 2018-03-07
Only roll back updates after a threshold is hit.

mrstegeman:


hobinjk:
If I apply this change to my local gateway's systemd service then kill the `node` process it goes directly to the OnFailure rollback instead of doing the StartLimit thing. I think instead of OnFailure the rollback should be specified as StartLimitAction

hobinjk:
Hmm, what I suggested about StartLimitAction is wrong but I'm not sure why OnFailure isn't behaving as expected

mrstegeman:
I believe this is the right way to do it according to the docs. How did you test, exactly? What I would do is:
1. Modify systemd unit
2. Reboot system or run `systemctl daemon-reload` to reload all the units
3. `kill -9 <pid>` a couple times.

It should, in theory, restart once, then go to the OnFailure action (I think).

hobinjk:
Those are the exact steps I took (I ran daemon-reload followed by restart then did the kill -9)

Someone else had the same question at https://serverfault.com/questions/786590/systemd-execute-command-after-start-limit-reached

PR #751, opened 2018-03-12
add json schema form to ui

### LINKED ISSUE #421: Add UI to allow users to configure an add-on
As a user I want to configure my add-on from the main UI

sogaani:
Fixes #421

I suggest that add-ons provide JSON-schema to configure.
This commit is a part that creates configuration form from JSON-schema.
This is porting from https://github.com/mozilla-services/react-jsonschema-form.

The following file is a sample of the form.
static/js/addons-config/index.html
static/js/addons-config/app.js

If you think that this suggestion is good, I will do the following Todo.

Todo: 
* Add CSS
* Add Array and Boolean input field.


benfrancis:
This is an interesting idea and is similar to how we generate the UI for Things from a Thing Description.

Is this level of complexity actually required for configuring add-ons though? Can you provide some example use cases? How does the developer hook up their own logic to the form?

Thinking about other solutions for #421...
At one extreme we could have just key-value pairs as values to configure (safest option). At another extreme an add-on could bundle its own HTML, CSS and JavaScript like an [options page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/user_interface/Options_pages) in Web Extensions (would need to be very careful about security).

I think we need a list of example use cases in #421 to help decide on the best solution.

sogaani:
@benfrancis 

> I think we need a list of example use cases in #421 to help decide on the best solution.

I have been thinking about three level complexity configuration cases.
If usecases are not enough to decide, I do not mind waiting for addon to come out a little more.

# First level: simple configuration.
* Addon has configuration items of fixed. (e.g. port number, enable experimental function boolean
* Configuration items may be structured.
```js
config: {
 a:{
  a-b:"hoge",
  a-c:"hoge"
 }
 b: "aa"
}
```
* Some configuration value have limitations. (e.g.  port number should be 1024 or more. gpio direction should be ```out``` or ```in```.
* There are options and required items in configuration items.

# Second level: A little complicated configuration.
* Assistance is required to enter configuration values.( e.g. Add devices with GPIO adapter, Add devices with broadlink adapter.

# Third level: Complicated configuration.
* Configuration value is dynamically. ( e.g. learn ir code with broadlink adapter.

# Feasibility
| level | key-value pair | JSON-schema |
|:-----------|:------------|:------------|
| first | can | can |
| second | ? | can |
| Third | can not | can not |

Could we support up to the second level with safest option ?

# Pros and Cons
## key-value pair
### Pros
* As a addon developper, easy to provide configuration because only need to know json.

### Cons
* As a addon developper, need to check configuration values that user input.
* As a addon developper, if errors is found in user input, addon may need to response a message prompting for correct value.
* As a user, may repeat form submission, because not know the correct input.

## JSON-schema
### Pros
* As a user, can enter configuration according to the guide.(e.g choice, format, value range

### Pros
* As a addon developper, learn JSON-schema.

static/js/addons-config/index.html
```diff
@@ -0,0 +1,31 @@
...
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#75AACF">
+    <title>JSON schema sample</title>
+    <link rel="icon" href="../../images/icon.png" type="image/png" />
+
+    <link rel="stylesheet" type="text/css" href="../../css/app.css" />
+
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/ajv/6.2.1/ajv.min.js"></script>
```
mrstegeman: I would prefer to pull AJV into `static/js/lib/` such that the UI can continue to work without an internet connection.

static/js/addons-config/number-field.js
```diff
@@ -0,0 +1,113 @@
...
+  this.formData = number;
+  if (this.onChange) {
+    this.onChange(this.formData);
+  }
+}
+
+NumberField.prototype.onNumberChange = function (value) {
+  const number = Number(value);
+
+  this.formData = number;
```
mrstegeman: nit: Don't set `this.formData` twice here. Just use an `else`.

static/js/addons-config/number-field.js
```diff
@@ -0,0 +1,113 @@
...
+
+  // range item
+  if (this.schema.hasOwnProperty('minimum') &&
+    this.schema.hasOwnProperty('maximum')) {
+    const field = document.createElement('div');
+    field.className = 'field-range-wrapper';
+    field.innerHTML = `
+        <input
+        type="range"
+        id=${id}
```
mrstegeman: There should be quotes around the `id` value.

static/js/addons-config/schema-field.js
```diff
@@ -0,0 +1,107 @@
...
+  this.required = required;
+  this.disabled = disabled;
+  this.readonly = readonly;
+
+  return this;
+}
+
+SchemaField.prototype.getFieldType = function () {
+  const FIELD_TYPES = {
+    //array: "ArrayField",
```
mrstegeman: These two would definitely be good to have.

static/js/addons-config/string-field.js
```diff
@@ -0,0 +1,91 @@
...
+        const selects = enumOptions.map(({ value, label }, i) => {
+            return `
+            <option key=${i} value=${value}>
+              ${label}
+            </option>`;
+        });
+
+        field.innerHTML = `
+        <select
+        id=${id}
```
mrstegeman: There should be quotes around the `id` value.

static/js/addons-config/string-field.js
```diff
@@ -0,0 +1,91 @@
...
+        class="form-control"
+        ${this.readonly ? 'readonly' : ''}
+        ${this.disabled ? 'disabled' : ''}
+        value=${value == null ? '' : value}
```
mrstegeman: There should be quotes around the `value` value.

static/js/addons-config/string-field.js
```diff
@@ -0,0 +1,91 @@
...
+        >
+        ${selects.join(' ')}
+        </select>`;
+    } else {
+        field.innerHTML = `
+        <input
+        id=${id}
```
mrstegeman: There should be quotes around the `id` value.

static/js/addons-config/string-field.js
```diff
@@ -0,0 +1,91 @@
...
+        class="form-control"
+        ${this.readonly ? 'readonly' : ''}
+        ${this.disabled ? 'disabled' : ''}
+        value=${value == null ? '' : value}
```
mrstegeman: There should be quotes around the `value` value.

static/js/addons-config/validater.js
```diff
@@ -0,0 +1,105 @@
+/**
```
mrstegeman: "Validater" should be "Validator" (everywhere in this file and elsewhere).

mrstegeman:
@sogaani This is fantastic! I tried out the sample and it's pretty cool. I'm looking forward to being able to do non-CLI add-on configuration 😄

sogaani:
@mrstegeman Thank you for review! I will develop TODOs which left and fix your comments this weeekend.

sogaani:
Work for JSON schema form is done.

Ultimately, it may be better to fix the following behaviors about validation error.
* Error is displayed from before input.
* Error item is not emphasized.

sogaani:
@mrstegeman 
I probably finished this work.


image

image

package.json
```
  "moziot": {
    "api": {
      "min": 1,
      "max": 2
    },
    "plugin": true,
    "exec": "{nodeLoader} {path}",
    "config": {
      "url": []
    },
    "schema": {
      "type": "object",
      "properties": {
        "url": {
          "type": "array",
          "title": "Urls",
          "items": {
            "type": "string",
            "default": "http://"
          }
        }
      }
    }
  }
```


It is necessary to change the format of config to use this configuration for the GPIO adapter and the Serial adapter.

now
```
    "config": {
      "gpios": {
        "18": {
          "name": "led",
          "direction": "out",
          "value": 0
        }
      }
    }
```

config to use this configuration.
```
    "config": {
      "gpios": [
         {
          "pin": 18,
          "name": "led",
          "direction": "out",
          "value": 0
        }
      ]
    },
    "schema": {
      "type": "object",
      "properties": {
        "gpios": {
          "type": "array",
          "items": {
            "type": "object",
            "required": [
              "pin",
              "name",
              "direction",
              "value"
            ],
            "properties": {
              "pin": {
                "type": "number"
              },
              "name": {
                "type": "string"
              },
              "direction": {
                "type": "string",
                "enum": [
                  "in",
                  "out"
                ]
              },
              "value": {
                "type": "number"
              }
            }
          }
        }
      }
    }
```
image

static/js/installed-addon.js
```diff
@@ -71,6 +74,10 @@ InstalledAddon.prototype.view = function() {
         </span>
       </div>
       <div class="addon-settings-controls">
+        <button id="addon-config-${Utils.escapeHtml(this.name)}"
+          class="text-button addon-settings-config ${configButtonClass}">
+          Config
```
mrstegeman: `Configure` would be better.

static/js/settings.js
```diff
@@ -414,6 +419,23 @@ var SettingsScreen = {
     });
   },
 
+  showAddonConfigScreen: function(id) {
+    this.backButton.href = '/settings/addons';
+    this.addonMainSettings.classList.add('hidden');
+    this.addonConfigSettings.classList.remove('hidden');
+
+    this.fetchAddonList().then(() => {
```
mrstegeman: You don't necessarily need to fetch the add-on list here. Instead, you can make sure it's already been fetched, like `showDiscoveredAddonsScreen` does:

```javascript
     let promise;
     if (this.installedAddons.size === 0 && this.availableAddons.size === 0) {
       promise = this.fetchAddonList();
     } else {
       promise = Promise.resolve();
     }

     promise.then(() => {
```

static/js/addons-config/number-field.js
```diff
@@ -0,0 +1,115 @@
...
+        class="form-control"
+        ${this.readonly ? 'readonly' : ''}
+        ${this.disabled ? 'disabled' : ''}
+        value=${value == null ? '' : value}
+        ${this.schema.multipleOf ?
+          'step=' + Number(this.schema.multipleOf) : ''}
+        min=${Number(this.schema.minimum)}
+        max=${Number(this.schema.maximum)}
+        />
+        <span class="range-view">${value}</span>`
```
mrstegeman: Missing semicolon.

static/js/addons-config/validator.js
```diff
@@ -0,0 +1,105 @@
...
+/* globals Ajv */
+
+var Validator = {};
+
+Validator._ajv = new Ajv({
+  errorDataPath: 'property',
+  allErrors: true,
+});
+
+Validator._reEscapeChar = /\\(\\)?/g
```
mrstegeman: Missing semicolon.

static/js/addons-config/validator.js
```diff
@@ -0,0 +1,105 @@
...
+  // Or match property names within brackets.
+  '\\[(?:' +
+  // Match a non-string expression.
+  '([^"\'].*)' + '|' +
+  // Or match strings (supports escaping characters).
+  '(["\'])((?:(?!\\2)[^\\\\]|\\\\.)*?)\\2' +
+  ')\\]' + '|' +
+  // Or match "" as the space between consecutive dots or empty brackets.
+  '(?=(?:\\.|\\[\\])(?:\\.|\\[\\]|$))'
+  , 'g')
```
mrstegeman: Missing semicolon.

static/js/addons-config/validator.js
```diff
@@ -0,0 +1,105 @@
...
+    function (match, expression, quote, subString) {
+      let key = match;
+      if (quote) {
+        key = subString.replace(Validator._reEscapeChar, '$1');
+      }
+      else if (expression) {
+        key = expression.trim();
+      }
+      result.push(key);
+    })
```
mrstegeman: Missing semicolon.

sogaani:
@mrstegeman 
I fixed the review point.

> This shouldn't really be happening with a checkbox, given that it's either checked or not checked. 

It is happen, if there is no default value on schema and config value.
If default is not set, set ```false``` as default.
If default is not set for range input with maximun and minimum, set minimum value as default.
This is a bit tricky.
Basically, it is better for addon developers to set the value of default to schema.

static/js/addons-config/array-field.js
```diff
@@ -0,0 +1,406 @@
...
+  }
+  return SchemaUtils.isObject(schema.additionalItems);
+};
+
+ArrayField.prototype.isAddable = function (formItems) {
+  const schema = this.schema;
+  let addable = true;
+
+  if (typeof schema.maxItems !== 'undefined') {
+    addable = addable && formItems.length < schema.maxItems;
```
mrstegeman: There's no need to do `addable &&` here, since addable is automatically true at this point.

static/js/addons-config/schema-form.js
```diff
@@ -0,0 +1,122 @@
...
+  } else if (!this.noValidate) {
+    this.errorField.render([]);
+  }
+};
+
+SchemaForm.prototype.validate = function (formData) {
+  return Validator.validateFormData(formData, this.schema);
+};
+
+SchemaForm.prototype.scrollToTop= function () {
```
mrstegeman: nit: add space before `=`

static/js/addons-config/schema-form.js
```diff
@@ -0,0 +1,122 @@
...
+  if (errors) {
+    this.scrollToTop();
+  } else {
+    this.applyButton.innerText = 'Applying...'
+    window.API.setAddonConfig(this.name, this.formData)
+    .then(() => {
+      page('/settings/addons');
+    })
+    .catch((err) => {
+      console.error(`Failed to set config add-on: ${this.name}\n${err}`);
```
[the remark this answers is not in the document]
sogaani: I will render err on error-field.

sogaani:
@mrstegeman 
I make remaining changes.

PR #755, opened 2018-03-13
Unpair from device when removing.

### LINKED ISSUE #350: Unpairing Devices
As a user, I want to unpair devices associated with the gateway

mrstegeman:
The actual unpairing process is left up to the adapter.

Fixes #350

### DIFF
--- a/src/test/integration/things-test.js
+++ b/src/test/integration/things-test.js
@@ -396,20 +396,6 @@ describe('things/', function() {
       }
     }
     expect(!found).assert('should not find thing in /things output');
-
-    res = await chai.request(server)
-      .get(Constants.NEW_THINGS_PATH)
-      .set('Accept', 'application/json')
-      .set(...headerAuth(jwt));
-    expect(res.status).toEqual(200);
-    expect(Array.isArray(res.body)).toBeTruthy();
-    found = false;
-    for (let thing of res.body) {
-      if (thing.href === Constants.THINGS_PATH + '/' + thingId) {
-        found = true;
-      }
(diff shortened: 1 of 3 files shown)


hobinjk:
Testing with virtual-things-adapter the first DELETE request didn't complete and the second errored with "Failed to remove thing virtual-things-0: Remove already in progress". This may be a side effect of vta using an out-of-date gateway-addon-node or something.

mrstegeman:
@hobinjk Can you try with this: https://github.com/mozilla-iot/gateway-addon-node/pull/1

mrstegeman:
@hobinjk I did confirm that the gateway-addon PR fixed this issue.

mrstegeman:
@dhylands Not currently. Z-Wave will continue to behave as it always has for now.

benfrancis:
This is arguably better than what we have now but I'm still not sure if it's the right UX. But I don't yet have a better suggestion.

src/test/integration/things-test.js
```diff
@@ -396,20 +396,6 @@ describe('things/', function() {
       }
     }
     expect(!found).assert('should not find thing in /things output');
-
-    res = await chai.request(server)
```
hobinjk: Why isn't the thing expected here? It looks like it should still be there but I might be wrong about how mock-adapter works

mrstegeman: There was previously some weirdness in that things could be present in the adapters but not in the gateway. Due to the changes in this patch set, that is no longer the case.

PR #756, opened 2018-03-14
Update README on setting up a *.mozilla-iot.org domains.

ziransun:
This is to make it clear for first time user on different
steps between using Mozilla's provided tunnel service and
using self-signed certificate.


mrstegeman:
@ziransun Thanks for clarifying this!

Issue #767, opened 2018-03-19
enable CORS for WEB thing API call


aragnac:
Hi,
I'm currently developping an angular app using the WEB Thing API of the mozilla IoT gateway. My problem is that the "Cross-Origin Request" sharing option seems to be inactivated on the server side.  I got this king of errors with status code 503. Which file(s) do I have to modify in order to allow CORS on server side ? 

Best regards,

Nicolas

![203error](https://user-images.githubusercontent.com/12186566/37599583-605f8962-2b85-11e8-8383-308d66f9505c.png)

sogaani:
If you want to allow "Cross-Origin Request" with all http request,  will be good to modify function:```configure```  in file:```src/router.js```.

You probably need to add the following code:
```
app.use(function(req, res, next) {
  res.header("Access-Control-Allow-Origin", "*");
  res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept");
  next();
});
```

aragnac:
thanks for your quick answer @sogaani,

I tried to add the following lines but I still got the same errors. If I call the next();  I've got a "can't get/" error. 

Nicolas

![corserror](https://user-images.githubusercontent.com/12186566/37602937-c3515138-2b8d-11e8-9694-3b292e8f70c8.png)

sogaani:
Sorry, I missed that we need to allow Authorization header at least.
```
      response.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept, Authorization");
```

sogaani:
@aragnac
Do you need OPTIONS request?
I think that maybe OPTIONS method is not implemented.

Following code work in my browser.
```
fetch('https://[domain].mozilla-iot.org/things/virtual-things-0',{headers:{Authorization:'Bearer [TOKEN]'}}).then(res => {console.log(res),res.json()}).then((json) => {console.log(json)}).catch((err) => console.log(err));
```

hobinjk:
@aragnac I'm sorry that you're encountering this issue! If you could provide the full list of request headers from the network tab it would be very helpful to understanding what's going on.

mrstegeman:
@aragnac Can you see if #768 fixes your issue, please?

aragnac:
@mrstegeman as mentioned upper in this issue, no it doens't fix it, unfortunetely.

hobinjk:
[...]

PR #768, opened 2018-03-19
Enable CORS

hobinjk:
Based on @sogaani's comment in #767, should make things a bit happier for Angular and other uses of xhr instead of fetch/ws.


hobinjk:
Hmmm, I don't think there's any app request that we want to prohibit third parties from performing but we may as well just to be safe

hobinjk:
Turns out the OPTIONS request sets the Accept header to 'text/html' so enabling CORS does need to take place unconditionally.

mrstegeman:
Weird. Ok.

Issue #697, opened 2018-02-20
OTA update pings


benfrancis:
As Mozilla I want to log when gateways contact the update server so that I can get a better idea of the number of active gateways which may need updating

Currently we can only count the number of active gateways using our TLS tunnelling service and a *.mozilla-iot.org domain. It appears that many users are opting out of this service and choosing to use their own DNS configuration.

If we anonymously logged when a gateway contacts the OTA update server for updates, this would give us a more complete picture of how many active gateway they are requiring the use of the OTA update server so we can plan resources accordingly.

As updates are currently statically hosted on GitHub this may require changing our update infrastructure to Mozilla infrastructure in order to log this information.

hobinjk:
The https://api.github.com/repos/mozilla-iot/gateway/releases endpoint lists the number of downloads for each component of the OTA update which right now shows that 5 attempts were made to OTA update to 0.3.0 from 0.2.1 and one confused person downloaded the gateway component thinking it was the gateway image.

PR #776, opened 2018-03-21
Make update URL configurable.

mrstegeman:
This is the first step required for #697.

### DIFF
--- a/tools/check-for-update.js
+++ b/tools/check-for-update.js
@@ -1,61 +1,63 @@
+const config = require('config');
+const exec = require('child_process').exec;
 const fetch = require('node-fetch');
 const fs = require('fs');
 const semver = require('semver');
 
-const exec = require('child_process').exec;
+const pkg = require('../package.json');
 
-const remote = 'mozilla-iot';
-
-const pkg = JSON.parse(fs.readFileSync('package.json', {encoding: 'utf-8'}));
-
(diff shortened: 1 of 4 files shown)


mrstegeman:
I suggest turning whitespace off in the diff: https://github.com/mozilla-iot/gateway/pull/776/files?w=1

tools/check-for-update.js
```diff
@@ -1,61 +1,63 @@
...
-      }
-      if (asset.name.match(/^gateway-[a-f0-9]+\.tar\.gz$/)) {
-        gatewayUrl = asset.browser_download_url;
-      }
-      if (asset.name.match(/^node_modules-[a-f0-9]+\.tar\.gz$/)) {
-        nodeModulesUrl = asset.browser_download_url;
-      }
+fetch(config.get('updateUrl'),
+      {headers: {'User-Agent': `mozilla-iot-gateway/${pkg.version}`}})
+  .then(res => {
```
hobinjk: nit: should be consistent with line 13 (which means it wouldn't have the extra indentation either)

tools/check-for-update.js
```diff
@@ -1,61 +1,63 @@
+const config = require('config');
+const exec = require('child_process').exec;
 const fetch = require('node-fetch');
 const fs = require('fs');
```
hobinjk: unused import causing tests to fail

PR #777, opened 2018-03-21
UI Fix: Hide authorization screen before login. (#757)

### LINKED ISSUE #757: OAuth:Appear authorization screen before login.
Authorization screen is Temporarily shown  when using oauth with not logined client.
We could put ```class='hidden'``` on body same as gateway root to avoid it.

ziransun:
Authorization screen is temporarily shown when using oauth
with not logined client. The fix is to put class='hidden'
on body for /oauth/authorize page.

Fixes #757


mrstegeman:
Thanks!

PR #782, opened 2018-03-21
Don't auto-update add-ons with .git directory.

### LINKED ISSUE #780: Auto updating of add-ons should be disabled if .git directory is present
Everything was good until I bumped the version in my package.json file, then it promptly wiped out all of my changes.

mrstegeman:
Fixes #780


mrstegeman:
@dhylands I added a log.

PR #784, opened 2018-03-21
Minor fixes to match WoT spec.

mrstegeman:


mrstegeman:
I always forget we store full thing descriptions in the database. I'll add fallbacks.

PR #785, opened 2018-03-21
Set profileDir from MOZIOT_HOME env var if it exists.

dhylands:


config/test.js
```diff
@@ -8,12 +8,8 @@
 
 'use strict';
 
-const os = require('os');
-const home = os.homedir();
-
 module.exports = {
   cli: false,
-  profileDir: `${home}/.mozilla-iot/test`,
```
dhylands: I removed this since it's already set in default.js and we don't need to set it a second time.

mrstegeman:
@dhylands You made Travis CI very unhappy.

dhylands:
I decided to abandon this PR.

I can't explain it, but the mere presence of process.env seems to cause intermittent test failures. I instead opted to put my override in local.js and local-test.js and not use process.env and the intermittent failures stopped.

PR #794, opened 2018-03-22
Implement and expose scope/readwrite restriction

hobinjk:
Adds a `scope` value to access token's payloads, restricting their use to a subset of paths on the gateway.

For example, '/things:readwrite' allows all access to all things. '/things/light1:read' allows only the GET and OPTIONS methods on '/things/light1'.

auth-request


mrstegeman:
I have one small suggestion that I think would make the UI a bit more intuitive. If "Allow for all things" is checked, can you disable the other checkboxes? It's a little confusing that I can check "Allow for all things", but then disable selected other things.

hobinjk:
Oh, my mistake. That represents a bug with the approach I was taking but I think disabling all the checkboxes is a better approach to take anyway.

hobinjk:
(updated accordingly)

PR #804, opened 2018-03-26
Add Local Token Service

hobinjk:
Builds on #794 to allow users to create a new authorization

settings-screen

auth-request

local-token-service

screen shot 2018-03-26 at 2 39 20 pm


src/controllers/oauth_controller.ts
```diff
@@ -109,6 +109,10 @@ function redirect(response: express.Response, baseURL: URL, params: {[key: strin
       url.searchParams.set(key, params[key].toString());
     }
   }
+  if (url.hostname === 'gateway.localhost') {
```
mrstegeman: `gateway.local`?

src/controllers/oauth_controller.ts
```diff
@@ -292,7 +315,10 @@ OAuthController.get('/allow', auth, async (request: express.Request, response: e
 OAuthController.post('/token', async (request: express.Request, response: express.Response) => {
   const requestData = request.body;
   if (requestData.grant_type === 'authorization_code') {
-    handleAccessTokenRequest(request, response);
+    let token = await handleAccessTokenRequest(request, response);
```
mrstegeman: No response is sent in the error case here.

src/models/oauthclients.ts
```diff
@@ -54,4 +54,9 @@ oauthClients.register(
                      'Test OAuth Client', 'super secret', '/things:readwrite')
 );
 
+oauthClients.register(
+  new ClientRegistry(new URL('https://gateway.localhost/oauth/local-token-service'), 'local-token',
```
mrstegeman: `gateway.local`?

src/views/local-token-service.mustache
```diff
@@ -0,0 +1,143 @@
...
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#75AACF">
+    <title>Things Gateway - Local Token Service</title>
+    <link rel="manifest" href="/app.webmanifest">
+    <link rel="icon" href="/images/icon.png" type="image/png" />
+
+    <link rel="stylesheet"
+          href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/default.min.css">
```
mrstegeman: Do we want to pull the highlight.js lib into our tree for cases with no internet access?

src/views/local-token-service.mustache
```diff
@@ -0,0 +1,143 @@
...
+
+    <!-- OAuth Authorization -->
+    <section class="selected">
+      <div class="title-bar">
+        <a href="/" class="back-button icon-button"></a>
+        <h1>Local Token Service</h1>
+      </div>
+
+      <div class="main-container">
+        Your local token is "{{token}}".
```
mrstegeman: The token would look better inside a `<code>` or `<pre>` block.

src/controllers/oauth_controller.ts
hobinjk: Thanks for pointing this out! I didn't realize `handleAccessTokenRequest` had cases where it doesn't send an error.

hobinjk: Oh, never mind, it does. I'll document the invariant properly though

src/controllers/oauth_controller.ts
hobinjk: I'm using gateway.localhost as a placeholder for redirecting back to the gateway. I chose the .localhost TLD to be reminiscent of gateway.local but not step on the toes of anyone who wants to use gateway.local in a client definition.

hobinjk:
I've updated the PR accordingly

benfrancis:
@hobinjk What is the use case for this feature? At first glance it looks like a developer feature which shouldn't necessarily be exposed through the UI for end users.

hobinjk:
@benfrancis The general idea is that we keep telling people interested in trying out the WoT API to copy-paste `API.jwt` which requires them to be comfortable with the browser console and will leave them with a user role token that stops working if they log out and can perform any operation on the Gateway.

The idea with this feature is to create a good onboarding experience for the WoT API. Since we don't have a "developer mode" I thought that putting it in the Authorizations section would be good since casual users will not accidentally discover it but less tech-savvy users will be able to click over to it instead of manually typing in a URL.

PR #806, opened 2018-03-27
Implement add-on actions and events.

### LINKED ISSUE #717: Actions and events are not preserved across the plugin bridge
1. Edit virtual-things-adapter so a device has actions and events (i.e. in the Device constructor `this.actions = new Map(); this.actions.set(...)`)
2. Add this device
3. Device's thing description in the database still has `actions: {}, events: {}`

mrstegeman:
Fixes #717

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -15,8 +15,6 @@ const config = require('config');
 const Constants = require('./constants');
 const EventEmitter = require('events').EventEmitter;
 const Deferred = require('./deferred');
-const PluginClient = require('./plugin/plugin-client');
-const PluginServer = require('./plugin/plugin-server');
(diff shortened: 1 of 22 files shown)


mrstegeman:
This is going to fail on Travis CI until the following PR is merged: https://github.com/mozilla-iot/gateway-addon-node/pull/4

hobinjk:
It would be nice to have a test of actions and events through mock-adapter but that can be left to future work if you'd prefer.

mrstegeman:
@hobinjk We should definitely do that. The actions, things, and rules engine tests already test the actions interface pretty thoroughly, and the rules engine tests events, but it would be good to expand that a bit.

PR #809, opened 2018-03-28
Finish making gateway compliant with WoT spec.

mrstegeman:


src/controllers/things_controller.js
```diff
@@ -296,7 +303,10 @@ ThingsController.ws('/:thingId/', function(websocket, request) {
           const actionParams = request.data[actionName].input;
           Things.getThing(thingId).then(thing => {
             let action = new Action(actionName, actionParams, thing);
-            return Actions.add(action);
+            return Actions.add(action).then(() => {
+              AddonManager.requestAction(
```
hobinjk: I believe that this should be `return`ed so that any errors bubble into the `websocket.send` error instead of being an unhandled promise rejection.

src/jwt-middleware.js
```diff
@@ -89,7 +89,7 @@ function middleware() {
     authenticate(req, res).
       then((jwt) => {
         if (!jwt) {
-          res.sendStatus(401);
+          res.status(401).end();
```
hobinjk: Is there a difference here?

src/test/integration/things-test.js
```diff
@@ -653,12 +651,10 @@ describe('things/', function() {
     expect(res).toBeTruthy();
 
     expect(messages[0].messageType).toEqual(Constants.EVENT);
-    expect(messages[0].data.name).toEqual(eventAFirst.name);
-    expect(messages[0].data.description).toEqual(eventAFirst.description);
+    expect(messages[0].data).toHaveProperty(eventAFirst.name);
```
hobinjk: This should also verify the event's data/description

src/jwt-middleware.js
mrstegeman: `sendStatus()` includes the actual error text. That should be fine, but it ended up causing test failures, so I switched it.

src/test/integration/things-test.js
mrstegeman: Events no longer include a description, but since I didn't update the calls to `new Event()`, what was previously the description is now the data. I'll update.

src/controllers/things_controller.js
mrstegeman: Good catch.

mrstegeman:
@hobinjk Updated.

PR #812, opened 2018-03-29
Speed up loading page.

sogaani:
In order to speed up loading web pages, i make the following changes.
* packing static files into one file.
* lazy loading some lib/*js files.

In my case the loading speed doubled.
I hope that this pull request will be effective in other's environment as well.


src/views/authorize.mustache
```diff
@@ -14,8 +14,11 @@
     <link rel="stylesheet" type="text/css" href="/css/thing.css" />
     <link rel="stylesheet" type="text/css" href="/css/authorize.css" />
 
+<<<<<<< HEAD
```
mrstegeman: Merge error 😄

mrstegeman:
@sogaani You'll also need to run `yarn` and commit the updated `yarn.lock` file.

hobinjk:
I looked into this a while ago and the primary concern at the time was to ensure that the compiled files would have corresponding source maps so debugging isn't negatively effected. I'll take a look at this when it's in a more finished state.

sogaani:
Still test fail.
I have no idea to fix it. Are there any advice?
```
  ● http redirector › should redirect GET /
    Not Found
      
      at Test.Object.<anonymous>.Request.callback (node_modules/superagent/lib/node/index.js:626:17)
      at IncomingMessage.<anonymous> (node_modules/superagent/lib/node/index.js:795:18)
      at emitNone (events.js:111:20)
      at IncomingMessage.emit (events.js:208:7)
      at endReadableNT (_stream_readable.js:1064:12)
      at _combinedTickCallback (internal/process/next_tick.js:138:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

sogaani:
I figured out cause of fail.
We need to exec webpack before test.

mrstegeman:
@sogaani I just rebased this patch set, since I just modified a ton of stuff with eslint. Hopefully I didn't break anything too badly.

sogaani:
@mrstegeman  Thank you! There was no problem with rebase. I fixed my part because there was a problem. (module.exports of Rule Engin)

sogaani:
rebase to resolve conflict.
@hobinjk Could you check this pull request? I believe that this is already work.

hobinjk:
This looks good to me, everything works as it's expected to except for the recently added src/view/local-token-service.mustache (e.g. https://localhost:4443/oauth/local-token-service?code=helloworld).

While I do have some concerns about the extra steps this adds to the build process, I think that the preservation of debugging information is really great. The performance characteristics of this are also very impressive, reducing the time-til-interactive by 16s to 6s on first view and 11s to 4s on repeat.

image of console showing debug information

before: https://www.webpagetest.org/result/180417_CD_1af6488672afee15c840d8b87e12c17b/
after: https://www.webpagetest.org/result/180417_XQ_efe5c779f5acec1ee75f93581e71215d/

sogaani:
Fix conflict and typo.

@hobinjk Thank you for reviewing.
> to except for the recently added src/view/local-token-service.mustache
Maybe we need to access ```https://localhost:4443/oauth/authorize?response_type=code&client_id=local-token&scope=/things:readwrite&state=asdf```.
Above URL works for me.

PR #813, opened 2018-03-29
CSS tweaks for add-on config screen.

mrstegeman:


mrstegeman:
@hobinjk A lot of this is whitespace. https://github.com/mozilla-iot/gateway/pull/813?w=1

static/css/addons-form.css
```diff
@@ -1,171 +1,204 @@
...
+}
+
+.addons-form > p > .form-group > fieldset > legend {
+  font-size: 2.5rem;
+  border-bottom: none;
+  padding-bottom: 2rem;
+}
+
+.addons-form > p > .form-group > fieldset > legend:before {
+  content: 'Configuring: ';
```
hobinjk: Is it possible to do this in HTML instead?

mrstegeman: Not easily. The form is completely auto-generated, and the top-level header is generated the exact same way as every nested header.

hobinjk: Okay, SGTM then

Issue #641, opened 2018-02-08
Investigate recompiling openzwave after migration from 0.3.0 image to Node LTS


hobinjk:
Also have a post-update script that can perform things like this
```
nvm install --lts
nvm use --list
nvm alias default node
```

hobinjk:
It will also require rebuilding OpenZWave which will be interesting

dhylands:
Yeah - I'm not sure we have all of the resources installed in the distributed image to rebuild (like some of the -dev modules). We shouldn't need to rebuild OpenZWave itself (i.e. the C++ library), just the node-openzwave-shared module

mrstegeman:
Related to #632

hobinjk:
Fixed by #814

hobinjk:
Although the open-zwave thing is still a concern

dhylands:
The addon-builder can now build node-openzwave-shared as one of the dependencies of the zwave-adapter.

hobinjk:
Fixed since it's a dependency now

PR #814, opened 2018-03-29
Add post-upgrade script

### LINKED ISSUE #632: Add post-update script to OTA
When doing things like migrating to an LTS version of Node or updating the python add-on library we need a function to run arbitrary scripts after the OTA update successfully completes.

Admittedly this is going to make the OTA process much more difficult to reason about since it will lose its ability to guarantee that it can completely revert an update

hobinjk:
Fix #632. Progresses forward on #641 but the zwave issue needs
consideration.

### DIFF
--- a/tools/post-upgrade.sh
+++ b/tools/post-upgrade.sh
@@ -7,3 +7,8 @@ nvm install ${NODE_VERSION}
 nvm use ${NODE_VERSION}
 # always use most recent version (lts)
 nvm alias default node
+
+# Upgrade gateway-addon Python package
+_url="git+https://github.com/mozilla-iot/gateway-addon-python#egg=gateway_addon"
+sudo pip2 install -U "$_url"
+sudo pip3 install -U "$_url"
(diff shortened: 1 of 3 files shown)


mrstegeman:
@hobinjk Could you add the following to `post-upgrade.sh`, please?

```bash
# Upgrade gateway-addon Python package
_url="git+https://github.com/mozilla-iot/gateway-addon-python#egg=gateway_addon"
sudo pip2 install -U "$_url"
sudo pip3 install -U"$_url"
```

hobinjk:
@mrstegeman Updated accordingly

mrstegeman:
Thanks! Are we worried about this messing things up if the update is rolled back?

hobinjk:
The node update will cause issues if the update fails. This can be resolved if we want to create an 0.3.2 release with `nvm alias default v7.10.1` in `upgrade-rollback.sh`. I'm not sure if we are that concerned at this stage but we should attempt to limit this script to backwards-compatible changes in the future.

hobinjk:
Oh wait, the rollback script used is actually this one. I'll update accordingly

PR #815, opened 2018-03-29
Allow users to add native web thing by URL from UI.

### LINKED ISSUE #69: Add Thing by URL
As a user I want to add a Web Thing by its URL, to be managed by my gateway.

This is for Web Things which use the direct integration pattern so provides its own Web Thing API but which the user might want to manage together with other Things using the gateway.

mrstegeman:
Fixes #69

### DIFF
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -10,7 +10,7 @@
 
 'use strict';
 
-const express = require('express');
+const PromiseRouter = require('express-promise-router');
 const Action = require('../models/action');
 const Actions = require('../models/actions');
 const ActionsController = require('./actions_controller');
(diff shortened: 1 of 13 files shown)


src/controllers/things_controller.js
```diff
@@ -34,22 +35,60 @@ ThingsController.get('/', function (request, response) {
...
+    webthing = true;
+
+    const key = 'addons.thing-url-adapter';
+    try {
+      const current = await Settings.get(key);
+      if (typeof current === 'undefined') {
+        throw new Error('Setting is undefined.');
+      }
+
+      current.moziot.config.urls.push(description.webthingUrl);
```
hobinjk: What happens if the list of urls already contains the web thing's url? (user presses enter twice or just forgets)

mrstegeman: It will be handled by the adapter just fine, since the adapter keeps track of "known URLs" when loading things.

PR #822, opened 2018-03-30
Add support for numerical and color properties to PropertySelect

hobinjk:
- [x] Investigate issue where rules appear to be duplicated
- [x] Clean up UI

### DIFF
--- a/static/js/rules/PropertySelect.js
+++ b/static/js/rules/PropertySelect.js
@@ -42,18 +42,117 @@ PropertySelect.prototype.clearOptions = function() {
 /**
  * Add an option to the options list
  * @param {String} name - text content of option
- * @param {Object} value - associated data
+ * @param {Object} ruleFragment - associated rule fragment with either trigger
+ * or property defined
  * @param {boolean} selected - if the option is selected
  */
-PropertySelect.prototype.addOption = function(name, value, selected) {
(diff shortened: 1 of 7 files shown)


hobinjk:
UI now looks like this: 
screen shot 2018-04-02 at 11 20 39 am

mrstegeman:
There seem to be some CSS issues:
screen shot 2018-04-05 at 08 35 52

hobinjk:
numberUnitProperty and its siblings are a bit too long to reasonably display but this is the best I've come up with so far:
screen shot 2018-04-05 at 12 27 02 pm

static/js/rules/PropertySelect.js
```diff
@@ -53,7 +53,10 @@ PropertySelect.prototype.addOption = function(name, ruleFragment, selected) {
     elt.classList.add('selected');
   }
   elt.dataset.ruleFragment = JSON.stringify(ruleFragment);
-  elt.textContent = name;
+  const nameElt = document.createElement('span');
+  nameElt.classList.add('property-select-name');
+  nameElt.textContent = name;
```
mrstegeman: If you set the `title` attribute here, we'll get some hover text. At least then you can differentiate the properties.

hobinjk: Oh, nice idea! I'll add it

Issue #778, opened 2018-03-21
Script for issuing a token


ziransun:
Create shell script for issuing a token that allow users to experiment with their gateways

ziransun:
I'd like to work on this if it's okay

hobinjk:
Definitely! Thanks for filing the issue. Let me know if you have any questions

ziransun:
I do have a question :).

My initial thought is to create a .js file (e.g. issue_token.js) that requires JSONWebToken module. And then in a .sh file, run "node issue_toke.js". 

In JSONWebToken module, it calls "const ec = require('../ec-crypto');", If I understand correctly the source for "ec-crypto" is a typescript file. From discussions on #179 it seems that we do need webpack or some kind of build/compile process for this.  For a script in /tools, I guess it is an independent code from webpack. Is it right?

Any suggestions?

hobinjk:
Oh no! This will be a bit more difficult than I thought. It looks like https://github.com/TypeStrong/ts-node should work to run the file.

To work around an issue with `ts-node` the exact command to use is `ts-node -O '{"module": "commonjs"}' issue_token.js`.

The shell file could make sure that the `ts-node` command exists and install it otherwise.

ziransun:
Thanks!

PR #842, opened 2018-04-05
Add shell script for issuing JWT Token. (#778)

ziransun:


mrstegeman:
@ziransun `eslint` found some issues

dhylands:
There is also https://github.com/mozilla-iot/curl-examples/blob/master/login.sh
The curl example doesn't appear to do everything that this one does, so I'm mostly just pointing it out to ensure that you're aware of it.

tools/issue-token.sh
```diff
@@ -32,6 +32,6 @@ echo "email address is: $email"
 read -p "Enter Your Password: "  password
 read -p "Enter Your Name: "  name
 
-echo "{\"id\":\"${id}\",\"email\":\"${email}\",\"password\":\"${password}\",\"name\":\"${name}\"}" > user.json
+echo "{\"email\":\"${email}\",\"password\":\"${password}\",\"name\":\"${name}\"}" > user.json
```
dhylands: Storing the password seems unnecessary, and insecure (it seems to getting stored in plaintext)

ziransun: Good point. Thanks.

PR #847, opened 2018-04-06
Manually set NVM_DIR

hobinjk:
Fix an issue where nvm wouldn't run properly in post-upgrade.sh due to the strange way it's invoked.

### DIFF
--- a/tools/post-upgrade.sh
+++ b/tools/post-upgrade.sh
@@ -2,7 +2,9 @@
 
 # Performs any necessary steps after the main upgrade process is complete.
 
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+# For some reason NVM_DIR isn't set at this point
+export NVM_DIR=/home/pi/.nvm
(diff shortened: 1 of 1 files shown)


tools/post-upgrade.sh
```diff
@@ -2,7 +2,9 @@
 
 # Performs any necessary steps after the main upgrade process is complete.
 
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+# For some reason NVM_DIR isn't set at this point
+export NVM_DIR=/home/pi/.nvm
+\. "$NVM_DIR/nvm.sh"  # This loads nvm
```
dhylands: It's because installing nvm appends those lines to the end of .bashrc and expects you to open a new terminal in order to activate nvm. So if you happen to install nvm into the current shell, you need to execute both of those lines to activate it.

nvm is actually a ton of aliases that need to be sourced into the current shell, rather than being a program.

hobinjk: This is running within a process which should have had .bashrc executed already

hobinjk: But yes, you're right, both lines are necessary

PR #849, opened 2018-04-06
Move add-on version into each package definition.

mrstegeman:


static/js/settings.js
```diff
@@ -409,7 +409,8 @@ const SettingsScreen = {
         let updateUrl = null, updateVersion = null, updateChecksum = null;
         if (this.availableAddons.has(name)) {
           const available = this.availableAddons.get(name);
-          if (available.version !== addon.version) {
+          const cmp = this.compareSemver(addon.version, available.version);
```
hobinjk: The compareSemver function can be replaced with semver.compare

mrstegeman: Not on the UI side...

mrstegeman:
Depends on https://github.com/mozilla-iot/addon-list/pull/47

PR #853, opened 2018-04-09
V0.4.0

Mikro007:
Why not include MySensors in this? https://www.mysensors.org/


mrstegeman:
@Mikro007 I'm not sure why this is a PR. We don't have an add-on for MySensors, so there's nothing we can include.

PR #860, opened 2018-04-09
Prevent adding duplicate web things.

### LINKED ISSUE #854: Can add same Web Thing multiple times
The UI lets me add a web thing using the same thing URL multiple times, but only the first instance appears in the Things screen

STR:
* Add thing by URL
* Add another thing by same URL

Expected:
* Error: "Web thing already added"

Actual:
* Adding allowed, but no new thing created.

mrstegeman:
Fixes #854

### DIFF
--- a/static/js/new-web-thing.js
+++ b/static/js/new-web-thing.js
@@ -224,14 +224,24 @@ NewWebThing.prototype.save = function() {
       'Content-Type': 'application/json',
     },
   }).then((response) => {
+    if (!response.ok) {
+      return response.text();
+    }
+
     return response.json();
-  }).then((json) => {
-    console.log(`Successfully created thing ${json}`);
+  }).then((description) => {
+    if (typeof description === 'string') {
+      throw new Error(description);
+    }
+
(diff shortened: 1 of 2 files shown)


static/js/new-web-thing.js
```diff
@@ -224,14 +224,24 @@ NewWebThing.prototype.save = function() {
...
+
+    console.log(`Successfully created thing ${description}`);
     this.nameInput.disabled = true;
     this.saveButton.innerHTML = 'Saved';
     this.saveButton.disabled = true;
     this.cancelButton.classList.add('hidden');
   }).catch((error) => {
-    console.error('Failed to save web thing:', error);
+    this.thingTypeLabel.innerText = error.message;
+    this.thingTypeLabel.classList.add('error');
```
benfrancis: Do we need to remove this class again in the non-error state, or is that already done somewhere else?

mrstegeman: Try now. I move the user back to the original URL input, since there's nothing else to be done at this point.

PR #870, opened 2018-04-11
install: Use sudo when needed in prepare-base-root.sh

rzr:
Change-Id: I4d6606897123d67e70299c0ce54949f58f59bd18

### DIFF
--- a/image/prepare-base-root.sh
+++ b/image/prepare-base-root.sh
@@ -21,16 +21,16 @@ npm install
 chmod +x run.sh
 
 # install and enable the systemd service
-cp config/mozilla-gateway-wifi-setup.service /etc/systemd/system/mozilla-gateway-wifi-setup.service
-systemctl enable mozilla-gateway-wifi-setup.service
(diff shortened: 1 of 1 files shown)


dhylands:
prepare-base-root.sh is already run under sudo from here:
https://github.com/mozilla-iot/gateway/blob/2acabaee1bb9dbb1f6d04026ba9ff9186d9dd9ea/image/prepare-base.sh#L140

rzr:
True I didn't notice that, but anyway it is also using sudo in some parts, so I made it runable by user.

Actually both scripts could be merged, it's confusing...

dhylands:
The intention was that prepare-base-root.sh would be run under sudo and not have to have sudo's all through it. If we sudo-ize it then there isn't any point in it being a separate script and it should be merged with prepare-base.sh

rzr:
yea prepare-base.sh is using sudo already, but using sudo once has some advantages too, it will not ask password again on long operations.

mrstegeman:
Replaced by #1179

PR #876, opened 2018-04-12
Update readme for windows user

sogaani:


README.md
```diff
@@ -108,6 +108,14 @@ Install [OpenSSL](http://slproweb.com/products/Win32OpenSSL.html) =< 1.0.2(norma
 
 [See here](https://github.com/JoshKaufman/ursa#windows-install).
 
+### Install python (Windows only)
+
+Pagekite and some addons work with python.
+
+Install Python 2.* from [here](https://www.python.org).
```
mrstegeman: The Python add-ons we have so far are Python 3.x, but Pagekite only works with 2.x. So could you update the instructions to have the user install both?

sogaani: I found several steps that the python add-ons work on windows.
I updated readme with these steps.

PR #884, opened 2018-04-12
Wait longer after killing a plugin.

mrstegeman:


mrstegeman:
Wait on this... not sure this is actually working.

PR #886, opened 2018-04-12
install: Add i2c enabler to make-prep.sh

rzr:
Should work with default config.txt where line is commented.

This is needed for building generic-sensors-addons:

https://discourse.mozilla.org/t/creating-i2c-add-on/26696/4

Change-Id: Ifb8006ded46ce78e537b1a5bf74453fcd635abb8

### DIFF
--- a/image/make-prep.sh
+++ b/image/make-prep.sh
@@ -15,6 +15,7 @@ SCRIPT_NAME=$(basename $0)
 SCRIPT_DIR=$(dirname $0)
 VERBOSE=0
 ENABLE_CONSOLE=1
+ENABLE_I2C=1
 ENABLE_WIFI=0
 WIFI_SSID=
 WIFI_PASSWORD=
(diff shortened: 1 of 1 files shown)


rzr:
Relate-to: https://github.com/mozilla-iot/addon-list/pull/54

image/make-prep.sh
```diff
@@ -283,6 +317,7 @@ main() {
     echo "  IMG_FILENAME = ${IMG_FILENAME}"
     echo " PREP_FILENAME = ${PREP_FILENAME}"
     echo "ENABLE_CONSOLE = ${ENABLE_CONSOLE}"
+    echo "ENABLE_I2C = ${ENABLE_I2C}"
```
dhylands: nit: The = are lined up on the other prints

rzr:
If you're ok to enable i2c in default images, that will be great, then I will update patch to add --no-i2c option to disable it.

dhylands:
Yeah - I think that makes sense. It won't go out in the 0.4.0 release, but I don't see any reason why it shouldn't be enabled by default.

rzr:
Hi, I checked on master  (also with  --no-i2c), so it should be safe to merge for v0.4.1

PR #887, opened 2018-04-12
Deduplicate new things.

### LINKED ISSUE #883: Duplicate discovered web thing appears during add by URL
STR:
* From Things screen click "+"
* Click "Add by URL..."
* Enter URL of a native Web Thing (which doesn't use mDNS or Bluetooth broadcast)
* Click Save
* Wait a few seconds

Expected:
* Discovered thing goes into "saved" state, then nothing happens until you click Done

Actual:
* A new (duplicate) discovered thing appears asking if you want to save it

It seems that the thing url add-on is telling the UI that a new thing has been discovered at the new URL it just learned about, even though that URL was manually added by the user. The gateway needs to de-dupe the list of discovered devices with the list of things which have already been saved.

mrstegeman:
Fixes #883

### DIFF
--- a/src/models/things.js
+++ b/src/models/things.js
@@ -116,9 +116,11 @@ const Things = {
    * @param Object New Thing description
    */
   handleNewThing: function(newThing) {
-    // Notify each open websocket of the new Thing
-    this.websockets.forEach(function(socket) {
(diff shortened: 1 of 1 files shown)


benfrancis:
Travis is unhappy, but surely unrelated.

PR #890, opened 2018-04-12
Exit plugin process after unload.

mrstegeman:


src/addon-loader.js
```diff
@@ -73,6 +73,7 @@ async function loadAddon(addonPath, verbose) {
   }
 
   const pluginClient = new PluginClient(packageName, {verbose: verbose});
+  pluginClient.on('unloaded', () => process.exit(0));
```
hobinjk: This should be guarded by `if (config.get('ipc.protocol') !== 'inproc')` so that it doesn't kill the Gateway process too.

PR #894, opened 2018-04-14
Fix schema bugs.

sogaani:
If an user input nothing on required field, we should set undefined in order to raise error.
User should able to select undefiend value on not required field.

### DIFF
--- a/static/js/addons-config/string-field.js
+++ b/static/js/addons-config/string-field.js
@@ -37,9 +37,15 @@ function StringField(schema,
   return this;
 }
 
+StringField.prototype.toFormData = function(value) {
+  return value === '' ? function() {}() : value;
+};
+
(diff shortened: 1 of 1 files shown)


static/js/addons-config/string-field.js
```diff
@@ -37,9 +37,15 @@ function StringField(schema,
   return this;
 }
 
+StringField.prototype.toFormData = function(value) {
+  return value === '' ? function() {}() : value;
```
mrstegeman: Why not just use `undefined` instead of the empty function? That would be much more clear.

sogaani: This deal with eslint rule "no-undefined". But I think that this is not clear, neither.
I will fix this.

PR #896, opened 2018-04-16
Return error when reject requestAction and removeAction

sogaani:
We should return 400 when add-on fail to create the action or remove the action.


src/plugin/adapter-proxy.js
```diff
@@ -56,9 +56,9 @@ class AdapterProxy extends Adapter {
     this.sendMsg(Constants.CANCEL_REMOVE_THING, {deviceId: device.id});
   }
 
-  sendMsg(methodType, data) {
+  sendMsg(methodType, data, defferd) {
```
mrstegeman: s/defferd/deferred/

src/constants.js
```diff
@@ -65,6 +65,8 @@ exports.REGISTER_PLUGIN_REPLY = 'registerPluginReply';
 exports.REMOVE_ACTION = 'removeAction';
 exports.REMOVE_THING = 'removeThing';
 exports.REQUEST_ACTION = 'requestAction';
+exports.METHOD_REJECTED = 'methodRejected';
```
mrstegeman: Since these are very specific to actions, can you make them something more appropriate, like `actionRequested` and `actionRejected`? You'll need to add another pair for removing actions.

src/plugin/plugin-client.js
```diff
@@ -87,11 +87,12 @@ class PluginClient extends EventEmitter {
     return this.deferredReply.promise;
   }
 
-  sendNotification(methodType, data) {
+  sendNotification(methodType, data, id) {
```
mrstegeman: I don't see a reason to add an additional field to every message.

src/plugin/plugin.js
```diff
@@ -226,20 +257,43 @@ class Plugin {
...
+  /**
+   * Generate an ID for a message.
+   *
+   * @returns {integer} An id.
+   */
+  generateMsgId() {
+    return ++this.nextId;
+  }
+
+  sendMsg(methodType, data, defferd) {
```
mrstegeman: s/defferd/deferred/

src/test/integration/actions-test.js
```diff
@@ -52,6 +83,20 @@ describe('actions/', function() {
     expect(err.response.status).toEqual(400);
   });
 
+  it('should fail when plugin reject requestAction', async () => {
```
mrstegeman: s/reject/rejects/

src/test/integration/actions-test.js
```diff
@@ -155,6 +200,34 @@ describe('actions/', function() {
     expect(err.response.status).toEqual(404);
   });
 
+  it('should fail when plugin reject removeAction', async () => {
```
mrstegeman: s/reject/rejects/

src/plugin/plugin-client.js
sogaani: If Id is undefined, ```JSON.stringify``` which is called on sendJson clears id.
Is this tricky?
Is it better to change like following?
```js
const json = {
       messageType: methodType,
       data: data,
};
if(typeof id !== 'undefined'){
  json.id = id;
}
```

mrstegeman: Yes, I would prefer not to rely on that.

In another comment, I indicated that I didn't see a need for the ID anyway, so maybe this just goes away. If not, you should probably just add it to `data`.

sogaani: OK. I will remove id tomorrow.

src/plugin/plugin.js
```diff
@@ -70,10 +74,67 @@ class Plugin {
 
   onMsg(msg) {
     DEBUG && console.log('Plugin: Rcvd Msg', msg);
+
+    // The first switch manages action method resolved or rejected messages.
+    switch (msg.messageType) {
+      case Constants.REQUEST_ACTION_RESOLVED: {
+        const actionId = msg.data.actionId;
+        const deffered = this.requestActionPromises.get(actionId);
```
mrstegeman: s/deffered/deferred/g

src/plugin/plugin.js
```diff
@@ -70,10 +74,67 @@ class Plugin {
...
+        if (typeof actionId === 'undefined' ||
+            typeof deffered === 'undefined') {
+          console.error('Plugin:', this.pluginId,
+                        'Unrecognized action id:', actionId,
+                        'Ignoring msg:', msg);
```
mrstegeman: Return after logging error.

src/plugin/plugin.js
```diff
@@ -70,10 +74,67 @@ class Plugin {
...
+        return;
+      }
+      case Constants.REQUEST_ACTION_REJECTED: {
+        const actionId = msg.data.actionId;
+        const deffered = this.requestActionPromises.get(actionId);
+        if (typeof actionId === 'undefined' ||
+            typeof deffered === 'undefined') {
+          console.error('Plugin:', this.pluginId,
+                        'Unrecognized action id:', actionId,
+                        'Ignoring msg:', msg);
```
mrstegeman: Return after logging error.

src/plugin/plugin.js
```diff
@@ -70,10 +74,67 @@ class Plugin {
...
+        return;
+      }
+      case Constants.REMOVE_ACTION_RESOLVED: {
+        const messageId = msg.data.messageId;
+        const deffered = this.removeActionPromises.get(messageId);
+        if (typeof messageId === 'undefined' ||
+            typeof deffered === 'undefined') {
+          console.error('Plugin:', this.pluginId,
+                        'Unrecognized message id:', messageId,
+                        'Ignoring msg:', msg);
```
mrstegeman: Return after logging error.

src/plugin/plugin.js
```diff
@@ -70,10 +74,67 @@ class Plugin {
...
+        return;
+      }
+      case Constants.REMOVE_ACTION_REJECTED: {
+        const messageId = msg.data.messageId;
+        const deffered = this.removeActionPromises.get(messageId);
+        if (typeof messageId === 'undefined' ||
+            typeof deffered === 'undefined') {
+          console.error('Plugin:', this.pluginId,
+                        'Unrecognized message id:', messageId,
+                        'Ignoring msg:', msg);
```
mrstegeman: Return after logging error.

src/plugin/plugin.js
```diff
@@ -226,20 +287,58 @@ class Plugin {
...
+   * @returns {integer} An id.
+   */
+  generateMsgId() {
+    return ++this.nextId;
+  }
+
+  sendMsg(methodType, data, deferred) {
     data.pluginId = this.pluginId;
+
+    // Methods which could fail should wait result.
```
mrstegeman: s/wait/await/

PR #900, opened 2018-04-17
Add udevadm script to allow serial port enumeration to work under OpenWRT

dhylands:


tools/udevadm
```diff
@@ -0,0 +1,57 @@
...
+  NAME=$(basename $1)
+  DEVICE_PATH=$(readlink -f /sys/class/tty/${NAME}/device)
+  if [ ! -e "${DEVICE_PATH}" ]; then
+    return
+  fi
+  echo "P: $(echo ${DEVICE_PATH} | cut -c5-)/tty/${NAME}"
+  echo "N: ${NAME}"
+  echo "E: DEVNAME=$1"
+  SUBSYSTEM=$(basename $(readlink -f ${DEVICE_PATH}/subsystem))
+  if [ "${SUBSYSTEM}" == "usb-serial" ]; then
```
mrstegeman: I ran this on the Pi. It didn't work unless I changed `==` to `=` (ditto on line 28). `==` isn't supported by all shells.

dhylands: Interesting. It seems dash (which is the default sh under ubuntu) doesn't support it, but ash (the shell used under openwrt) does.

I happen to have my sh set to bash on my dev machine, which is why I didn't notice it.

I'll change it anyways.

PR #901, opened 2018-04-17
Fix issue where addon migration would race with addon loading

hobinjk:
I'll also uplift this to the v0.4.0 but probably won't update the tag since the image didn't need this (and 0.4.1 will include this)

### DIFF
--- a/src/user-profile.js
+++ b/src/user-profile.js
@@ -50,20 +50,31 @@ const Profile = {
    * Manually copy, then remove, to prevent issues with cross-device renames.
    */
   renameDir: function(src, dst) {
-    ncp(src, dst, (e) => {
-      if (e) {
-        throw e;
-      }
-
-      rimraf(src, (err) => {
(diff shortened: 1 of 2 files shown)


src/user-profile.js
```diff
@@ -196,16 +202,24 @@ const Profile = {
...
           const lstat = fs.lstatSync(oldFname);
 
           if (fname !== 'plugin' && lstat.isDirectory()) {
             // Move existing add-ons.
-            this.renameDir(oldFname, newFname);
+            addonRenames.push(this.renameDir(oldFname, newFname));
           }
         }
+
+        Promise.all(addonRenames).then(() => {
```
mrstegeman: Couldn't there still be a race condition if AddonManager is in the middle of loading things when this hits?

hobinjk:
This appears to work, I'm going to test more with a full OTA but it appeared to handle migration properly

PR #923, opened 2018-04-21
Add ChoiceDetail type

steveatinfincia:
Greetings :)

As part of my work on [picamera-webthing](https://github.com/infincia/picamera-webthing) I've added several types to the Gateway.

This one, `ChoiceDetail`, is essentially a lightly styled HTML `select` element that allows the parent `Thing` to supply a list of options for it via the `choices` argument in the `ChoiceDetail` constructor.

It passes the selected choice to the `Thing` via the `Thing.setProperty(name, value)` call. This follows the example of the generic Detail types used by `UnknownThing`, allowing an arbitrary number of `ChoiceDetail` to be used by a `Thing` and preventing them from being closely coupled together.

The CSS here is just enough to avoid making the `select` element look out of place in a Detail bubble and may need some further thought on the best look & feel. You can see what it looks like in the `picamera-webthing` repo, which uses `ChoiceDetail` for the 'Resolution', 'Framerate', and 'Exposure'  properties.

## Other Gateway Changes

The choices should really be provided by the property description of the Web Thing itself, rather than hardcoded in the `Thing` in the Gateway, however that appears to require some changes in other parts of the Gateway to whitelist a new `choices` key, as it doesn't currently show up in the browser when adding a new Thing.

You can see an example of how this might work on the Web Thing side [here](https://github.com/infincia/picamera-webthing/blob/374eae1/picamera-webthing.py#L274), and an example of the Gateway side [here](https://github.com/infincia/gateway/blob/550a168/static/js/camera.js#L102).

The Web Thing declares a property of type `choice`, and provides a list of choices in the `choices` key. The Gateway side reacts to the presence of one or more `choice` properties, rather than expecting one to exist with a specific name, and then uses the provided `choices` list, if available.

Thanks :)

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -141,6 +141,30 @@
   background-image: url('../images/bulb-off.png');
 }
 
+
+/* Choice Detail */
+
+.choice-detail {
+
+}
+
+.choice-detail .choice-input {
+    max-width: 95%;
+    background-color: rgba(0,0,0,0);
+    color: white;
+    border: none;
+    font-size: 0.9em;
+    overflow: hidden;
+    max-width: 95%;
+    outline: none;
+    text-align: center;
+    text-align-last: center;
+}
+
+.choice-detail .choice-input option {
+    color: black; /* Needed to fix dropdown text color on Windows */
+}
+
 /* Color Light */
 
 .on-off-label, .color-light-label {
--- a/static/js/choice-detail.js
+++ b/static/js/choice-detail.js
@@ -1,0 +1,72 @@
+/**
+ * ChoiceDetail
+ *
+ * A bubble showing property choices that can be set on a thing
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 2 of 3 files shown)


benfrancis:
Hi Stephen,

Thanks for this contribution. Note that the `type` of a property in a Thing Description refers to data types as [defined](http://json-schema.org/latest/json-schema-core.html#rfc.section.4.2.1) in JSON Schema. That includes:
- null
- boolean
- object
- array
- number
- string

In the case of your `resolution` property it appears that would be a `string`. If you want to further constrain that property to a set of pre-defined values then I think the way to do that in JSON Schema is with an `enum`.

```
{
  ...
  "properties": {
    "resolution": {
      "type": "string",
      "enum": ["320x240",
               "640x480",
               "800x600",
               "1024x768",
               "1296x972",
               "1640x1232",
               "3280x2464"]
    }
  }
}
```

We don't currently support that feature of JSON Schema which is part of the JSON Schema Validation specification, so support would need to be added for that in order to create a choice detail like the one you'd like to add.

It would also be interesting to define a standard camera thing type or capability, to define which properties, actions and events a camera can expose.

mrstegeman:
Closed in favor of #1160.

PR #927, opened 2018-04-22
Ready to add action UI on thing detail screen.

sogaani:
I make following changes to get ready to add action UI.
* Add thingDetail screen to things which not supported.
* Delete duplicated codes.

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -83,10 +83,16 @@
 /* On/Off Switch */
 
 .on-off-switch .thing-icon {
+  border-radius: 6.4rem;
+  transform: translate(0);
+  background-color: #5d9bc7;
   background-image: url('../optimized-images/on-off-switch.png');
 }
 
 .on-off-switch.on .thing-icon {
+  border-radius: 6.4rem;
+  transform: translate(0);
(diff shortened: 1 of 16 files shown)


sogaani:
@mrstegeman 
Thanks notifying conflicts.
I rebased!

mrstegeman:
@sogaani This looks pretty good to me. The one major breakage I noticed was that certain thing types should not render the expanded UI icon in the things overview screen:
* `onOffSwitch`
* `onOffLight`
* `binarySensor`

Those three thing types only have one property, so there's no need to allow the user to open the expanded view. However, if `thing` (the unknown thing type) only has one property, we should still render the expander, since we only render a generic icon for that type.

sogaani:
@mrstegeman It was intended to display links, because I have thought the actions and events will be displayed on detail view. However, we still not have actions and events on detail view , and I delete links.
By your pointing out, I notice that we should check if `onOffSwitch` provides actions and events to display links.

mrstegeman:
@sogaani Those three types do not implement any actions or events, so there still won't be a need for the expanded view. Thanks for fixing!

Issue #793, opened 2018-03-22
OAuth: Change output of /things to reflect token's scope


hobinjk:
i.e. if the gateway has things test-1, test-2, and test-3 but a token grants `/things/test-1:read /things/test-2:read` a GET to /things authenticated by that token should list test-1 and test-2 instead of being blocked.

ziransun:
I'd like to work on this. Thanks!

mrstegeman:
Fixed by #929

PR #929, opened 2018-04-23
OAuth: Change output of /things to reflect token's scope. (#793)

ziransun:

### DIFF
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -26,10 +26,40 @@ const ThingsController = PromiseRouter();
  * Get a list of Things.
  */
 ThingsController.get('/', function(request, response) {
-  Things.getThingDescriptions(request.get('Host'), request.secure)
-    .then(function(things) {
-      response.status(200).json(things);
-    });
+  if (request.jwt.payload.role !== Constants.USER_TOKEN) {
+    if (!request.jwt.payload.scope) {
+      response.status(400).send('Token must contain scope');
+    } else {
+      const scope = request.jwt.payload.scope;
+      if (scope.indexOf(' ') === -1 && scope.indexOf('/') == 0 &&
+        scope.split('/').length == 2) {
+        Things.getThingDescriptions(request.get('Host'), request.secure)
+          .then(function(things) {
+            response.status(200).json(things);
+          });
+      } else {
+        // Get thingIds in scope
+        const paths = scope.split(' ');
+        const ids = new Array(0);
+        for (const path of paths) {
+          const parts = path.split(':');
+          const segments = parts[0].split('/');
+          ids.push(segments[segments.length - 1]);
+        }
+        Things.getListThingDescriptions(ids,
+                                        request.get('Host'),
+                                        request.secure)
+          .then(function(things) {
+            response.status(200).json(things);
+          });
+      }
+    }
+  } else {
+    Things.getThingDescriptions(request.get('Host'), request.secure)
+      .then(function(things) {
+        response.status(200).json(things);
+      });
+  }
 });
 
 /**
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -32,21 +32,21 @@ ThingsController.get('/', function(request, response) {
     } else {
       const scope = request.jwt.payload.scope;
       if (scope.indexOf(' ') === -1 && scope.indexOf('/') == 0 &&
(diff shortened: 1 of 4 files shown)


src/test/integration/oauth-test.js
```diff
@@ -348,12 +348,11 @@ describe('oauth/', function() {
       .set(...headerAuth(jwt));
     expect(res.status).toEqual(200);
 
-    // TODO(hobinjk): filter things instead of blocking #793
     let err = await pFinal(chai.request(server)
       .get(Constants.THINGS_PATH)
       .set('Accept', 'application/json')
       .set(...headerAuth(jwt)));
-    expect(err.response.status).toEqual(401);
+    expect(res.status).toEqual(200);
```
hobinjk: This test currently doesn't verify that the filtering is working as planned. It should be more like
```
res = chai.request(server)
  .get(Constants.THINGS_PATH)
  .set(...)
  .set(...);
expect(res.status).toEqual(200);
expect(res.body.length).toEqual(1);
expect(res.body[0].href).toEqual('/things/test-1');
```

src/controllers/things_controller.js
```diff
@@ -26,10 +26,40 @@ const ThingsController = PromiseRouter();
...
+          .then(function(things) {
+            response.status(200).json(things);
+          });
+      } else {
+        // Get thingIds in scope
+        const paths = scope.split(' ');
+        const ids = new Array(0);
+        for (const path of paths) {
+          const parts = path.split(':');
+          const segments = parts[0].split('/');
```
hobinjk: This makes the assumption that thing hrefs are always of the form `/thing/thing-id` which is at this point okay. However, it would be more WoT API compliant and future-compatible to do the filtering by the path itself using `thing.href`

src/jwt-middleware.js
```diff
@@ -73,7 +73,7 @@ function scopeAllowsRequest(scope, request) {
     const access = parts[1];
     const readwrite = access === Constants.READWRITE;
     path = parts[0];
-    if (requestPath.startsWith(path)) {
+    if (requestPath.startsWith(path) || path.startsWith(requestPath)) {
```
hobinjk: This incorrectly grants access to '/'. Instead of trying to generally accept prefixes of path it would be less error-prone to special-case instances where path starts with Constants.THINGS_PATH and requestPath is Constants.THINGS_PATH.

src/models/things.js
```diff
@@ -64,6 +64,42 @@ const Things = {
     });
   },
 
+  /**
+   * Get a list of Things by their IDs.
+   *
+   * {Array} ids IDs of the list of Things to get.
```
hobinjk: As above, it would be better for this to filter on hrefs since thing ids are not guaranteed to correspond to the href in the jwt's scope.

ziransun:
Thanks very much for the review comments. Code has been update. Please review when you have time.

src/controllers/things_controller.js
```diff
@@ -26,10 +26,39 @@ const ThingsController = PromiseRouter();
...
-    .then(function(things) {
-      response.status(200).json(things);
-    });
+  if (request.jwt.payload.role !== Constants.USER_TOKEN) {
+    if (!request.jwt.payload.scope) {
+      response.status(400).send('Token must contain scope');
+    } else {
+      const scope = request.jwt.payload.scope;
+      if (scope.indexOf(' ') === -1 && scope.indexOf('/') == 0 &&
+        scope.split('/').length == 2) {
```
hobinjk: This has a small bug because authorization codes have scope `/oauth:readwrite`. It would be better to make sure that scope.split(':')[0] is equal to Constants.THINGS_PATH.

src/test/integration/oauth-test.js
```diff
@@ -385,7 +386,7 @@ describe('oauth/', function() {
     expect(jwt).toBeTruthy();
 
     const err = await pFinal(chai.request(server)
-      .get(Constants.THINGS_PATH)
+      .get(`${Constants.THINGS_PATH}/${TEST_THING.id}`)
```
hobinjk: This test shouldn't need modification. An authorization code should not be able to access `/things`. The change I suggested above should fix this behavior.

ziransun:
Code updated (second time). Please review. Thanks!

hobinjk:
This looks great, thanks for the hard work!

PR #933, opened 2018-04-24
Run webpack/jest/eslint from node_modules.

mrstegeman:
Doing so eliminates the need for installing these things globally
on your system.


mrstegeman:
Ah, that works.

Issue #349, opened 2017-10-23
Test User Interface


hobinjk:
The UI is currently untested. Approaches this could take include the following:
- [ ] Manual testing checklist
- [ ] [Mochitest](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Mochitest)
- [ ] [Casper](http://casperjs.org/)/[Slimer](https://slimerjs.org)
- [ ] [Selenium](http://www.seleniumhq.org/)

hobinjk:
From initial investigation I'm pretty torn between Casper and Selenium. Selenium has a slight edge due to its long history and use of the WebDriver spec so I'm going to try writing a simple test using it.

### Nightmare
 - Promises, built on Phantom
 - https://segment.com/blog/ui-testing-with-nightmare/

### Phantom
 - Chrome headless

### Slimer
 - Phantom API with firefox
 - Supports promises

### Casper
 - Layer on top of Phantom and Slimer
 - http://casperjs.org/
 - A little weird, not natively promise-based
 - built-in testing support

### Puppeteer
 - Chrome headless, slightly different API
 - https://medium.com/@ankit_m/ui-testing-with-puppeteer-and-mocha-part-1-getting-started-b141b2f9e21
 - https://facebook.github.io/jest/docs/en/puppeteer.html
 - https://github.com/xfumihiro/jest-puppeteer-example

### Selenium/WebDriverJS
 - Pretty okay
 - https://blog.testproject.io/2017/06/07/javascript-testing-with-selenium-webdriver-mocha/

### Marionette
 - Python
 - Lots of firefox-specific testing functions like switching to chrome context

### Mochitest
 - Not really documented for use outside of firefox

### Firefox Puppeteer
 - Literally recommended against in the docs

sogaani:
If https://github.com/mozilla-iot/gateway/pull/812 is approval, we can use [jsdom](https://github.com/jsdom/jsdom).
I want to port tests for JSON SCHEMA form from [here](https://github.com/mozilla-services/react-jsonschema-form/tree/master/test).

hobinjk:
@sogaani I think for the thing-view it makes more sense to use the browser since it's less tightly coupled with the implementation. There are plans for us to refactor the thing views based on a capabilities model (as opposed to the current type-based model) which would significantly change the class structure but ideally not change the displayed UI.

The jsdom tests are very cool though and definitely something I like for testing components that are less likely to change dramatically.

sogaani:
[...]

PR #935, opened 2018-04-24
Add thing view tests with onOffSwitch.

sogaani:


sogaani:
close this pullrequest, because testing on browser is better solution. #349

PR #936, opened 2018-04-25
Add menu scrim #846

### LINKED ISSUE #846: Menu scrim
We should probably come up with a more general fix for hiding the main menu than #840 to prevent various strange interactions.

We can use a transparent "scrim" which covers the whole rest of the screen when the menu is displayed. Clicking on the scrim (i.e. clicking outside of the menu) closes the menu.

sogaani:
fix #846


mrstegeman:
Pushed up one small fix. Otherwise, looks nice!

PR #968, opened 2018-04-30
Fix issue where registration page was never displayed.

mrstegeman:
Since index.html was moved into static/ for webpack purposes,
RootController was never being hit, as index.html was being served
by the static file handler. Since RootController was never used,
the user was never being redirected to the domain registration
page.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -43,8 +43,16 @@ const Router = {
     });
 
     // First look for a static file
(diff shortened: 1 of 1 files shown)


mrstegeman:
@rzr You'll probably be interested in this fix, as I know you were seeing this issue quite a bit last week.

rzr:
Thx for the highlight I will check this out ;)

PR #971, opened 2018-04-30
OTA generate-release improvements

hobinjk:


hobinjk:
Wow, I can't believe my brain shut off long enough for me to commit that, thanks for the correction!

PR #980, opened 2018-05-03
Display invalid rules

hobinjk:
screen shot 2018-05-03 at 15 40 24
screen shot 2018-05-03 at 15 41 08

### DIFF
--- a/static/css/rule.css
+++ b/static/css/rule.css
@@ -24,6 +24,10 @@
   background: #46769C;
 }
 
+#rule-view > .title-bar.invalid {
+  background: #7f3939;
+}
+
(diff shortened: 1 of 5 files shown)


benfrancis:
I like it.

The only thing I'm wondering is whether it's worth adding the word "INVALID" somewhere to make it a bit more explicit what the red colour means. Maybe not in the edit rule view because shouting at you to tell you your rule is invalid while you're still editing it might be a bit jarring, but in the rules view I think it's worth making it more explicit. What do you think?

invalid_rule
(The text in the mockup is Open Sans, font-weight bold, white, 60% opacity).

PR #981, opened 2018-05-03
Add support for multi-effect rules

hobinjk:
squid


hobinjk:
Better image of full multi-effect experience:
screen shot 2018-05-03 at 15 43 57

hobinjk:
Notable drawback at this point is that there's no way to create/edit multi-effect rules on very small screens. I think it will require a bit of a redesign of some areas

Issue #909, opened 2018-04-19
Developer Settings


benfrancis:
As a developer I want to access information and functions of the gateway that an everyday user wouldn't use.

E.g. system logs, enable/disable SSH, Oauth development

benfrancis:
Developer section in Settings:

![settings](https://user-images.githubusercontent.com/552417/40841439-5ccd63a8-65a2-11e8-99cb-3dc5089ed829.png)

benfrancis:
Developer icon: https://drive.google.com/file/d/1ScB6dODZVwsXPe0nx24QjHD-jmeE3s0R/view?usp=sharing

Issue #217, opened 2017-07-12
Reboot button


benfrancis:
As a user I would like to be able to reboot the gateway from the UI so that I don't have to log in via the command line or just cut the power to restart it.

Possibly also a shut down.

mrstegeman:
The back-end is done for this, we just need to expose it to the UI. We have the back-end for both "restart gateway" and "reboot raspberry pi".

benfrancis:
I'm wondering where in the UI this button should go.

A couple of possibilities:
- Settings (but I'm not sure which section)
- Main menu (As part of combined logout/reboot/shutdown option)

antti202:
Main menu would be convenient (for me): 
Swap "Log out" to "(icon) Power" ->
- Logout

- Reboot

- Shutdown

gerritanne:
Yep, sounds good..

Kind regards,
Gerrit Haan

flatsiedatsie:
The `Power Settings` addon already provides this functionality.

That being said: would be great to have it built in.

Issue #216, opened 2017-07-12
Settings - Enable/disable SSH


benfrancis:
As a power user I would like to enable SSH on my gateway so that I can log in via the command line.

dhylands:
This should use the same technique that raspi-config uses.
https://github.com/asb/raspi-config/blob/3a5d75340a1f9fe5d7eebfb28fee0e24033f4fd3/raspi-config#L327-L345

codeadict:
Is someone actively working on this one?

benfrancis:
@codeadict I don't think so. People are just using the method of dropping a file called ssh onto the boot partition.

If you're interested in implementing this I can create a UI mockup if you like? I think it may require a new section in settings.

codeadict:
Yeah i'm interested, i can do it :). For now, what does the ssh file needs to have on it?

benfrancis:
Just an empty file, see https://github.com/mozilla-iot/wiki/wiki/Logging-into-the-Raspberry-Pi

codeadict:
Thanks!

dhylands:
Note that creating the ssh file will only enable ssh, it won't disable ssh. To disable ssh will require something like mentioned in comment 2 above.

rzr:
If you're building images you can use:

    ./make-prep.sh --ssh # (...)

dhylands:
Yes - but the images we ship should have ssh disabled (for security). This is to have a mechanism in the UI to enable/disable SSH.

benfrancis:
UX spec: https://docs.google.com/presentation/d/1rMeg6hep8JQflqwbiNIbhSoK9StCeVZ2npphmQHKda4/edit#slide=id.g3c095fbf0e_6_1

UI mockup:
![developer-settings](https://user-images.githubusercontent.com/552417/40841501-95eaad08-65a2-11e8-9625-cfcc75397b59.png)

mrstegeman:
@benfrancis Do we want this to be present outside of the RPi (or other platforms where we've implemented the required bits of code)? Should it be gone, or just disabled?

benfrancis:
SSH icon: https://drive.google.com/file/d/1fgtzqfcgJ40TxixkbnBXLP29wOOgJYcz/view?usp=sharing

benfrancis:
It would be good to be able to do feature detection and disable the options which aren't available on the current platform (e.g. disabling the text box, making "view logs" un-clickable). Can you think of a straightfoward way to do that?

[...]

PR #990, opened 2018-05-08
Implement back-end for platform settings.

mrstegeman:
References #216
References #217
References #909

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ $ jest src/test/{test-name}.js
   * **`log-timestamps.js`** - Utilities for adding timestamps to console logging functions
   * **`oauth-types.js`** - OAuth types
   * **`passwords.js`** - Password utilities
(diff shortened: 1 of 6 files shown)


jaller94:
@mrstegeman Has this feature been until in 0.5.2? I still see it in the code but I fail to find it in the web interface of my gateway.
Does it need to be activated?

mrstegeman:
@jaller94 This commit should be in all 0.5.X versions. It's accessed via _Settings->Developer_. Most things are only available on the Raspberry Pi for now.

mrstegeman:
Yep, that sound right. Are you looking for another setting in particular?

mrstegeman:
Nope, not yet. We haven't enabled that from the UI yet (#217). It can be done via the [API](https://github.com/mozilla-iot/gateway/blob/master/src/controllers/settings_controller.js#L393), though.

PR #992, opened 2018-05-09
Implement actions UI.

### LINKED ISSUE #907: Actions UI
As a user I want to request an action on a thing so I can modify its state or affect a physical change

mrstegeman:
Fixes #907

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -424,3 +424,99 @@
 #things a {
   text-decoration: none;
 }
+
+.action-button {
+  text-transform: uppercase;
+  height: 4rem;
+  width: 8rem;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.advanced-action-button {
+  position: relative;
+  top: 1.2rem;
+}
+
+.action-input-link:link,
+.action-input-link:visited,
+.action-input-link:hover,
+.action-input-link:active {
+  color: #fff;
+  position: relative;
+  top: 1rem;
+}
+
+.action-input {
+  max-width: 60rem;
+  margin: 0 auto;
+  text-align: left;
+  background-color: #5288af;
+  border-radius: 0.5rem;
+  padding: 3rem;
+}
+
+.action-input-title {
+  width: 100%;
+  text-align: center;
+  margin-top: -1rem;
+  margin-bottom: 2rem;
+  font-size: 2.8rem;
+}
+
+.action-input-form {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+  grid-gap: 1rem;
+  grid-auto-rows: minmax(4rem, auto);
+}
+
+.action-input-name {
+  margin-top: .4rem;
+  text-align: right;
+  font-size: 2rem;
+}
+
+.action-input-unit {
+  margin-top: .6rem;
+  text-align: left;
+}
+
+.action-input-string,
+.action-input-number {
+  width: 95%;
+  color: #333;
+}
+
+.action-input-checkbox {
+  display: none;
+}
+
+.action-input-checkbox + label:before {
+  content: "";
+  display: inline-block;
+  width  : 3rem;
+  height : 3rem;
+  margin : 0.1rem 0;
+  background: url("../optimized-images/checkbox-sprite.png") no-repeat 0 0;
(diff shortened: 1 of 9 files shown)


hobinjk:
With the single-thing example from webthing-node I'm not seeing an ActionDetail for the fade action

mrstegeman:
I only attached the action details to the `thing` type, since there are no actions defined in the spec for the other thing types. You can change the thing type for the single-thing example from `dimmableLight` to `thing` and it will work.

hobinjk:
Oh right, of course! Tested locally and LGTM

mrstegeman:
@benfrancis Would you mind checking this out? The latest virtual-things-adapter (v0.2.0) has a new thing with actions and events, so you can test with that.

benfrancis:
This looks great!

Seeing this in action, we could do with a bit more user feedback that an action request has been sent. We could start by adding an active state with a different background colour on the action button when you press it:

- `:disabled`: `#79a1bd`
- (normal): `#698ba4`
- `:hover`: `#59768a`
- `:active`: `#496071`

In addition to this, this wasn't in the spec, but I think the action input form should be dismissed when you press the button and it's submitted. The downside is that makes it harder to issue multiple action requests of the same type in a row, but it makes it much more obvious that the action request has been sent and if the action modifies properties you can immediately see the result. If the form just hangs around it looks like nothing happened.

We could go further than that and add a toast, in the same style as the event notifcations, to show that your action request has been submitted. What do you think? (Happy to do that part in a follow-up).

I wonder whether the touch target on the ellipsis (...) could be larger? It's hard to hit on a touchscreen. We could even make the whole bubble a hyperlink, as long as you can still press the button.

I realise the truncation of action names on buttons isn't ideal, I'll think about that.

Do we not have support for a "name" property of properties, actions and events to provide a human readable string (as opposed to the camelCase IDs)? Same for action inputs. If not I can file a follow-up for that (the W3C are currently proposing calling that "label" rather than "name" though.

One small nit: In the back end URL router the URLs are `actions/*` (plural), in the front end URL router they are `/action/*` (singular). Let's make them both plural for consistency.

As some point we may want to make some of this into web components.

mrstegeman:
> This looks great!
> 
> Seeing this in action, we could do with a bit more user feedback that an action request has been sent. We could start by adding an active state with a different background colour on the action button when you press it:
> 
>     :disabled: #79a1bd
>     (normal): #698ba4
>     :hover: #59768a
>     :active: #496071

Done.

> In addition to this, this wasn't in the spec, but I think the action input form should be dismissed when you press the button and it's submitted. The downside is that makes it harder to issue multiple action requests of the same type in a row, but it makes it much more obvious that the action request has been sent and if the action modifies properties you can immediately see the result. If the form just hangs around it looks like nothing happened.

Done.

> We could go further than that and add a toast, in the same style as the event notifcations, to show that your action request has been submitted. What do you think? (Happy to do that part in a follow-up).

Let's do that as a follow up, once I get the toast architecture figured out.

> I wonder whether the touch target on the ellipsis (...) could be larger? It's hard to hit on a touchscreen. We could even make the whole bubble a hyperlink, as long as you can still press the button.

Done.

> I realise the truncation of action names on buttons isn't ideal, I'll think about that.

Yeah, that was my biggest pain point.

> Do we not have support for a "name" property of properties, actions and events to provide a human readable string (as opposed to the camelCase IDs)? Same for action inputs. If not I can file a follow-up for that (the W3C are currently proposing calling that "label" rather than "name" though.

`"name"` is only ever used in the spec for _Thing_.

> One small nit: In the back end URL router the URLs are actions/* (plural), in the front end URL router they are /action/* (singular). Let's make them both plural for consistency.

Done.

> As some point we may want to make some of this into web components.

Definitely.

benfrancis:
> "name" is only ever used in the spec for Thing.

Huh, you're right. I must be getting confused between specifications again. Seems like something I should file an issue for to address.

One last thing, the touch target on the ellipsis is bigger but now appears to be obscuring half of the button so you can accidentally click the ellipsis rather than the button. The touch target should either butt up against the bottom of the button, or extend across the whole bubble with the button as an unobscured island of touch target in the middle! (if that's possible...).

mrstegeman:
@benfrancis Good catch! I left out 1 CSS rule.

Issue #908, opened 2018-04-19
Events UI


benfrancis:
As a user I want to be notified of an event which occurred on a thing so that I can act on that information

benfrancis:
UX spec: https://docs.google.com/presentation/d/1rMeg6hep8JQflqwbiNIbhSoK9StCeVZ2npphmQHKda4/edit#slide=id.g38cd194295_0_33

PR #998, opened 2018-05-10
Add support for overflow menu and message area.

mrstegeman:
These will both be required for #908, but are useful on their own
as well.


hobinjk:
Mostly LGTM. There's a slight visual bug if the message area is empty (but I'm guessing that if the overflow button is visible then there's stuff in the message area).

screen shot 2018-05-11 at 6 29 10 am

mrstegeman:
Yeah, the button is completely useless with no menu items, so this should never happen. Also, that's the "overflow menu". The message area comes up in the bottom center.

PR #1000, opened 2018-05-11
Implement events UI

### LINKED ISSUE #908: Events UI

mrstegeman:
Fixes #908

### DIFF
--- a/static/js/event-list.js
+++ b/static/js/event-list.js
@@ -1,0 +1,190 @@
+/**
+ * EventList.
+ *
+ * Represents a list of events for a thing.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+'use strict';
+
+const API = require('./api');
+const App = require('./app');
+const Utils = require('./utils');
+
+const EventList = function(description) {
+  this.limit = 100;
+  this.href = new URL(description.href, App.ORIGIN);
+  this.wsHref = this.href.href.replace(/^http/, 'ws');
+
+  // Build up the event schema now. This is used later when building list items.
+  this.schema = {};
+  for (const eventName in description.events) {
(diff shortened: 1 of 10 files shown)


static/js/event-list.js
```diff
@@ -0,0 +1,190 @@
...
+
+    body += '<br>';
+  }
+
+  const el = document.createElement('div');
+
+  const item = `<li class="event-item">
+    <div class="event-header">
+      <span class="event-name">${Utils.escapeHtml(event.name)}</span>
+      <span class="event-time" original="${event.timestamp.toISOString()}"
```
hobinjk: It would be more standard to use 'data-original' here and el.dataset.original later

static/js/event-list.js
```diff
@@ -0,0 +1,190 @@
...
+            title="${event.timestamp.toLocaleString()}">
+        ${Utils.fuzzyTime(event.timestamp)}
+      </span>
+    </div>
+    <div class="event-body">${body.substring(0, body.length - 4)}</div>
```
hobinjk: A comment about why this is -4 would be nice

static/js/things.js
```diff
@@ -300,6 +309,73 @@ const ThingsScreen = {
...
+    };
+
+    // Fetch a thing from the server
+    fetch(`/things/${encodeURIComponent(thingId)}`, opts).then((response) => {
+      if (response.status == 404) {
+        this.thingsElement.innerHTML = this.THING_NOT_FOUND_MESSAGE;
+        return;
+      }
+
+      response.json().then((description) => {
```
hobinjk: nit: this could be
```
  if (response.status == 404) {
    return Promise.resolve();
  }

  return response.json();
}).then((description) => {
```
to remove an unnecessary level of indentation and the duplication of `this.thingsElement.innerHTML = this.THING_NOT_FOUND_MESSAGE;`

mrstegeman:
@hobinjk Thanks for the pointers! Everything's been addressed.

PR #1004, opened 2018-05-13
Fix certbot configs after moving them.

### LINKED ISSUE #1003: Recent update (0.4.0?) broke letsencrypt renewal, with "CertStorageError: expected /home/pi/mozilla-iot/etc/live/eagle.mozilla-iot.org/cert.pem to be a symlink"
My gateway is currently stuck with an expired Let's Encrypt cert (expired May 7) and it's unable to get a new one. This renders its web portal unusable in Firefox for Android (which AFAICT does not expose a way for users to click through its expired-cert SSL error page).

My gateway started at version 0.3.0 when I originally set it up, and (after receiving OTA updates) it's currently at version 0.4.0, which it received on April 18th. I'm guessing it's been unable to get a new cert since that update.

I suspect this has to do with the "Everything now lives in ~/.mozilla-iot" change [listed in the 0.4.0 release notes](https://github.com/mozilla-iot/gateway/releases/tag/0.4.0). CC @dhylands

mrstegeman:
Fixes #1003

### DIFF
--- a/tools/renew-certificates.sh
+++ b/tools/renew-certificates.sh
@@ -2,6 +2,11 @@
 
 MOZIOT_HOME="${MOZIOT_HOME:=/home/pi/.mozilla-iot}"
 
+for file in $(find "${MOZIOT_HOME}/etc/renewal" -type f -name '*.conf'); do
+    sed -i -e 's_/mozilla-iot/etc_/.mozilla-iot/etc_g' "${file}"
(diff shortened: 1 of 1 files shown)


benfrancis:
Does this warrant a bugfix release to prevent peoples' certificates from expiring?

mrstegeman:
@benfrancis Probably. I'm sure there are some other bug fixes that would be nice to have.

PR #1012, opened 2018-05-17
Remove sleep from rule engine tests

sogaani:
I failed rule-engin tests sometimes, because using fixed time sleep to wait async process.
This PR replace `sleep` with `waitForExpect ` to observe conditions multitimes.

### DIFF
--- a/src/test/rules-engine/index-test.js
+++ b/src/test/rules-engine/index-test.js
@@ -1,5 +1,6 @@
 const {server, chai, mockAdapter} = require('../common');
 const Settings = require('../../models/settings');
+const {waitForExpect} = require('../expect-utils');
 
 const pFinal = require('../promise-final');
 const {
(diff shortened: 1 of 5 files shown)


src/test/rules-engine/index-test.js
```diff
@@ -536,23 +540,23 @@ describe('rules engine', function() {
...
-      .get(`${Constants.THINGS_PATH}/${thingLight1.id
-      }${Constants.ACTIONS_PATH}`)
-      .set('Accept', 'application/json')
-      .set(...headerAuth(jwt));
-    expect(res.status).toEqual(200);
-    expect(Array.isArray(res.body)).toBeTruthy();
-    expect(res.body.length).toEqual(1);
-    expect(res.body[0]).toHaveProperty('blink');
+    await waitForExpect(async () => {
+      Events.add(new Event('surge',
```
hobinjk: The adding the event should be outside of the waitForExpect so that the test only adds one event.

sogaani: Thanks for the review. Since responding the rule-engin API dose not means that rule can work, I change that the rule-engin waits response until connecting websocket.

hobinjk:
Thanks! This now just needs the conflicts resolved

Issue #928, opened 2018-04-23
Cannot connect to gateway.local or ip address


TonyCanning:
Hi all,
I've followed this:
https://hacks.mozilla.org/2018/02/how-to-build-your-own-private-smart-home-with-a-raspberry-pi-and-mozillas-things-gateway/
and gotten to 
https://hacks.mozilla.org/files/2018/02/wifi_connecting.png
![image](https://user-images.githubusercontent.com/10809745/39128290-556708ce-46ff-11e8-9e4a-d938a1a7ab60.png)

when I navigate to gateway.local I get:
![image](https://user-images.githubusercontent.com/10809745/39128350-767b2018-46ff-11e8-8370-9cd7b76113f3.png)

So I put the pi on screen and checked for the ip address and tried connecting to it.
Same as above.

I haven't done anything other than the instructions (haven't changed gateway name or anything).

Anyone help?

rzr:
I think I faced that issue couple of weeks ago, but I've been reported it has been fixed on maser

So is this reproductible on master  ? or 0.4.0 ? 

you can try by adding :8080 port

mrstegeman:
Fixed by #1020

PR #1018, opened 2018-05-18
package: Add le-acme-core module

rzr:
To prevent observed logged failure on
 "Choose a secure web address for your gateway:" page:

  Error: Cannot find module 'le-acme-core'

Bug: https://github.com/mozilla-iot/gateway/issues/928
Signed-off-by: Philippe Coval <rzr@users.sf.net>

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "html-loader": "^0.5.5",
     "html-webpack-plugin": "^3.2.0",
     "jsonwebtoken": "^8.1.0",
+    "le-acme-core": "^2.1.3",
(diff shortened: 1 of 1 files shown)


mrstegeman:
@rzr This should be unnecessary. If you just run `yarn`, it should pull in this dependency (it did for me).

rzr:
installer script is calling "npm install" not yarn so I faced this issue

mrstegeman:
@rzr It turns out `npm install` gave you a newer version of `greenlock` than what `yarn` was giving us.

mrstegeman:
The complete fix is here: #1020

PR #1021, opened 2018-05-20
Make schema-form resusable.

sogaani:
The things description has some field described by JSON-schema, e.g. action input/output, event output.
Maybe, JSON-schema form which used by add-on config is usable for display these descriptions.


mrstegeman:
@sogaani Would you mind fixing the conflict?

PR #1022, opened 2018-05-20
The properties on thing-view reflects the value which PUT property API returns

### LINKED ISSUE #972: UI seems to ignore propertyChanged for read-only properties
With a zwave temperature sensor, the value is read-only. The UI lets you try to enter a new value, and the zwave adapter now rejects the promise and sends a propertyChanged notification for the original temperature.

The UI, however, shows the temperature that the user entered and doesn't reflect the propertyChanged notification from the adapter.

sogaani:
Fixes #972 

* The `propertyChanged` message is sometimes not reached UI, because creating websocket connection is asynchronous with displaying thing-detail-view.
* Even if `notifyPropertyChanged` is called while reject `setValue`, the `propertyChanged` message sometimes reaches UI  earlier than `PUT property API` response.

### DIFF
--- a/src/addons-test/mock-adapter/mock-adapter.js
+++ b/src/addons-test/mock-adapter/mock-adapter.js
@@ -15,6 +15,9 @@ class MockProperty extends Property {
     super(device, name, propertyDescription);
     this.unit = propertyDescription.unit;
     this.description = propertyDescription.description;
+    this.readOnly = propertyDescription.hasOwnProperty('readOnly') ?
+      propertyDescription.readOnly :
+      false;
     this.setCachedValue(propertyDescription.value);
     this.device.notifyPropertyChanged(this);
   }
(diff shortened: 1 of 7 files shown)


mrstegeman:
@sogaani I'm hesitant to merge this until we decide on our official solution to https://github.com/mozilla-iot/wot/issues/38

sogaani:
@mrstegeman mozilla-iot/wot#38 seems to discuss about read only property, but this PR fixes about UI input form value when `setProperty` rejected by some reason. Is it same?

mrstegeman:
My concern was that you're adding a `readOnly` property to the thing description, which is not presently in the spec.

sogaani:
@mrstegeman I understood your concern. I agree with avoiding add a `readOnly` property. I used `readOnly` with `mock-adapter` in order to make a property which rejects `setProperty`. So, I can change it to use property name like `rejectProperty` or make other adapters for tests. Could this change resolve your concern?

mrstegeman:
@sogaani Yes, I'd prefer the property name change, similar to what you did for actions a while back. Also, you have a conflict 😄

PR #1027, opened 2018-05-22
Fix failing to display available addon list

sogaani:
When access `/settings/addons` and click `+` button to access `/settings/addons/discovered` as soon as possible, the Gateway does not display any addons.

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -49,6 +49,7 @@ const SettingsScreen = {
 
     this.availableAddons = new Map();
     this.installedAddons = new Map();
+    this.fetchAddonDeferred = null;
   },
 
   show: function(section, subsection, id) {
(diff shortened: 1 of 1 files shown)


static/js/settings.js
```diff
@@ -330,6 +331,32 @@ const SettingsScreen = {
...
+    if (this.fetchAddonDeferred) {
+      return this.fetchAddonDeferred.catch(() => {
+        this.fetchAddonDeferred = null;
+        return this.getAddonList();
+      });
+    }
+
+    let resolveFunc = null, rejectFunc = null;
+
+    this.fetchAddonDeferred = new Promise((resolve, reject) => {
```
mrstegeman: This would be more clear:

```javascript
    this.fetchAddonDeferred = new Promise((resolve, reject) => {
      this.fetchAddonList().then(() => {
        resolve();
      }).catch((e) => {
        console.error(`Failed to parse add-ons list: ${e}`);
        reject(e);
      });
    });
```

sogaani: I get the point and addressed.

PR #1033, opened 2018-05-25
Cache static files on service worker.

sogaani:


hobinjk:
Excellent! This approach is much better than my original "cache everything then eventually update in the background" strategy. I'll test it out in more detail tomorrow. The one thing I'd recommend is to bump the cache identifier to mozilla-iot-0.0.2 since this is a significant change in function from the old style.

sogaani:
@hobinjk Oh, bump cache version is important point! Probably, we should bump version per build. I will update cache version while building webpack. Furthermore, we should delete old cache when update cache version.

hobinjk:
Great! Looks good to me

PR #1034, opened 2018-05-25
Support custom Thing UIs.

### LINKED ISSUE #910: Custom Thing UI
As a thing developer I want to provide a custom web interface for my webthing so that I can provide a better user experience than the automatically generated default UI

mrstegeman:
Depends on mozilla-iot/gateway-addon-node#15
Depends on mozilla-iot/gateway-addon-python#16

Fixes #910

### DIFF
--- a/src/models/thing.js
+++ b/src/models/thing.js
@@ -64,13 +64,28 @@ const Thing = function(id, description) {
       rel: 'events',
       href: `${this.href}/events`,
     },
-    {
-      rel: 'alternate',
-      mediaType: 'text/html',
-      href: this.href,
-    },
-    // TODO: add websocket URL
   ];
+
+  const uiLink = {
+    rel: 'alternate',
+    mediaType: 'text/html',
+    href: this.href,
+  };
+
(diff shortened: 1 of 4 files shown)


hobinjk:
I've been unable to add a thing from thing-url-adapter@0.1.11 which retains the uiHref of a thing from webthing-node. I'm setting thing.uiHref and thing type in webthing-node's example and localhost:8888 has the correct ui href. However, once I add this to the Gateway it shows up as ui href /things/path-to-thing. Is there something I should fix with this setup?

mrstegeman:
Have you run `yarn` since I merged in the change to gateway-addon-node?

hobinjk:
Oh right! I had, but gateway-addon hasn't had a version bump since the uiHref commit so yarn resolved it to the original 0.4.0

mrstegeman:
@hobinjk Yeah, I didn't bump the version but I did update `yarn.lock`.

hobinjk:
Even running `rm -r node_modules; yarn` I'm still getting the old 0.4.0, I think yarn is refusing for some weird cache-based reason.

PR #1039, opened 2018-05-29
Allow user to rename thing.

### LINKED ISSUE #874: Rename discovered thing
use-case: I have discovered light bulbs that I move around the house. Changing the name involves currently deleting and re-adding to change the name in the web front end.

Have an update to name so movement of devices and name changes is easier.

mrstegeman:
Fixes #874


src/test/integration/things-test.js
```diff
@@ -148,6 +148,37 @@ describe('things/', function() {
     expect(err.response.status).toEqual(404);
   });
 
+  it('rename a thing', async () => {
```
hobinjk: It would be nice to also test failing to rename a thing (i.e. missing name and optionally name = '')

src/controllers/things_controller.js
```diff
@@ -223,15 +223,34 @@ ThingsController.patch('/:thingId', function(request, response) {
...
+ */
+ThingsController.put('/:thingId', (request, response) => {
+  const thingId = request.params.thingId;
+  if (!request.body || !request.body.hasOwnProperty('name')) {
+    response.status(400).send('name parameter required');
+    return;
+  }
+
+  Things.getThing(thingId).then((thing) => {
+    return thing.setName(request.body.name);
```
hobinjk: While prohibited in the UI, I can manually make a request through curl which sets the thing's name to '' or ' ' which isn't that big a deal but would be nice to prohibit.

mrstegeman:
@hobinjk Updated

PR #1044, opened 2018-05-30
Add EqualityTrigger

hobinjk:


hobinjk:
@mrstegeman That behavior should be fixed now

mrstegeman:
@hobinjk Same behavior 😦

hobinjk:
Now with a fix for the exact behavior and not a closely-related one

PR #1045, opened 2018-05-30
Handle user entering PIN.

### LINKED ISSUE #668: Need a mechanism to input a PIN during pairing
For things like Bluetooth or HomeKit, the user will need to input a device-specific PIN code while pairing with a given device.

mrstegeman:
Depends on mozilla-iot/gateway-addon-python#19
Depends on mozilla-iot/gateway-addon-node#16
Fixes #668


src/controllers/things_controller.js
```diff
@@ -63,6 +63,28 @@ ThingsController.get('/', (request, response) => {
   }
 });
 
+ThingsController.patch('/', async (request, response) => {
```
hobinjk: Why is thingId part of the body instead of part of the URL?

mrstegeman: Because the thing hasn't actually been added to the gateway yet. It's in the weird limbo state between "pairing" and "saving".

mrstegeman:
@hobinjk Yeah... The question is what to do. We have the error area that we could populate with a hint or something, but currently, all the UI has available is a regex. We could allow the adapters to also send a hint as to what it's looking for.

@benfrancis Thoughts?

benfrancis:
I assume by "invalid" you mean "invalid format" as opposed to "incorrect pin" for which the mockups already have an error.

I think disabling the button is OK. If the UI knows the expected format then you could also use [constraint validation](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation) and use the :invalid CSS selector on the text input (the default style for that might be OK).

mrstegeman:
@benfrancis I like that. Updated to use `setCustomValidity()` after testing the adapter-provided regex.

PR #1049, opened 2018-06-01
Tweak domain settings UI to match mockups

benfrancis:
Some slight UI tweaks to match the UX specification at https://docs.google.com/presentation/d/1rMeg6hep8JQflqwbiNIbhSoK9StCeVZ2npphmQHKda4/edit#slide=id.g3c095fbf0e_0_0

There's one remaining issue which I'll file a follow-up for.

Also changes default mDNS host name back to gateway.local to match all our documentation and blog posts.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -43,7 +43,7 @@ module.exports = {
         localControl: {
           mdnsServiceType: "http",
           mdnsServiceName: "Mozilla IoT Gateway",
-          mdnsServiceDomain: "gateway",
+          mdnsServiceDomain: "mygateway",
 
           mdnsTxt: {
             desc: "Web of Things gateway",
--- a/config/default.js
+++ b/config/default.js
@@ -43,10 +43,10 @@ module.exports = {
         localControl: {
           mdnsServiceType: "http",
           mdnsServiceName: "Mozilla IoT Gateway",
-          mdnsServiceDomain: "MozillaIoTGateway",
+          mdnsServiceDomain: "gateway",
 
           mdnsTxt: {
-            desc: "A local IoT Gateway",
+            desc: "Web of Things gateway",
             protocol: "http, https, Web Sockets",
             power: "6 watts",
           },
--- a/static/css/settings.css
+++ b/static/css/settings.css
@@ -201,7 +201,7 @@
   display: none;
 }
 
-.experiment-item, .update-item, .authorization-item, .domain-item {
+.experiment-item, .update-item, .authorization-item {
   display: block;
   margin: 1rem 0;
   font-size: 1.8rem;
@@ -219,7 +219,8 @@
 .adapter-item,
 .addon-item,
 .authorization-item,
-.discovered-addon-item {
+.discovered-addon-item,
+.domain-item {
   display: block;
   font-size: 1.8rem;
   list-style-type: none;
@@ -231,8 +232,13 @@
 .user-item,
 .adapter-item,
 .addon-item,
-.authorization-item {
+.authorization-item,
+.domain-item {
   background-color: #5288af;
+}
+
+.domain-item {
+  padding: 2rem;
 }
 
 .discovered-addon-item {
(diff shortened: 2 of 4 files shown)


mrstegeman:
@benfrancis There is a small problem with setting the default back to `gateway`. `avahi` in the Raspbian image already takes care of that for us, so setting this to `gateway` actually produces `gateway (2).local`.

benfrancis:
What do you mean by "complimentary mNDS domain"?

Wait, so changing gateway.local to something else or disabling "local access" actually has no effect on the gateway broadcasting itself as gateway.local? That sounds broken. I assume one use case for this feature would be if you have multiple domains on the same network they can have different hosts, but currently it sounds like they all broadcast gateway.local regardless.

Is there anything we can do to fix this?

mrstegeman:
@benfrancis Yes, that's correct. There was a bit of discussion about this in https://github.com/mozilla-iot/gateway/issues/82#issuecomment-381177821

I think where this feature really comes in handy is outside of our Raspbian image. This allows any user, regardless of OS or privileges to broadcast an mDNS service. For instance, if the gateway is running as a user without `sudo` privileges, or on Windows, then it is impossible for the gateway to edit `/etc/hostname` and restart the `avahi` service.

That's why I said that this is really complementary to what the OS may or may not already be doing.

nherriot:
@benfrancis  - i think that you don't need to bother with avahi now. so any shell scripts etc that are setting that up can disappear, as the Mozilla platform handles it now. (or someone please correct me if I've misunderstood?). So this should simply the code base and dependencies on lots of config and scripts on startup.

mrstegeman:
@nherriot `avahi` is already part of the Raspbian image, it's nothing that we've configured ourself. So we'd actually have to remove it. I don't see any problem with letting them coexist.

benfrancis:
@mrstegeman OK, I see. Sorry I didn't pick up on this earlier.

The problem with them co-existing is that I think this is actively confusing for users. When I go to Domain Settings and view the current host name and mDNS broadcast setting, my assumption is that reflects the current state of the gateway.

Currently it appears that (in the Raspbian image, which is the closest thing we currently have to "production") the default is that the gateway broadcasts "gateway.local", but that Domain Settings may show a different host name and/or may show mDNS as disabled when it is actually still broadcasting.

The only reason I can think of for changing the .local host name (which BTW is not what was originally intended by "custom domain" in #82) is if you want two different gateways on the same local network broadcasting different mDNS hosts. Currently two (Raspbian-based) gateways will still broadcast gateway.local and conflict with each other, regardless of what the user configures in the UI.

For the use case of disabling the mDNS broadcast (e.g. if a gateway is being connected to an untrusted local network and you don't want it to broadcast itself) it will appear that the mDNS broadcast is disabled, but it is in fact still broadcasting as gateway.local.

@nherriot The problem I can see with disabling Avahi is that a main use case for the gateway.local broadcast is for first time setup. There are some stages of first time setup (e.g. WiFi setup) during which the main gateway process is not running. Disabling this would break some of the user documentation we currently have around first time setup (e.g. when using Ethernet rather than WiFi).

If we're going to have a feature to enable/disable and change the host name of the mDNS broadcast, I think it needs to be the *only* mDNS broadcast.

I'm going to remove that change from this pull request so we can land the UI tweaks, but I think this issue needs addressing before we ship this feature.

mrstegeman:
@benfrancis Can you open a separate issue for this so we can discuss further?

benfrancis:
See #1055

nherriot:
Mmmm. I've got quite a few raspberry pi's and I'm using rasbian stretch at the moment. None of them start and broadcast on gateway.local. They start on: 
  raspberrypi.local
What I'm suggesting is that whatever the Mozilla Gateway did to change that could be left as is (i.e let the system come up as raspberrry.pi). So the raspberry pi keeps it's own mDNS name.
The user gets the flexibility of choosing the name of the gateway locally in the Mozilla settings. I thought changing doc's are easy compared to code. :-)

If you want it to be the only mDNS broadcast then you will have to stop the avahi daemon broadcasting, which i think is a cool way to go if you are wanting to make this platform a nodejs top to bottom. However - you may upset developers who normally want to do:
  /> ssh pi@raspberry.local
That's a tough one!

Or you could have the nodejs app configure the daemon - but I thought the conversation on that topic was bad idea as the gateway needed root privileges ?

PR #1061, opened 2018-06-04
Add title and icon to settings screens.

mrstegeman:
These were in the mock-ups but were never implemented.


static/index.html
```diff
@@ -167,18 +176,18 @@
       </section>
       <section id="experiment-settings" class="hidden settings-section">
         <ul>
-          <li id="floorplan-experiment-item" class="experiment-item">
-            <input id="floorplan-experiment-checkbox" class="experiment-checkbox" type="checkbox">
-            <label for="floorplan-experiment-checkbox">Floorplan</label>
+          <li id="rules-experiment-item" class="experiment-item">
+            <input id="rules-experiment-checkbox" class="experiment-checkbox" type="checkbox">
+            <label for="rules-experiment-checkbox">Rules Engine</label>
```
hobinjk: These should still be called Rules and Floorplan as far as I know

mrstegeman: Yeah, I was unsure about whether or not to update these. The UI mock-ups had these names, so I changed them.

@benfrancis Which shall we go with?

benfrancis: I don't mind too much, but "Rules Engine" and "Floorplan UI" are probably a bit more descriptive.

To be honest we could probably just remove these two and remove the underlying settings as well for 0.5. I think our intention was to keep these two as options for one release after having these features turned on by default.

"Rules Engine" and "Floorplan UI" are fine, but if we have time we should just remove them. "Voice controls" should also change to "Smart Assistant" once the smart assistant UI is implemented.

mrstegeman: Ok, I'll merge as is and remove the settings shortly.

PR #1062, opened 2018-06-04
Support more complicated json schema on actions input.

sogaani:


mrstegeman:
Before I review this, let's make sure this is something we want to support.

@benfrancis Thoughts here? Do we want to support a large swath of JSON Schema for action inputs, or are we trying to keep it simple?

sogaani:
@mrstegeman OK. I just need enum schema for my addon. If keep input simple, I want to create another PR to support enum schema.

benfrancis:
Currently our specification just says an [action](https://iot.mozilla.org/wot/#action-object) input is defined by a JSON Schema, but doesn't go into details.

The W3C Thing Description specification (which we're trying to converge with) uses a *subset* of JSON Schema, so it's possible that not all features will be supported. The current draft [says](https://w3c.github.io/wot-thing-description/#dataschema) `type` (e.g. for an action input) can be one of "boolean, integer, number, string, object, array, null". enum could be a reasonable addition to that, but it isn't currently specified.

I think currently in our UI implementation we only support objects that go one level deep. Supporting nested objects and arrays could get quite complex for example!

@sogaani To help with reviewing this, can you list the JSON Schema features you want to add support for, and how they would appear in the UI?

Note that I'm not entirely satisfied with the current action UI we already have (which I designed!), as I think think the UX for providing action inputs is already a bit too complicated. It can probably be improved.

sogaani:
I want to add support 'type string' with enum options.

```
            selectInput: {
              type: 'string',
              enum: [
                'option #0',
                'option #1',
                'option #2',
                'option #3',
              ],
            },
```
image

Over all input schema.
```
        input: {
          type: 'object',
          required: [
            'numberInput',
          ],
          properties: {
            numberInput: {
              type: 'number',
              minimum: 0,
              maximum: 100,
              unit: 'percent',
            },
            integerInput: {
              type: 'integer',
              unit: 'meters',
            },
            stringInput: {
              type: 'string',
            },
            booleanInput: {
              type: 'boolean',
            },
            selectInput: {
              type: 'string',
              enum: [
                'option #0',
                'option #1',
                'option #2',
                'option #3',
              ],
            },
          },
        },
```



> The W3C Thing Description specification (which we're trying to converge with) uses a subset of JSON Schema, so it's possible that not all features will be supported. The current draft says type (e.g. for an action input) can be one of "boolean, integer, number, string, object, array, null". enum could be a reasonable addition to that, but it isn't currently specified.

I get it and be sure this PR definitely over spec, this PR add support any nested schema and other complicated options, e.g. [`oneOf`](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html), `ref`. But this is not useful, since we have no use case of these complicated schema.

benfrancis:
I think adding the enum seems completely reasonable and the UI representation makes sense.

As you suggest, I would hesitate to add any of the other more complex JSON Schema features until we have identified use cases and determined whether these will actually be supported by the W3C specification.

Perhaps we can just land the enum support for now.

sogaani:
@benfrancis I will make another PR which support the enum, or should I wait reflecting enum to wot draft?

sogaani:
> Perhaps we can just land the enum support for now.

I missed comment.
I will create another PR!

PR #1063, opened 2018-06-04
Update gateway version to match release.

mrstegeman:


hobinjk:
This may as well go to 0.5.0 since master is where we'll be cutting that release from whereas 0.4.1 is on its own weird little timeline

mrstegeman:
@hobinjk Fair enough. Changed.

PR #1089, opened 2018-06-06
Use replace instead of set to not mess with browser's back button

### LINKED ISSUE #1072: can't use back button on login page
Go to the gateway.local login page using an existing tab with some browsing history.  Try to use the back button.  The result is always that the login page reloads.

hobinjk:
Fix #1072


benfrancis:
Hah, thanks for this James, so simple!

PR #1091, opened 2018-06-06
Rules MultiTrigger and Pulse/Set Support

hobinjk:
Only creates OR triggers without a UI to change to AND (and other advanced options like set/pulse).

### DIFF
--- a/src/rules-engine/triggers/MultiTrigger.js
+++ b/src/rules-engine/triggers/MultiTrigger.js
@@ -1,0 +1,81 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.*
+ */
+
+const assert = require('assert');
+const Events = require('../Events');
+const Trigger = require('./Trigger');
+
+const ops = {
+  AND: 'AND',
+  OR: 'OR',
+};
+
+/**
+ * A Trigger which activates only when a set of triggers are activated
+ */
+class MultiTrigger extends Trigger {
+  /**
+   * @param {TriggerDescription} desc
+   */
+  constructor(desc) {
+    super(desc);
+    assert(desc.op in ops);
+    this.op = desc.op;
+    const fromDescription = require('./index').fromDescription;
+
+    this.triggers = desc.triggers.map((trigger) => {
+      return fromDescription(trigger);
+    });
+
+    this.states = new Array(this.triggers.length);
+    this.state = false;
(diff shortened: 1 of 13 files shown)


hobinjk:
This is in a mostly viable state. I'm interested to see what other people think about the drop-down menu styling:
screen shot 2018-06-22 at 14 54 19

hobinjk:
(hilariously enough the baseline of the normal text is aligned with that of the drop-down text but there's an optical illusion going on)

hobinjk:
(forgot to remove WIP from title)

benfrancis:
Text gets a bit crowded, can we make this fit the available space a bit better, maybe by moving the title upwards or extending the box downwards?

screenshot from 2018-06-29 15-36-36

benfrancis:
I think I found a bug. If you drag a device from the left side to the right side, it doesn't remove the "or" statement from the text description.

hobinjk:
Thanks for the review! Comments addressed

benfrancis:
Thanks, I think this is better. I'm still not 100% sure users will understand the difference, but as long as the default makes sense and the option of more flexibility is there, that's a good place to be in. We can test it with users.

A couple of points:
1. I think "if" and "and" rather than "while" and "or" would be better defaults.
2. I noticed that you can't change while/if until you've dragged in a thing, which I expected to be able to do.

This is really unrelated to this pull request, but is there a reason that all the devices have the symbol for a power switch? I assume that the intention of the designer was that the symbol would reflect the type of device. I think we should show the capability icon as in the floorplan, but without the circle outline. Would you like me to generate images for that and maybe do that in a follow-up? The two images on the rule summary card on the main rules screen was probably also meant to reflect the device types, although that's a bit more tricky now you can include more than two devices.

hobinjk:
The current default is equivalent to "while" so I don't think that changing to "if" is a good idea. Switching to "and" is reasonable though. I can track down whatever's not allowing you to pre-select.

Generating images for use as rule icons would be great! Lack of assets is the only reason that every device is an onoffswitch.

benfrancis:
> Generating images for use as rule icons would be great! Lack of assets is the only reason that every device is an onoffswitch.

I filed https://github.com/mozilla-iot/gateway/issues/1155 with some icons :)

hobinjk:
Comments addressed! I'll get to the icons in the future. I also need to remember to add a play button for the rule preview animation

hobinjk:
https://twitter.com/hobinjk/status/1005182266404999174
There isn't a way of triggering this in the UI but the functionality is there

hobinjk:
Thanks! I improved the display of the drag hints to be bolder

Issue #1071, opened 2018-06-05
Can't always get back to things screen from thing detail view


sogaani:
1. Open browser and access URL `yourgatewaydomain/things/<things-id>` direct.
2. back button does not work.

benfrancis:
It's a bit tricky to know where to go back to.

The thing detail view can be reached either from the things view or the floorplan view so I assume this button uses window.history.back() to go back to whichever page you just came from. If you link to the thing URL directly then it will probably go back to the new tab page!

sogaani:
Oh, I see. It is reasonable behavior. I didn't understand that thing-detail view can be reached from floorplan view.

benfrancis:
I think this is a valid bug report because if you deep link to the detail page there's no way to get back to the main things screen. I'm just not sure what the fix is.

dhylands:
Perhaps store a hidden field on the page which has the needed information?

benfrancis:
Yeah. If you start on that page though, where should you go back to?

Perhaps it should default to going back to the main things screen, unless there's a query string or something which tells it the referrer was the floorplan page.

PR #1099, opened 2018-06-13
Fix issue with back button.

mrstegeman:
page.js updates window.location AFTER routing, which is why we
no longer use it.

STR:
1. Add Virtual Actions & Events Thing
2. Navigate to the thing's actions or events page
3. Click the back button

Expected:
Go back to thing detail page.

Actual:
Go back to things list.


mrstegeman:
@hobinjk https://github.com/mozilla-iot/gateway/issues/1071

PR #1101, opened 2018-06-14
Remove uglify plugin from webpack.

mrstegeman:
The generated source maps were breaking the devtools debugger.


hobinjk:
This increases the main bundle's size from 315 Kb to 586 Kb which in theory increases the file's contribution to page load from 0.8s to 1.5s which isn't too bad in the context of the 13 seconds the full page takes to load.

PR #1118, opened 2018-06-21
Support labels on properties and actions.

### LINKED ISSUE #1109: Human-friendly labels for properties and actions
As a developer I want to provide a human friendly name for properties and actions of my thing, which can be used by a user interface.

As per https://github.com/mozilla-iot/wot/pull/98

mrstegeman:
Fixes #1109

### DIFF
--- a/src/models/thing.js
+++ b/src/models/thing.js
@@ -32,6 +32,9 @@ const Thing = function(id, description) {
   this.id = id;
   this.name = description.name || '';
   this.type = description.type || '';
+  this['@context'] =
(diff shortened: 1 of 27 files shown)


mrstegeman:
@hobinjk Yep, the same as any other change in the thing description.

hobinjk:
Ideally we could allow dynamically updating thing descriptions (possibly limiting to just the human-facing properties like label) but I'm sure that would cause problems somewhere.

hobinjk:
(having to re-add is much less of a problem once we implement the ability to reorder things too)

PR #1121, opened 2018-06-22
Support node 10

### LINKED ISSUE #1120: Deal with the existence of Node 10
- [ ] Fix dependencies that break under node 10
- [ ] Pin nvm to install node 8
- [ ] Run tests on node 10

hobinjk:
Fix #1120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,3 +41,4 @@ after_script:
 node_js:
   - 8
   - 9
+  - 10
--- a/README.md
+++ b/README.md
@@ -54,9 +54,9 @@ Close and reopen your terminal window. Use nvm to install a LTS version of node 
 default version.
 
 ```
-$ nvm install --lts
-$ nvm use --lts
-$ nvm alias default lts/*
+$ nvm install --lts=carbon
+$ nvm use --lts=carbon
+$ nvm alias default lts/carbon
 ```
 
 Verify that node and npm have been installed:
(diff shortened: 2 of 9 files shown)


hobinjk:
This also increases the impact of intermittent travis failures by around 50% which is unfortunate

mrstegeman:
@hobinjk Is the `yarn.lock` change due to running `yarn update`?

dhylands:
It would be good to figure out whats causing the travis failures before merging this.

hobinjk:
Yes, I ran `yarn upgrade` which upgraded far more dependencies than I intended and broke travis

dhylands:
> Yes, I ran yarn upgrade which upgraded far more dependencies than I intended and broke travis

Do we need a separate yarn.lock for node 8 versus node 10?

mrstegeman:
Travis CI also failed while running with node 10, so I don't think so.

mrstegeman:
I'd like to note that some of our add-ons (probably) will not work with node 10 right now.

hobinjk:
No compiled addons will work with node 10 since its NODE_MODULE_VERSION is different. `thing-url-adapter` and `zwave-adapter` were the only adapters I saw with other incompatibilities (which `yarn upgrade` fixed)

dhylands:
I think that just means that we need to build 2 different flavors of the addons and add the NODE_MODULE_VERSION into the list.json file.

mrstegeman:
Yeah, we'll definitely have to put the node version in `list.json`. Let's do that for 0.6, as it's going to be a pretty large change again.

hobinjk:
:+1:

mrstegeman:
I've filed #1122 for the add-ons.

PR #1123, opened 2018-06-24
Support enum property on action input.

sogaani:
```
{
  type: 'string',
  enum: [
    'enum string1',
    'enum string2',
    'enum string3',
  ],
}
```
image

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -481,6 +481,16 @@
   text-align: left;
 }
 
+.action-input-enum {
+  height: 2.75rem;
+  background-color: #d2d9de;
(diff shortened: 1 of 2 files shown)


mrstegeman:
@benfrancis Is this something we want to support? Not sure if the W3C has had any discussions about enums.

benfrancis:
@mrstegeman Actually yes, the latest Editor's Draft includes enum in the [DataSchema](https://w3c.github.io/wot-thing-description/#dataschema) object which can be used by properties, events and action inputs/outputs.

I think this is probably fairly safe.

PR #1127, opened 2018-06-24
Rename BooleanProperty to BinaryProperty.

mrstegeman:


benfrancis:
Do people prefer BinaryProperty or BooleanProperty?

I think I got the name BinaryProperty from BinarySwitch and BinarySensor which is terminology used by Z-Wave command classes (Zigbee call a Binary Sensor a Simple Sensor). It now occurs to me that you could have a property which is actually a binary file or stream of some kind, e.g. an image or video stream (people have asked about this).

Maybe "boolean" is actually a better choice?

mrstegeman:
@benfrancis I don't care either way.

mrstegeman:
@hobinjk @benfrancis Which way do we want this to go? I'll just abandon this if we're sticking with Boolean.

hobinjk:
It makes sense for a binary sensor to have a binary property. On the other hand, it makes some sense for an unknown property of type boolean to be called a boolean property. In general, I think it's best for properties to refer to the logical type "binary data" than the data type "boolean".

To exaggerate the difference, "NumberProperty" makes more sense than a "FloatProperty"

This is just my opinion though so I'd be interested in seeing what @benfrancis thinks

dhylands:
I think that using Boolean is better to use than Binary. Otherwise, I think that a device which has Binary Data might be confused with a Binary Device (i.e. one that just has 2 states).

Although I also think that we need more semantics that just Boolean for 2-state devices. For example, a door lock with a boolean state could be locked or unlocked. It isn't obvious what true maps to, especially when the property currently needs to be named "on". If the property were "locked" with values of true or false, then its pretty clear.

mrstegeman:
@dhylands We can now attach labels (and descriptions) to properties, so that should clear up the confusion.

mrstegeman:
I vote that we stick with `BooleanProperty`.

benfrancis:
> I think that using Boolean is better to use than Binary. Otherwise, I think that a device which has Binary Data might be confused with a Binary Device (i.e. one that just has 2 states).

I agree, and I'm tempted to leave BooleanProperty as it is for that reason. But changing BinarySwitch and BinarySensor to BooleanSwitch and BooleanSensor does seem a bit strange. The best compromise I can think of is that a BinarySwitch should just have a BooleanProperty.

To be honest, I suspect BinarySensor and BinarySwitch won't really be used that much once we have more specific capabilities with added semantics.

> Although I also think that we need more semantics that just Boolean for 2-state devices. For example, a door lock with a boolean state could be locked or unlocked. It isn't obvious what true maps to, especially when the property currently needs to be named "on". If the property were "locked" with values of true or false, then its pretty clear.

We have already defined an OnOffSwitch which is a BinarySwitch with the added semantics that true means on and false means off. We could also define a lock capability with locked and unlocked states (which could just be a boolean property called "locked", or it could be an enum with locked, locking, unlocked, unlocking states).

mrstegeman:
@benfrancis `BinarySwitch` isn't curren't a thing, but I think a `BinarySensor` with a `BooleanProperty` is fine.

benfrancis:
Yeah I just updated the schemas doc to change BinaryProperty to BooleanProperty and noticed that BinarySwitch isn't even listed. An OnOffSwitch has an OnOffProperty. Only a BinarySensor has a BooleanProperty (because OnOffProperty may not be appropriate). I think that's fine.

PR #1132, opened 2018-06-25
Add some RPi libraries supporting Python Bluetooth

pchri:
I only install the libraries.
The Python modules can be installed by the adapter that needs them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,14 @@ The following is required in order to let node use the Bluetooth adapter.
 $ sudo setcap cap_net_raw+eip $(eval readlink -f `which node`)
 ```
 
+### Install Bluetooth and BT Low Energy support libraries (Raspberry Pi Only)
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,6 +83,14 @@ The following is required in order to let node use the Bluetooth adapter.
 $ sudo setcap cap_net_raw+eip $(eval readlink -f `which node`)
 ```
 
+### Install Bluetooth and BT Low Energy support libraries (Raspberry Pi Only)
```
mrstegeman: I would change this to "Linux Only", as it will also apply to other Linux installations.

image/prepare-base.sh
```diff
@@ -49,7 +49,11 @@ sudo apt install -y \
   dnsmasq \
   git \
   hostapd \
+  libboost-python-dev \
+´ libboost-thread-dev \
```
mrstegeman: Stray backtick

PR #1146, opened 2018-06-27
Allow user to select (and change) the device type.

### LINKED ISSUE #1090: Choose device type
As a user I want to choose how my device is represented on the gateway so that I can decide which of its capabilities I'm mainly using it for.

As a user I want to change how my device is represented after adding it so I can change which icon is used.

If a device implements multiple capabilities the user is given a choice when adding the device which capability they want represented in the main things screen, i.e. which icon should be used. The user can also always choose the "Custom Device" type and [choose their own icon](#491) (defaults to the generic "THING" icon).

The user should also be able to change this choice at a later time by clicking on the "Edit" option in the overflow menu in the Thing Detail view.

UI mockup:
choose_device_type

mrstegeman:
Custom icons will come in a secondary PR.

Fixes #1090


static/js/context-menu.js
```diff
@@ -60,7 +62,50 @@ const ContextMenu = {
...
       case 'edit':
         this.nameInput.value = e.detail.thingName;
+        this.thingType.innerHTML = '';
         this.thingIcon.style.backgroundImage = `url("${e.detail.thingIcon}")`;
+
+        for (const capability of e.detail.capabilities) {
+          const option = document.createElement('option');
+          option.value = capability;
+
+          if (e.detail.selectedCapability === capability) {
```
hobinjk: I'm not sure if this is the place to put it, but one thing I'm noticing is that Lights without a selected capability will have their drop-down default to On/Off Switch instead of Light

static/js/context-menu.js
```diff
@@ -60,7 +62,50 @@ const ContextMenu = {
...
+              break;
+            case 'Light':
+              option.innerText = 'Light';
+              break;
+            default:
+              option.innerText = capability;
+              break;
+          }
+
+          this.thingType.appendChild(option);
```
hobinjk: Another behavior is that if there are no options, i.e. a typeless unknown thing, the drop-down menu is the kind of subpar empty drop-down (which encourages clicking but then shows up with nothing) instead of having a fallback "Thing" option. This also applies to the add thing menu

screen shot 2018-06-28 at 12 41 36 pm

static/js/context-menu.js
mrstegeman: @hobinjk That's not the default. The capabilities are added to the dropdown in the order they're listed in the thing description. I think it's pretty much impossible for the UI to determine what the default should be for any given device.

For instance, if you have a smart plug with a light, the primary device would actually be the `OnOffSwitch`, not the `Light`, but that's impossible to tell with a flat `@type` array.

mrstegeman: So I guess if we want `Light` to show up first, we need to add that as the first member of the thing description at the adapter level.

static/js/context-menu.js
mrstegeman: Yeah, I left that intentionally. I'm going to be adding custom icon selection after this is merged, so there will _always_ be something in the dropdown.

mrstegeman:
@hobinjk `ColorControl` just shows up as a box filled with color, not an input. If you're talking about the `ColorDetail`, the input shows up for me just as it always has (except on Safari).

static/js/context-menu.js
mrstegeman: Alternatively, we could just sort the capabilities, that way they're at least in a consistent order.

mrstegeman:
@hobinjk Fixed.

static/js/context-menu.js
hobinjk: I think we should establish a simple hierarchy of capabilities where the ones with more functionality from our UI are preferred. I'm not in favor of saying that web thing developers should sort their exposed capabilities in order of preference since to me that feels very non-obvious.

I think it's reasonable that the gateway's UI as a consumer of thing descriptions would make arbitrary decisions to prefer those capabilities which it can best represent. In the case of a smart plug that advertises both Light and OnOffSwitch, I think it makes sense to default to Light since the thing's developer had some reason to include it.

mrstegeman: @hobinjk I like that. Pushed.

PR #1147, opened 2018-06-28
Fix up default property detail types.

mrstegeman:


mrstegeman:
@hobinjk This should fix the issue you saw when choosing `OnOffSwitch` for your Hue light.

PR #1149, opened 2018-06-29
security: post OAuth token to client webapps if any

rzr:
When loaded from window.opener CORS is blocked by default,
so this is workaraounding security issues like:

  Uncaught DOMException: Blocked a frame with origin (...)
  from accessing a cross-origin frame. at (...)

For any security concern this change can be reverted anytime.

Change-Id: I42af71ae822491150c019cff9688356b1a0e2532

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -74,5 +74,6 @@ module.exports = {
       ssid_base: 'WebThings Gateway',
     },
   },
+  oauthPostToken: false,
   oauthTestClients: false,
 };
--- a/src/router.js
+++ b/src/router.js
@@ -10,6 +10,7 @@
 
 'use strict';
 
+const config = require('config');
 const compression = require('compression');
 const Constants = require('./constants');
 const express = require('express');
@@ -43,7 +44,11 @@ const Router = {
       }
 
       // Disable embedding
-      response.set('Content-Security-Policy', 'frame-ancestors \'none\'');
+      response.set('Content-Security-Policy',
+                   config.get('oauthPostToken') ?
+                     'frame-ancestors filesystem:' :
+                     'frame-ancestors \'none\''
+      );
 
       next();
     });
--- a/src/views/local-token-service.handlebars
+++ b/src/views/local-token-service.handlebars
@@ -123,6 +123,18 @@ let text = client.get("<span class="origin">https://gateway.local</span>/things"
         </div>
       </div>
     </section>
+    <script>
+if ({{oauthPostToken}}) {
+  window.addEventListener("message", function(ev) {
+   if (ev.origin !== 'file://') {
+     return;
+   }
+   if (ev.data.message === "token") {
+      ev.source.postMessage({ message: { token: "{{token}}" }}, "*");
+    }
+  });
+}
+    </script>
   </body>
 </html>
 
--- a/src/controllers/oauth_controller.ts
+++ b/src/controllers/oauth_controller.ts
@@ -12,6 +12,7 @@ import * as express from 'express';
 import { URL } from 'url';
 import * as assert from 'assert';
 const JSONWebToken = require('../models/jsonwebtoken');
+const config = require('config');
 import * as Database from '../db';
 import {
   scopeValidSubset, Scope, ScopeAccess, ScopeRaw, ClientId, ClientRegistry
@@ -265,6 +266,7 @@ OAuthController.get('/local-token-service', async (request: express.Request, res
   let token = await handleAccessTokenRequest(request, response);
   if (token) {
     response.render('local-token-service', {
+      oauthPostToken: config.get('oauthPostToken'),
       token: token.access_token
     });
   }


hobinjk:
At this level of integration you would be better of setting up a proper OAuth service by editing src/models/oauthclients.ts as described in [this setup document](https://github.com/mozilla-iot/wiki/wiki/Running-OAuth-Locally).

rzr:
Yea I tried this once, but on some browsers  was not possible from file://... origin (tizen local webapp) so I hacked around to make it work somehow, until a better solution is found.

FYI, I used this bit for PWA app:
https://vimeo.com/277663462#webthing-webapp-pwa-20180629rzr

Maybe I will refactor this proof of concept later, but I wished the code could support several platforms.

hobinjk:
Yeah, unfortunately this would make it really easy for an attacker to embed a frame of the local token service in their page and trick the user into giving them complete access to their gateway.

I think it's possible to limit the message listener to only respond if the message originates from a file:// URL through ev.origin or something. That would make this PR much more secure.

rzr:
Yes I was about to suggest this, meanwhile I got proper OAuth on http.
but Instead of testing file:// protocol I prefer to test location.hostname != null, in case other protocol to be used, would that be fine ?

hobinjk:
I would prefer a whitelist of specific protocols so that the exact behavior is completely specified.

mrstegeman:
@rzr @hobinjk Is this still needed for anything?

rzr:
I think so it was needed for an "offline" webapp PoC... (loaded from file:// protocol), I made it evolve to support various frameworks (eg: aframe) , so it would be nice to have it in.

hobinjk:
:+1:, after #1712 this just needs an additional modification to https://github.com/mozilla-iot/gateway/blob/master/src/router.js#L46 to have `'filesystem:'` instead of `'none'`

hobinjk:
Additionally, since this adds the ability for any random file:// page to clickjack the gateway, the modification to the content-security policy should be guarded by a config option (like https://github.com/mozilla-iot/gateway/commit/7fbaf569954b412570dd6cc9376ffc75083cc5ff)

config/default.js
```diff
@@ -74,5 +74,6 @@ module.exports = {
       ssid_base: 'WebThings Gateway',
     },
   },
+  oauthPostToken: !false,
```
mrstegeman: This should default to false.

src/views/local-token-service.handlebars
```diff
@@ -123,6 +123,15 @@ let text = client.get("<span class="origin">https://gateway.local</span>/things"
         </div>
       </div>
     </section>
+    <script>
+/* postToken to client apps if any */
+if (config.get('oauthPostToken') && (window.location.protocol === "file:") {
```
mrstegeman: This is a handlebars template that gets used in the UI. Instead of using `config`, you'll need to pass a value into the template.

hobinjk: The Content-Security-Policy setting means we'll only be embedded in trusted situations so this doesn't need the config.get here

config/default.js
hobinjk: This should be `false`

src/views/local-token-service.handlebars
```diff
@@ -123,6 +123,15 @@ let text = client.get("<span class="origin">https://gateway.local</span>/things"
...
+  window.addEventListener("message", function(ev) {
+    if (ev.data.message === "token") {
+      ev.source.postMessage({ message: { token: "{{token}}" }}, "*");
+    }
+});
```
hobinjk: The indentation of this is a bit off and the close brace from the initial `if` statement is missing

config/default.js
rzr: yes, I used !false as reminder

src/views/local-token-service.handlebars
rzr: yes that code didnt work actually I figured out I used other hack to get token, next version works anyway

rzr: I still use oauthPostToken to prevent a not wanted post action 
see next revison

config/default.js
rzr: fixed

src/views/local-token-service.handlebars
rzr: fixed

src/router.js
```diff
@@ -43,7 +44,11 @@ const Router = {
       }
 
       // Disable embedding
-      response.set('Content-Security-Policy', 'frame-ancestors \'none\'');
+      response.set('Content-Security-Policy',
+                   config.get('oauthPostToken') ?
+                     'frame-ancestors filesystem:' :
```
mrstegeman: I believe `filesystem:` needs single quotes around it: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

rzr: Seems not , well with quotes I got an alert that they were ignored... now i am unsure

mrstegeman: I think you're right. Confusing docs on MDN.

mrstegeman:
@hobinjk Any other objections?

src/router.js
rzr: Yes I fixed MDN with help of @sheppy (thx)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors$revision/1535837

PR #1150, opened 2018-06-29
Allow user to choose a custom icon.

### LINKED ISSUE #491: Set custom icon for Thing
As a user I want to choose a custom icon for a Thing so the icon better represents the thing (e.g. a lamp icon for a lamp plugged into a smart plug which would have a smart plug icon by default).

mrstegeman:
Fixes #491

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "@webcomponents/webcomponentsjs": "^2.0.2",
     "archiver": "^2.1.1",
     "asn1.js": "^5.0.1",
-    "atob": "^2.1.1",
     "awesome-typescript-loader": "^5.0.0",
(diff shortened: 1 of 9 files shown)


hobinjk:
There doesn't seem to be any surfacing of the "payload too large" error that I get whenever I choose an image that is too large (admittedly I don't think anyone wants to have a 1920x1080 desktop background as an icon).

I'm also unable to set a small .png generated through the screenshot tool as the icon with the error "The image can not be displayed because it contains errors"

mrstegeman:
@hobinjk That's what I get for only testing with small SVGs...

mrstegeman:
@hobinjk Those issues should be fixed now.

PR #1151, opened 2018-06-29
Replace default icons.

mrstegeman:


mrstegeman:
These were provided by @benfrancis. It reduces the amount of duplicate icons we have all over the place, and provides base icons for all supported capabilities.

PR #1153, opened 2018-07-02
Fix CSS issues on small screens.

mrstegeman:


mrstegeman:
@hobinjk I noticed that as well.

PR #1160, opened 2018-07-03
Add EnumDetail.

mrstegeman:
Modernized version of #923, with support for proper JSON-schema
enum.

@benfrancis Is this something we'd like to have? It looks like `enum`
is present in the current W3C draft.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -230,6 +230,7 @@ require('./components/property/brightness');
 require('./components/property/color');
 require('./components/property/color-temperature');
 require('./components/property/current');
+require('./components/property/enum');
 require('./components/property/frequency');
 require('./components/property/instantaneous-power');
 require('./components/property/label');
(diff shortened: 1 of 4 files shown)


mrstegeman:
screenshot_2018-07-03 things gateway

mrstegeman:
@benfrancis In @steveatinfincia's [picamera-webthing](https://github.com/infincia/picamera-webthing/blob/master/picamera-webthing.py#L292), he used a unit of `FPS` for the framerate enum.

trentrand:
Are there any plans to support enum option labels; whether through use of the non-standard `enumNames` property or by creating an EnumProperty via an `anyOf` property?

benfrancis:
Hmm, I hadn't thought of that. Not currently no. Where have you seen the non-standard `enumNames` property? Is there not a standard way to do this?

Isn't `anyOf` meant to be for validation? i.e. to validate that a given value can be valid against a given number of sub-schemas?

mrstegeman:
@benfrancis See #1310

PR #1166, opened 2018-07-05
Allow python3 to use the Bluetooth adapter

pchri:
I would have thought that python inherited the capability from node (that's what the 'i' flag in +eip means), but it seems not to be the case.

### DIFF
--- a/image/prepare-base.sh
+++ b/image/prepare-base.sh
@@ -39,9 +39,6 @@ export NVM_DIR="$HOME/.nvm"
 nvm install ${NODE_VERSION}
 nvm use ${NODE_VERSION}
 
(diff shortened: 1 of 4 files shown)


image/prepare-base.sh
```diff
@@ -72,6 +72,9 @@ sudo pip3 install "$_url"
 _url="git+https://github.com/mycroftai/adapt#egg=adapt-parser"
 sudo pip3 install "$_url"
 
+# Allow python3 to use the Bluetooth adapter
+sudo setcap cap_net_raw+eip $(eval readlink -f `which python3`)
```
mrstegeman: Could you move this up to line 43, or move the other `setcap` call down here?

PR #1171, opened 2018-07-06
Make properties of capabilities more flexible.

mrstegeman:
Also, make InstantaneousPowerProperty for SmartPlug optional.

References mozilla-iot/schemas#2

### DIFF
--- a/static/js/components/capability/smart-plug.js
+++ b/static/js/components/capability/smart-plug.js
@@ -62,12 +62,19 @@ class SmartPlugCapability extends BaseComponent {
 
     this._icon = this.shadowRoot.querySelector('#icon');
     this._label = this.shadowRoot.querySelector('#label');
+
+    this._havePower = false;
     this._on = false;
     this._power = 0;
+
     this._onClick = this.__onClick.bind(this);
   }
 
   connectedCallback() {
+    this._havePower =
+      typeof this.dataset.havePower !== 'undefined' ?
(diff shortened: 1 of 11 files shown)


static/js/components/capability/smart-plug.js
```diff
@@ -62,12 +62,19 @@ class SmartPlugCapability extends BaseComponent {
...
     this._on = false;
     this._power = 0;
+
     this._onClick = this.__onClick.bind(this);
   }
 
   connectedCallback() {
+    this._havePower =
+      typeof this.dataset.havePower !== 'undefined' ?
+        this.dataset.havePower :
```
hobinjk: If someone specifies `data-have-power="false"` this evaluates to `"false"` which is truthy (oh Javascript, what a beautiful language)

mrstegeman:
@hobinjk Fixed the one issue. That would be rather convoluted, by I can do it if you think it'll make a big difference. In my mind, all properties from legacy devices will have no `@type`, while things moving forward should (ideally) have a `@type` for each property.

hobinjk:
My only concern would be if someone makes a DIY thing which is a control panel with both a "level" for some potentiometer input and a "brightness" for an LED. It's relatively easy to just tell this hypothetical person that they need to come up with better names though

mrstegeman:
@hobinjk Addressed.

hobinjk:
Great, thanks! LGTM

PR #1177, opened 2018-07-09
Use min/max/unit of LevelProperty.

mrstegeman:


static/js/property-detail/level.js
```diff
@@ -27,12 +46,15 @@ class LevelDetail {
   }
 
   view() {
+    const min = `min=${Utils.escapeHtml(this.min)}`;
```
hobinjk: This should both be quoted

mrstegeman: Good catch! Fixed.

PR #1178, opened 2018-07-09
Use Avahi for mDNS on Raspberry Pi.

### LINKED ISSUE #1055: Domain Settings doesn't reflect full mDNS state
See https://github.com/mozilla-iot/gateway/pull/1049#issuecomment-393892397

* The default value which appears in the Local Access host name setting does not reflect the "gateway.local" hostname which Avahi is broadcasting
* When the "Local Access" setting is disabled in settings, Avahi is still broadcasting "gateway.local"
* If the user tries to set the local host to "gateway.local", it conflicts with Avahi and gets set to something else instead

mrstegeman:
Fixes #1055


benfrancis:
Awesome, thanks @mrstegeman

PR #1183, opened 2018-07-10
Return intent payload in successful command creation

hobinjk:


hobinjk:
the hacked-together prototype: https://github.com/hobinjk/commander

PR #1186, opened 2018-07-11
Fix issue with level component.

mrstegeman:
If the initial value was the minimum, the number input was not
being set.

### DIFF
--- a/static/js/components/property/level.js
+++ b/static/js/components/property/level.js
@@ -145,6 +145,9 @@ class LevelProperty extends BaseComponent {
     this._upgradeProperty('value');
     this._upgradeProperty('disabled');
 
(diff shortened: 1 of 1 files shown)


static/js/components/property/level.js
```diff
@@ -145,6 +145,9 @@ class LevelProperty extends BaseComponent {
     this._upgradeProperty('value');
     this._upgradeProperty('disabled');
 
+    // Set the initial value so that both inputs are set properly.
+    this.value = parseInt(this.min, 10);
```
hobinjk: Is there a chance that value could already be initialized here from the call to _upgradeProperty?

mrstegeman: Good call. Updated and should be safer now.

PR #1190, opened 2018-07-12
Use webcomponents on floorplan view

### LINKED ISSUE #912: Show thing status on floorplan
As a user I would like to see the status of things in the floorplan view so that I can get an at-a-glance overview of the status of all my things

hobinjk:
Replace SVG icons with html webcomponents. Fix #912.

- [x] Remove any vestigial code
- [x] Rename 'svg' format to 'floorplan' or equivalent

### DIFF
--- a/static/css/floorplan.css
+++ b/static/css/floorplan.css
@@ -20,6 +20,37 @@
   background-position: center;
 }
 
+.floorplan-thing {
+  position: absolute;
+  font-size: 2rem;
+  width: 12.8rem;
+  text-align: center;
+  color: white;
+  cursor: pointer;
+  user-select: none;
+}
+
+.floorplan-thing-link {
+  display: block;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background: rgba(0, 0, 0, 0.0);
+  z-index: 10;
+}
+
+#floorplan.editing .floorplan-thing-link {
+  display: none;
+}
+
+.floorplan-thing-name {
+  margin-top: 0.5rem;
+  max-height: 6rem;
+  overflow: hidden;
+}
+
 #floorplan-edit-button {
   bottom: 2rem;
   right: 2rem;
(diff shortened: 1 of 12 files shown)


benfrancis:
OK, so there's a lot I like about this. I like being able to re-use the web components.

The one big problem, which is why I implemented this in SVG in the first place, is that the icons don't scale with the floorplan.

So while this looks OK on a large screen:

screenshot from 2018-07-13 19-47-13

When it scales down things start to go wrong:
screen shot 2018-07-13 at 19 43 07

Whereas before it would have looked like this:
screen shot 2018-07-13 at 19 41 07

On mobile it's basically game over:
screen shot 2018-07-13 at 19 49 10

whereas before it would have looked like this:
screen shot 2018-07-13 at 19 55 40


Did you try scaling web components proportionally to the floorplan background? I don't imagine it works as well as pure SVG but given the icons themselves are SVG it shouldn't be too bad. If we can get that to work, this might be a workable solution.

I intentionally reduced the information density on the mockups for the floorplan version with a subtly simpler representation because the text ends up being very small using these web components, but we can see how it looks just scaling the existing web components.

A few other comments:
- The text was intentionally hidden on the non-edit view because it has a tendency to look messy and once positioned a thing's position should indicate which device it is
- When the text is displayed (in the edit screen) it needs some kind of drop shadow so it still shows up on white backgrounds
- The cursor should change to a grab icon when in edit mode to show that icons are dragged rather than clicked
- Is there any way to stop the flashing that we see when transitioning between floorplan and detail views while it's re-drawing the web component?
- Any idea why there's a red square outline on the thing icons in the active state? Is that a browser default?

hobinjk:
Oh right, sorry for forgetting about the scaling! I had it planned initially but after scale(calc(128px/10vmin)) didn't work I shelved it for later. It will be a simple matter of attaching a resize listener.

I can make the readability changes, thanks for pointing them out!

I don't think there's any easy way to stop the flash of a not-completely-loaded page but it's possible that using the page() function over real navigation will speed things up.

I haven't seen this red square outline so I'm not too sure what could be going on there.

hobinjk:
Great, thanks! Filed #1193 for the flash of placeholders. I agree that the naming isn't perfect and filed #1194 for that

PR #1196, opened 2018-07-17
Manage js library with yarn

sogaani:
Latest page.js which supports `a` tag on SVG avoids reloading things-detail page, when user clicks icon on floorplan view.


hobinjk:
For some reason this breaks the highlighting on the local token service. I thought it would be fixed by running the code in window.addEventListener('load', ...) but something else is preventing it from working.

PR #1204, opened 2018-07-20
Add interactivity to floorplan

hobinjk:
Tap to interact with tappable things like lights. Long press to visit a
thing's detail page.

### DIFF
--- a/static/css/floorplan.css
+++ b/static/css/floorplan.css
@@ -27,7 +27,6 @@
   height: 12.8rem;
   text-align: center;
   color: white;
-  cursor: pointer;
   user-select: none;
   touch-callout: none;
   -webkit-touch-callout: none;
--- a/static/css/floorplan.css
+++ b/static/css/floorplan.css
@@ -29,6 +29,8 @@
   color: white;
   cursor: pointer;
   user-select: none;
+  touch-callout: none;
+  -webkit-touch-callout: none;
 }
 
 #floorplan.editing .floorplan-thing {
--- a/static/js/floorplan.js
+++ b/static/js/floorplan.js
@@ -489,8 +489,6 @@ const FloorplanScreen = {
     if (this.interactTimeout) {
       if (thing.handleClick) {
         thing.handleClick();
-      } else if (this.selectedThing) {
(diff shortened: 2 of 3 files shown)


static/js/floorplan.js
```diff
@@ -10,6 +10,7 @@
 
 'use strict';
 
+const page = require('page');
```
benfrancis: This helps a lot.

static/css/floorplan.css
```diff
@@ -29,6 +29,8 @@
   color: white;
   cursor: pointer;
   user-select: none;
+  touch-callout: none;
+  -webkit-touch-callout: none;
```
benfrancis: Huh, I didn't know this existed.

static/js/floorplan.js
```diff
@@ -158,13 +164,31 @@ const FloorplanScreen = {
...
       });
     });
   },
 
+  addInteract: function(thing) {
+    thing.element.addEventListener('click', this.blackHole);
+    thing.element.addEventListener('mousedown', this.onPointerDown);
+    thing.element.addEventListener('mouseup', this.onPointerUp);
+    thing.element.addEventListener('touchstart', this.onPointerDown);
+    thing.element.addEventListener('touchend', this.onPointerUp);
```
benfrancis: One of these events seems to be firing on right click, so that right clicking an icon turns it on and off (or navigates to the detail view in the case of custom things). This probably shouldn't happen.

benfrancis:
Oh, I just noticed this totally breaks in edit mode. Dragging the icons around navigates triggers the timeout and navigates to the detail view.

hobinjk:
Thanks for the review! I've fixed the interaction bugs. I agree that this interaction is sub-par. I think it would be great to somehow "splat" out the details of things (or a proper menu) on the floorplan but that would be rather difficult to implement.

static/css/floorplan.css
hobinjk: Yeah, it's apparently only an iOS thing but I threw it in there just in case

hobinjk:
Great, thanks! I've removed that tap-to-visit for consistency

PR #1205, opened 2018-07-20
Add speech input to Assistant

### LINKED ISSUE #1200: Merge speech and assistant experiments

hobinjk:
Fix #1200 

The basic flow of press button -> speak -> text shows up -> command is executed works. However, it requires a major refactor of speech code to have visual feedback without a ton of copy-pasting.


hobinjk:
The coolest part of this PR is that now the speech experiment uses the same implementation as the Assistant which means that no matter how you talk to it there's a persistent log of its actions and what it thinks you said.

hobinjk:
That nitpick opened up a whole can of worms, wow! It turns out that the canvases were defaulting to be 300x150 which was fine until I tried to reposition the circle by calculating exactly where it should be based on the layout. There was even an existing workaround where the code secretly scaled all of the circles to be ellipses to work around the weird default canvas dimensions.

Either way the circle now has a pixel perfect position

PR #1206, opened 2018-07-23
build: Install missing dependencies

rzr:
This has been tested in docker's debian:stable image

Also removing hardcoded Pi user as it could support other id (even root),
or other target device.

Change-Id: I537ad8c8c3a3b4af161b61a7968073f4d3b7cee1

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,14 +3,22 @@
 NVM_VERSION="v0.33.8"
 NODE_VERSION="--lts=carbon"
 
-cd /home/pi
+[ "$USER" != "" ] || USER="pi"
+[ "$HOME" != "" ] || HOME="/home/${USER}"
+cd "${HOME}"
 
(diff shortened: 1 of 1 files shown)


mrstegeman:
@rzr We've already got a [Docker image](https://github.com/mozilla-iot/gateway-docker), will yours do anything different?

rzr:
Oh you made that too, well mine is just testing the gateway on raspbian (using qemu binfmt),
without installing services:

https://github.com/TizenTeam/gateway/blob/sandbox/rzr/devel/master/Dockerfile

maybe both can be used in different use cases

Anyway this change is safe to merge:

https://github.com/mozilla-iot/gateway/pull/1206

rzr:
FYI this change will help to get support of gateway on ARTIK7, (this should also help for ARTIK5 or ARTIK10)
If any one is able to test let me know I will share some instructions.

PR #1217, opened 2018-07-27
Run all of webpack through cache-loader

hobinjk:
Reduces first load time from 2:49 to 2:19 and repeat load to 0:54. I'm continuing to look into other opportunities and it remains to be seen whether this persists in an image.

Notably, @dhylands, the rpi-image-builder script will have to be modified to run webpack once and bundle in the .cache-loader directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.swp
 .DS_Store
 .nyc_output/
+.cache-loader/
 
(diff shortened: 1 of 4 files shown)


hobinjk:
This needs a bit more work, it's dropping the CSS on the floor somewhere

Issue #1211, opened 2018-07-26
Add page titles to Assistant and Things screens


benfrancis:
Kathy pointed out that some pages have titles and others don't.

Let's try adding titles to the Assistant and Things screens and see how well it works. The title should scroll off the page though, rather than have the content disappear behind an invisible line beneath it.

We could also try adding a title to the floorplan page, but it might need a dropshadow to account for white backgrounds, so the icon probably wouldn't work.

May require some experimentation.

mrstegeman:
We decided this wasn't a great idea. See #1323 for discussion.

Issue #1198, opened 2018-07-18
Horizontal margin clipping


benfrancis:
At some point after 0.4 a change was made that means scrollable views get clipped between an opaque top and bottom margin. I suspect this may be related to adding titles to views like the title on the Settings screen and possibly to get around some layout issues.

Before:
![screen shot 2018-07-18 at 18 16 07](https://user-images.githubusercontent.com/552417/42897121-e9038ae4-8ab6-11e8-9c60-6a0f6655150e.png)

After:
![screen shot 2018-07-18 at 18 16 09](https://user-images.githubusercontent.com/552417/42897124-ee2b74d2-8ab6-11e8-9576-9d42f2c8229d.png)

Before:
![screen shot 2018-07-18 at 18 16 25](https://user-images.githubusercontent.com/552417/42897132-f25df890-8ab6-11e8-9bb0-bc75195757bf.png)

After:
![screen shot 2018-07-18 at 18 16 27](https://user-images.githubusercontent.com/552417/42897138-f9cb9c2c-8ab6-11e8-9f88-589230ec75a4.png)

I think it looks worse than it did before and we should try and change it back. I think the Settings title should scroll like the Rules title does, rather than stay absolutely positioned while the settings menu scrolls underneath it. Otherwise UI elements disappear behind an invisible line and it looks broken.

hobinjk:
This also applies in more difficult situations like the thing details view where on a very small screen there needs to be a decision of either clipping or requiring scrolling to view everything. The other place this decision is a bit harder to make is on the Things page with the new thing, menu, and speech buttons which will have thing icons drawn beneath them on smaller screens.

I think the most important thing is for it to be consistent everywhere, which seems tough. We could also improve the clipping solution by creating a designated title bar visually separated from the rest of the page. That would be a pretty major redesign

benfrancis:
Related:

<img width="1279" alt="screen shot 2018-07-27 at 19 42 30" src="https://user-images.githubusercontent.com/552417/43340774-51ae0b4e-91d5-11e8-970b-9cdb8a8864be.png">

hobinjk:
#553 is the last iteration of this. Here's a couple mockups of different approaches to the padding-less approach:
Change nothing, remove padding:
<img width="152" alt="screen shot 2018-07-27 at 3 28 52 pm" src="https://user-images.githubusercontent.com/2776089/43342812-0e0e250e-91b2-11e8-8fbf-cd55477561ff.png">

[...]

PR #1222, opened 2018-07-27
Visual polish for 0.5.0

hobinjk:


benfrancis:
Thanks for doing this. I had generally avoided drop shadows in order to keep a flat visual design, but this is very subtle and does solve the problem on the floorplan, so nice idea.

Sorry if my screenshots in #1198 confused matters, but my intention wasn't to completely remove the padding on the things screen, just to make it part of the scrollable area rather than outside it. i.e. in the initial state it should look like this:
screen shot 2018-07-30 at 16 00 31

rather than this:
screen shot 2018-07-30 at 16 00 32

...but should then scroll underneath the buttons (as it was in 0.4) rather than disappearing behind an invisible border (as it currently is in master). We can then potentially also add a Things title which scrolls off the page, as per #1211.

One other thing, I've noticed a regression in the rules screen where the menu button now appears instead of the back button:
screen shot 2018-07-30 at 15 54 46

hobinjk:
Right, makes sense. I've fixed most of the issues but noticed an overarching problem with the approach I took to calling out the buttons which also applies to the use of webcomponents on the floorplan. The issue is that the icons have a low opacity fill. This blends well into any backdrop but also blends into the icons on the things list. I fixed this by giving them the light background color to make them stand out from the things list. However, this backfires on the floorplan's edit view where the backdrop is dark instead of light so the manual blending doesn't look correct. The webcomponents also have an incorrect background based on the light one.

screen shot 2018-07-30 at 11 49 54

The webcomponents are straightforward to fix but this means that I'll have to take a different approach on the buttons if we want them to be visually distinct. The easiest way to fix this is to do what the thing edit view does. It hides all irrelevant buttons on the floorplan edit view and specifies its own back button which can then be styled correctly. This would also fix an issue where you can navigate away in the middle of editing the floorplan and leave the screen in a weird intermediate state.

tl;dr: spotted overarching issue while fixing minor issues, will fix and update PR accordingly

hobinjk:
Hmm, the icon opacity is actually specified in the css, not the svg. It doesn't change the approach I'll be taking but I'm updating here so there are no misconceptions left by my previous comment.

hobinjk:
screen shot 2018-07-30 at 12 50 00

With the buttons now more muted I'm more okay with the components sticking out like that, let me know if you want me to change anything.

hobinjk:
Hmm, sorry for forgetting about Settings. Looking into it the design of the view, it would be a major refactoring to get it to scroll properly. It would be a very nice improvement though so if I get any time I'll try to tackle it.

PR #1233, opened 2018-07-31
Visual polish round two

hobinjk:


static/index.html
```diff
@@ -41,7 +41,9 @@
         <div id="thing-title-speech-flex"></div>
       </div>
       <div id="things-container">
+        <div class="things-pad"></div>
```
benfrancis: I'm really not keen on adding empty divs with no semantic purpose just to add padding. Why is this necessary now when it wasn't in 0.4?

This looks fine visually so I definitely wouldn't hold the release on this nit-pick, but I feel like there's a cleaner solution.

hobinjk: The divs are necessary to do the convoluted "vertical-align but if scrolling then pad by an invisible scrolling 7.6rem" layout. It wasn't necessary before because the layout would either clip or not have the padding at the top. This was the only solution I could find.

hobinjk:
Interesting! I tracked it down and it turns out that the same thing that added clipping added a div that looked like it wouldn't do anything but actually made it so that the 0.4-era approach would fail. 
I've reverted to using the exact 0.4-era approach

hobinjk:
Oops, my bad, added back the text-align: center I overzealously deleted.

Issue #1238, opened 2018-08-01
gateway advertises gateway.local even if that isn't the hostname


dhylands:
When I run the gateway on my linux development machine and try to ping gateway/local from that machine, it pings itself.

This is clearly incorrect. If it's going to be advertising anything it should be advertising the hostname and not a hard coded 'gateway'.

This winds up causing conflicts with the raspberry pi, which will then advertise gateway-2.local while this is happening.

And sure enough, on the host machine, I see:
```
2018-08-01 12:23:52.595 DNS config is: gateway
```

PR #1240, opened 2018-08-01
Change the default MDNS name to be the hostname

dhylands:


dhylands:
This fixes #1238

benfrancis:
@dhylands What effect does this actually have? Might it broadcast something other than gateway.local even if the user hasn't changed that in settings? If so, how would they know?

dhylands:
> @dhylands What effect does this actually have? Might it broadcast something other than gateway.local even if the user hasn't changed that in settings? If so, how would they know?

On the Raspberry Pi, this should make no significant difference since the hostname is gateway.

However, when I run the gateway on my linux machine it now advertises my linux machines hostname rather than gateway.local

PR #1242, opened 2018-08-01
Fixes to prepare-base.sh for the merged wifi setup

### LINKED ISSUE #1241: Error running prepare-base.sh
I just finished building a base image and saw this error:
```
Now using node v8.11.3 (npm v5.6.0)
Creating default alias: default -> lts/carbon (-> v8.11.3 *)
Now using node v8.11.3 (npm v5.6.0)
hostapd.service is not a native service, redirecting to systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install disable hostapd
Synchronizing state of dnsmasq.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install disable dnsmasq
cp: cannot stat 'config/hostapd.conf': No such file or directory
```

That comes from: https://github.com/mozilla-iot/gateway/blob/6724d401cee3ca278b5f448c9d5adcfccd1910b3/image/prepare-base-root.sh#L23

Those files originated from here:
https://github.com/mozilla-iot/gateway-wifi-setup/tree/master/config

So I think that they might need to relocated if we're using a merged gateway/wifi-setup

dhylands:
This should fix #1241


dhylands:
The make-prep.sh script copies a few files (prepare-base.sh and prepare-base-root.sh) to the home directory. Most of the files the scripts needs are actually embedded inside the script itself as a "here" document.

At this stage, we're setting up to build the base image, so the gateway repository hasn't been checked out or anything.

I'd probably update the scripts to use "here" documents in the future. See this for an example:
https://github.com/mozilla-iot/gateway/blob/6724d401cee3ca278b5f448c9d5adcfccd1910b3/image/prepare-base.sh#L79-L101

PR #1243, opened 2018-08-01
Fix path to hostapd and dnsmasq config files

### LINKED ISSUE #1241: Error running prepare-base.sh

hobinjk:
Fix #1241

### DIFF
--- a/image/prepare-base-root.sh
+++ b/image/prepare-base-root.sh
@@ -20,6 +20,6 @@ systemctl disable hostapd.service
 systemctl disable dnsmasq.service
 
 # install hostapd and dnsmasq config files
-cp config/hostapd.conf /etc/hostapd/hostapd.conf
-cp config/dnsmasq.conf /etc/dnsmasq.conf
+cp image/config/hostapd.conf /etc/hostapd/hostapd.conf
+cp image/config/dnsmasq.conf /etc/dnsmasq.conf
 echo 'DAEMON_CONF="/etc/hostapd/hostapd.conf"' >> /etc/default/hostapd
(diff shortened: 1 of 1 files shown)


dhylands:
prepare-base-root.sh is now small enough, that I think I'm just going to merge it back into prepare-base.sh, which already has a bunch of sudo's in it.

dhylands:
Looks like we need to copy the post-upgrade script as well.

What does it expect the current directory to be? I see that it touches the .post_upgrade_complete file.

hobinjk:
Yeah, I think you can just revert 84e91b9bd421ea788d7959b57950cf7556750fb5. I wrote that patch with the assumption that prepare-base would run with the gateway present. I'll look at what I can remove from post_upgrade.sh or if I can improve its dismal performance some other way

PR #1249, opened 2018-08-02
ui: Allow multiple same properties in single thing

### LINKED ISSUE #1148: MultiLevelSensor gets confused with multiple LevelProperties with different units
If a `MultiLevelSensor` thing has multiple `LevelProperty` properties, it adapts the last updated value as its current main value. It does however not adapt the unit of the updated property, so the center of the thing will contain a number with the wrong unit.

rzr:
Previously item ids were duplicated,
the observed consequence is that any click on widgets
is seen as click on 1st declared element.

For example, a thing with 2 switches is displayed correctly,
and click events are also triggered but will be "redirected"
to 1st switch only, this is not desired.

Change-Id: I27ef75a95889c37224aa7da6b0fbf8e2ad57ba65
Bug: https://github.com/mozilla-iot/gateway/issues/1148
Fixes #1148
Origin: https://github.com/tizenteam/gateway

### DIFF
--- a/static/js/components/property/switch.js
+++ b/static/js/components/property/switch.js
@@ -64,7 +64,7 @@ class SwitchProperty extends BaseComponent {
       transition: 0.1s;
     }
 
-      .webthing-switch-property-switch:checked + .webthing-switch-property-slider::after {
+    .webthing-switch-property-switch:checked + .webthing-switch-property-slider::after {
       transform: translate(3.65rem, 0.35rem);
     }
 
@@ -85,9 +85,9 @@ class SwitchProperty extends BaseComponent {
   <div id="container" class="webthing-switch-property-container">
     <div id="contents" class="webthing-switch-property-contents">
       <form>
-          <input type="checkbox" id="switch-${BaseComponent.count}"
+        <input type="checkbox" id="switch-${BaseComponent.count}"
           class="webthing-switch-property-switch">
-          <label id="slider-${BaseComponent.count}" for="switch-${BaseComponent.count}" class="webthing-switch-property-slider">
+        <label id="slider-${BaseComponent.count}" for="switch-${BaseComponent.count}" class="webthing-switch-property-slider">
         </label>
       </form>
       <div id="label" class="webthing-switch-property-label"></div>
--- a/static/js/components/property/switch.js
+++ b/static/js/components/property/switch.js
@@ -11,8 +11,10 @@
 
 const BaseComponent = require('../base-component');
 
-const template = document.createElement('template');
-template.innerHTML = `
+class SwitchProperty extends BaseComponent {
+  constructor() {
+    const template = document.createElement('template');
+    template.innerHTML = `
   <style>
     :host {
       display: inline-block;
@@ -62,7 +64,7 @@ template.innerHTML = `
       transition: 0.1s;
     }
 
-    .webthing-switch-property-switch:checked + #slider::after {
+      .webthing-switch-property-switch:checked + .webthing-switch-property-slider::after {
       transform: translate(3.65rem, 0.35rem);
     }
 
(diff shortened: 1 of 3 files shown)


mrstegeman:
@rzr 

```
> eslint .

/home/travis/build/mozilla-iot/gateway/static/js/components/base-component.js
  15:11  error  'idSuffix' is assigned a value but never used. Allowed unused vars must match /^_/  no-unused-vars

/home/travis/build/mozilla-iot/gateway/static/js/components/property/boolean.js
  17:13  error  Infix operators must be spaced            space-infix-ops
  18:5   error  Unexpected var, use let or const instead  no-var

/home/travis/build/mozilla-iot/gateway/static/js/components/property/switch.js
  17:13  error  Infix operators must be spaced            space-infix-ops
  18:5   error  Unexpected var, use let or const instead  no-var

✖ 5 problems (5 errors, 0 warnings)
  4 errors, 0 warnings potentially fixable with the `--fix` option.
```

rzr:
If it helps, I can share an example to test using webthing-node (currently using webthing-iotjs)

[](https://twitter.com/RzrFreeFr/status/1028008260253745152#artik710.png "#ARTIK710 WebThing")

static/js/components/property/action.js
```diff
@@ -77,20 +79,18 @@ template.innerHTML = `
       display: inline-block;
     }
   </style>
-  <div id="container" class="webthing-action-container">
-    <div id="contents" class="webthing-action-contents">
-      <button id="button" type="button" class="webthing-action-button"></button>
+  <div id="container-${BaseComponent.count}" class="webthing-action-container">
```
hobinjk: Since all of these elements are retrieved using `this.shadowRoot.querySelector` I believe a better way of solving the ID conflicts is to use querySelector with the class instead of the id and drop the id altogether.

static/js/components/property/action.js
```diff
@@ -77,20 +79,18 @@ template.innerHTML = `
...
+  <div id="name-${BaseComponent.count}" class="webthing-action-name"></div>
 `;
 
-class Action extends BaseComponent {
-  constructor() {
     super(template);
 
-    this._button = this.shadowRoot.querySelector('#button');
-    this._name = this.shadowRoot.querySelector('#name');
+    this._button = this.shadowRoot.querySelector(`#button${this.idSuffix}`);
```
hobinjk: e.g. this becomes `this.shadowRoot.querySelector('.webthing-action-button')`.

static/js/components/property/switch.js
```diff
@@ -80,27 +82,25 @@ template.innerHTML = `
...
-  <div id="container" class="webthing-switch-property-container">
-    <div id="contents" class="webthing-switch-property-contents">
+  <div id="container-${BaseComponent.count}" class="webthing-switch-property-container">
+    <div id="contents-${BaseComponent.count}" class="webthing-switch-property-contents">
       <form>
-        <input type="checkbox" id="switch"
+        <input type="checkbox" id="switch-${BaseComponent.count}"
           class="webthing-switch-property-switch">
-        <label id="slider" for="switch" class="webthing-switch-property-slider">
+        <label id="slider-${BaseComponent.count}" for="switch-${BaseComponent.count}" class="webthing-switch-property-slider">
```
hobinjk: Here the input might need to be nested within the label to allow the dropping of the `for` property which would otherwise be subject to an id conflict.

rzr: yes the next patch

rzr: Got your point but I fear we still need IDs : see comments at bottom of : https://stackoverflow.com/questions/41811551/custom-styling-for-checkbox-with-input-within-label

I have updated the patch to make it smaller and easier to review, once OK I will generalize this to other widgets.

rzr:
Feedback welcome, I plan to add changes for sliders soon

PR #1263, opened 2018-08-10
Switch from gcc 4.9 to 4.8 since 4.8 exists

### LINKED ISSUE #1262: Travis failure: missing gcc 4.9
All travis builds are failing due to gcc 4.9 missing from this PPA https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test even though the list on that page has gcc-4.9 on it.

Running on my local ubuntu installation the gcc-4.9 package is also missing from that PPA so this is rather strange all things considered.

hobinjk:
Fix #1262, which may be only an intermittent issue anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
-      - gcc-4.9
-      - g++-4.9
+      - gcc-4.8
+      - g++-4.8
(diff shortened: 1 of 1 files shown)


hobinjk:
Travis is just exploding, 🤷‍♂️ hopefully it will resolve itself

hobinjk:
And just like that, Travis works again

Issue #1245, opened 2018-08-02
Changing local domain redirects to broken URL


benfrancis:
STR:
* Settings -> Domain
* Change gateway.local to gateway2.local
* Click Update host name

Expected:
* Redirects to http://gateway3.local (or it could just show a message saying it succeeded)

Actual:
* Redirects to https://gateway3.local:4443 which shows a security exception

The security exception isn't a huge deal but a) Not sure why it adds port 4443 and b) It would be better if it could just navigate to http://gateway3.local using plain HTTP or just not navigate anywhere given it looks like an error message.

PR #1282, opened 2018-08-20
Disable redirect at changing local domain.

sogaani:
Fixes https://github.com/mozilla-iot/gateway/issues/1245.

This PR disable redirect rather than redirect http, because a) the user might change local domain outside of local network and b) enabling http access may have some issue. I think we should access gateway with https from outside of local network. However, restricting http access only from local network have to use modifiable headers like `Host` or `x-forwarded-for`. There are possibility to have security issue.

image

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -186,6 +186,9 @@ const SettingsScreen = {
     addDomainLocalButton.addEventListener(
       'click', this.onLocalDomainClick.bind(this));
 
+    localDomainName.addEventListener(
+      'input', this.onLocalDomainNameChange.bind(this));
+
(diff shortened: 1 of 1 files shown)


benfrancis:
Sorry I didn't get to this earlier. It might be a good idea to make the background-color of the button transparent in the disabled state, so it doesn't still look like a clickable button.

Alternatively we could use a toast to confirm that the change has taken place rather than changing the button into a non-interactive element.

benfrancis:
Thank you @sogaani 

I would suggest *not* making the toast clickable, because the notification disappears after a timeout so doesn't make a great touch target.

It could just say update succeeded.

PR #1285, opened 2018-08-22
[tiny] enable icon on android and iOS

Swaagie:
When pinning the things gateway to your mobile start screen this will make both phone OS to use the icon.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -10,6 +10,7 @@
     <title>Things Gateway</title>
     <link rel="manifest" href="/app.webmanifest">
     <link rel="icon" href="/optimized-images/icon.png" type="image/png" />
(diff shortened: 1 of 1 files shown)


hobinjk:
@benfrancis This LGTM but I'm flagging you for review since you have opinions about the mobile app ecosystem :)

benfrancis:
@Swaagie You should see the icon. How are you adding to home screen and what browser and version of the OS are you using?

Chrome, Firefox and Samsung Internet all support Progressive Web Apps with an icon in the manifest which is used when adding to homescreen.

But even if you just add a link to the page to your homescreen it should pick up the favicon.

Swaagie:
@benfrancis I'm on a lingeageOS@14 with latest Firefox. Just adding a link to the homepage, I'll try from Chrome as well when I'm back home

PR #1286, opened 2018-08-22
build: Lock NPM dependencies for npm install

rzr:
I observed a regression caused to dependencies updates,
build is file but you can't use UI to add things
(browser complains on module.export = App)
Probably caused by babel update from beta49 to rc2 (TBC).

Project prefers yarn over npm,
but npm install could be used by developers or scripts.

Note that, Adding a (potentially unaligned) lock file,
could create ambiguity
but IMHO it's better to support both than only yarn.

Related links:

https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files

https://github.com/yarnpkg/yarn/issues/5654#issuecomment-414960279

Change-Id: I9489e365b6d3a70102a7d2ae1e44feb7aeb28c06

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,18248 @@
+{
+  "name": "things-gateway",
+  "version": "0.5.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0-beta.49",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.49.tgz",
+      "integrity": "sha1-vs2AVIJzREDJ0TfkbXc0DmTX9Rs=",
+      "requires": {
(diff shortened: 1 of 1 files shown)


Swaagie:
Imho committing both lockfiles until `npm` and `yarn` settle on a solution would improve developer experience. Everybody benefits from having both. If you don't like big `git diff`s something we've been doing lately is adding a `.gitattributes` file with the following content:

```
package-lock.json binary
```

It will prevent git/github from showing the diffs. In addition the `yarn.lock` file could be added as well. If you still want to see the diffs (after updating a specific package) `git diff --text` will force show the diff.

hobinjk:
Thanks for looking into this! I think in the future we want to switch to only NPM since yarn doesn't really have enough benefits to justify installing yet another binary on the pi. This is great step in that direction.

Swaagie:
@rzr I can take a look at it tomorrow. I've spend way to much time in webpack and babel configs... 😛

PR #1291, opened 2018-08-23
Fix Settings scrolling

### LINKED ISSUE #1290: Settings scrolling issues
I've noticed some bugs which are probably regressions from #1270.

STR:
* Enter responsive mode with a small screen size (e.g. 320x480)
* Go to Settings
* Scroll down
* Select an option

Expected:
* The relevant setting screen opens in the start position (not scrolled)

Actual:
* The UI elements are scrolled off the screen
* In some cases the + button is not fixed at the bottom of the screen

Basically I think each setting needs to be a separate scrollable element. I'm pretty sure this used to work so I'm not sure what we broke :(

hobinjk:
Moves settings titles from a shared element to a
section-by-section element, fixing setting section titles while
maintaining scroll-with-content behavior.

Fix #1290

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -63,15 +63,6 @@
 
     <!-- Settings View -->
     <section data-view="settings" id="settings-view" class="hidden">
-      <div id="section-title" class="hidden">
-        <div id="section-title-back-flex"></div>
-        <div id="section-title-container">
-          <img id="section-title-icon" alt="Section Title" src="#" />
-          <span id="section-title-name"></span>
-        </div>
-        <div id="section-title-speech-flex"></div>
-      </div>
-
       <nav id="settings-menu" class="hidden settings-section">
         <ul>
           <li class="settings-item"><a id="domain-settings-link" href="/settings/domain">Domain</a></li>
@@ -132,7 +123,6 @@
         <section id="user-settings-main" class="hidden settings-subsection">
           <ul id="users-list">
           </ul>
-          <button id="add-user-button" class="icon-button"></button>
         </section>
         <section id="user-settings-edit" class="hidden settings-subsection">
           <form id="edit-user-form" class="user-form">
@@ -177,7 +167,6 @@
         <section id="addon-main-settings" class="hidden settings-subsection">
           <ul id="installed-addons-list">
           </ul>
-          <button id="discover-addons-button" class="icon-button"></button>
         </section>
         <section id="addon-config-settings" class="hidden settings-subsection">
         </section>
@@ -236,6 +225,8 @@
         </ul>
       </section>
       <a href="/settings" id="settings-back-button" class="hidden back-button icon-button"></a>
+      <button id="discover-addons-button" class="icon-button hidden"></button>
+      <button id="add-user-button" class="icon-button hidden"></button>
     </section>
 
     <!-- Rules View -->
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -29,15 +29,13 @@ const SettingsScreen = {
    */
   init: function() {
     this.view = document.getElementById('settings-view');
-    this.titleElement = document.getElementById('section-title');
-    this.titleName = document.getElementById('section-title-name');
-    this.titleIcon = document.getElementById('section-title-icon');
     this.menu = document.getElementById('settings-menu');
     this.domainSettings = document.getElementById('domain-settings');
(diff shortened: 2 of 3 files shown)


static/js/settings.js
```diff
@@ -55,14 +55,70 @@ const SettingsScreen = {
...
+                            '/optimized-images/add-on.svg');
+    this.insertTitleElement(this.experimentSettings, 'Experiments',
+                            '/optimized-images/experiments-icon.png');
+    this.insertTitleElement(this.updateSettings, 'Updates',
+                            '/optimized-images/update-icon.svg');
+    this.insertTitleElement(this.authorizationSettings, 'Authorizations',
+                            '/optimized-images/authorization.svg');
+    this.insertTitleElement(this.developerSettings, 'Developer',
+                            '/optimized-images/developer-icon.svg');
+
```
benfrancis: Is there are particular reason these headings need to be added dynamically rather than being part of the original HTML?

static/js/settings.js
```diff
@@ -55,14 +55,70 @@ const SettingsScreen = {
...
+  insertTitleElement: function(section, name, icon) {
+    const elt = document.createElement('div');
+    elt.classList.add('section-title');
+    elt.innerHTML = `
+      <div class="section-title-back-flex"></div>
+      <div class="section-title-container">
+        <img class="section-title-icon" alt="Section Title" src="${icon}" />
+        <span class="section-title-name">${name}</span>
+      </div>
+      <div class="section-title-speech-flex"></div>
```
benfrancis: I realise this is the HTML that was already in use (I hadn't noticed before), but is it possible to centre align this title without using empty these empty divs? The flex approach is working brilliantly in some cases.

screen shot 2018-08-24 at 12 33 13

Ideally this would be a more semantically meaningful HTML tag like h2, centered with text-align or margin: auto. Perhaps with a width of 100% and padding to account for the icons.

static/index.html
```diff
@@ -63,15 +63,6 @@
...
     <section data-view="settings" id="settings-view" class="hidden">
-      <div id="section-title" class="hidden">
-        <div id="section-title-back-flex"></div>
-        <div id="section-title-container">
-          <img id="section-title-icon" alt="Section Title" src="#" />
-          <span id="section-title-name"></span>
-        </div>
-        <div id="section-title-speech-flex"></div>
-      </div>
-
```
benfrancis: Now that this HTML is part of the layout of individual settings screens, everything has been shifted downwards.

screen shot 2018-08-24 at 12 42 02

screen shot 2018-08-24 at 12 41 46

screen shot 2018-08-24 at 12 41 38

static/js/settings.js
```diff
@@ -55,14 +55,70 @@ const SettingsScreen = {
...
+    this.insertTitleElement(this.domainSettings, 'Domain',
+                            '/optimized-images/domain-icon.png');
+    this.insertTitleElement(this.userSettingsMain, 'Users',
+                            '/optimized-images/users-icon.png');
+    this.insertTitleElement(this.userSettingsEdit, 'Edit User',
+                            '/optimized-images/user.svg');
+    this.insertTitleElement(this.userSettingsAdd, 'Add User',
+                            '/optimized-images/user.svg');
+    this.insertTitleElement(this.adapterSettings, 'Adapters',
+                            '/optimized-image/adapters-icon.png');
```
benfrancis: Missing icon?

screen shot 2018-08-24 at 12 44 43

FWIW "section title" is not very helpful alt text!

static/js/settings.js
hobinjk: Adding them to the original HTML would have been at least an hour worth of copy-pasting markup that we may want to change in the future (going to go into more detail about that below).

static/js/settings.js
hobinjk: I took the flex approach originally after considering all other easier alternatives. The key thing here is that at significantly limited screen widths (like 240px) the icon can overlap with the back button. There's an unrelated issue here which is making the approach look less clever which is that settings-section has a 1.5rem padding not taken into account by the title.

If we do want to modify this approach in the future, a change to insertTitleElement is just a couple of keystrokes away. The complexity of the current solution is why I avoided hard-coding all the HTML for the title element as it would make further reconsideration of the design difficult.

static/js/settings.js
hobinjk: Updated accordingly

PR #1298, opened 2018-08-24
Fix overlapping setting-title and UL elements.

sogaani:
After https://github.com/mozilla-iot/gateway/pull/1291, some setting page elements are overlapped.
image

### DIFF
--- a/static/css/settings.css
+++ b/static/css/settings.css
@@ -219,6 +219,7 @@
   width: 100%;
   max-width: 60rem;
   padding: 2rem;
+  padding-top: 9.6rem;
   max-height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


static/css/settings.css
```diff
@@ -219,6 +219,7 @@
   width: 100%;
   max-width: 60rem;
   padding: 2rem;
+  padding-top: 9.6rem;
```
hobinjk: I think for them to remain vertically centered this should be padding: 9.6rem 2rem;

hobinjk: (i.e. also pad the bottom)

hobinjk: Oh this is weird, the user-settings ul is fine but the addon-settings ul is wrong. I'm guessing this will flip the two and make the user-settings ul too padded and the addon-settings ul good.

hobinjk: Ah, user-settings was also wrong, it just looked correct since it was shorter and triggered a min-height somewhere

Issue #1041, opened 2018-05-29
Verify visual consistency across revisions


hobinjk:
Take the screenshots automated ui testing generates and put them through https://github.com/mapbox/pixelmatch or https://github.com/gemini-testing/looks-same

PR #1303, opened 2018-08-26
Compare screenshot on browser tests.

sogaani:
Fixes https://github.com/mozilla-iot/gateway/issues/1041

* How this work.
    * If a test calls `saveStepScreen`, it store screenshot in `/browser-test-output`.
    * After a test which calls `saveStepScreen` is done, compare screenshots taken with screenshots in `/src/test/browser/screenshots`.
    * If there is difference, store diff image in `browser-test-diff` and change test status to pending. 
    * If there is new screenshots, change test status to pending.
    * we should check diff images and a screenshots taken. If screenshots taken is correct, copy those to `/src/test/browser/screenshots` manually and retry tests.

* Issue
    * html element `input type=range` is disappeared in a screenshot with headless Firefox and webdriverio.


hobinjk:
This is awesome! It will be really great to have this working.

One minor issue is that when running this locally the first set of tests (things-test, I think) failed but the testing process moved on to addon-test and deleted all of the diffs and output screenshots so I couldn't see what was happening in things-test.

I think it might be good to have a flag for the test running process which disables screenshot comparison since it takes a while and is sensitive to OS and browser version. For example, the cause of the failures I'm seeing is a bunch of 1px shifts caused by minor differences in my environment.

src/test/browser/browser-common.js
```diff
@@ -111,6 +114,9 @@ afterEach(async () => {
 });
 
 async function saveStepScreen(stepStr) {
+  if (process.env.DISABLE_SCREENSHOT_COMPARISON) {
```
hobinjk: I think this check can be removed so that we can still see the output screenshots even if we don't want to compare them.

hobinjk:
It looks like this should mostly work once it's rebased on master. I'm not sure why the virtual things screenshots aren't matching.

benfrancis:
First, thank you for this contribution.

This involves checking in quite a lot of binary files to GitHub and will continue to increase the size of the git checkout as screenshots are updated. Just to check, this is only for testing and won't affect image size on the gateway, right?

hobinjk:
As currently implemented these would be included in the image/in OTA bundles, adding 2 MB. It would be doable to manually drop the images directory but it would need some planning.

sogaani:
@benfrancis As currently the Gateway often changes UI, git history will increase fast. We can avoid downloading all history with `git clone --depth 1` command which download only latest commit. However, large size of git history might affect us to develop in the future.

Another idea is comparing with `current branch` and `master branch` or `base of an current branch`. As we develop on `master branch` and it should be right, we have only to check different between `master branch` and `current branch`.
An issue about this idea is that if `master branch` have some issue about UI, we cannot find it on test.

Which one do you think better?

benfrancis:
I wonder if we can make all these binary files part of our test infrastructure rather than being checked into the git repository every time they change? Git is not well suited to storing lots of large binary files.

The screenshots definitely shouldn't make it into the production image which gets flashed onto gateway hardware. I think we need to figure out how to generate a production image which doesn't include any development dependencies.

I hope we can get this figured out because more automated testing of the front end is definitely needed.

sogaani:
> I wonder if we can make all these binary files part of our test infrastructure rather than being checked into the git repository every time they change?

Yes, I don't think it is nesesary that screenshots check into the git repository. It is enough to compare the screenshots taken from `master branch` and the screenshots taken from `current branch`. I will work on it tomorrow.

mrstegeman:
If necessary, we could probably set up a separate repo with git-lfs for storing screenshots and such.

mrstegeman:
@sogaani Could you rebase, please?

sogaani:
Rebased commit.

How to compare UI with parent branch:
* Step1: Take screenshots on parent branch with follwing command.
    ```
    $ npm run screenshots
    ```
    This command switch parent branch and take screenshots into the folder `./browser-test-screenshots`.

* Step2: Run test
    ```
    $ npm test
    ```
    If you have the screenshots in the folder `./browser-test-screenshots`, `npm test` should compare UI with screenshots stored.

This PR only work on a individual local workspace.
I'd like to enable to this future in travis. However, we need the place where upoload screenhots.

.gitignore
```diff
@@ -10,10 +10,17 @@ tmp/
 /build
 node_modules/
 static/js/lib/stm_web.min.js
+package-lock.json
```
mrstegeman: I think we're now keeping `package-lock.json` in the repo for people not using `yarn`.

sogaani:
I'm sorry, I have misunderstood.
I think that I make a wrong commit. I'm going to fix it soon.

benfrancis:
What's the status on this increasing the size of the production image?

sogaani:
Currently, this PR does not increase the image size, but just store images to user local storage.

sogaani:
ok, I will rebase this weekend.

hobinjk:
Just fixed the last tiny merge conflict, will be merging this once the tests pass

PR #1313, opened 2018-08-28
Disallow ns* subdomains.

mrstegeman:
Those are reserved for our nameservers.


mrstegeman:
@hobinjk Already [done](https://github.com/mozilla-iot/registration_server/commit/2f2c5beab70f90e4d98d3501082dc5a584816906#diff-e7e2d2cacabaf203c0f166c2c69cd57aR414)

PR #1314, opened 2018-08-28
Don't copy old gateway on upgrade.

### LINKED ISSUE #1140: OTA Update shouldn't copy over old gateway
It did this previously when we stored add-ons and the database in the gateway folder but right now this is just a wasted step.

mrstegeman:
Fixes #1140

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -20,12 +20,6 @@ wget $node_modules_archive_url
 
 gateway_old=./gateway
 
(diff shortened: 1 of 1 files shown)


hobinjk:
I have an existing branch which incorporates this and does some other speedups. Sorry for not making the #1315 PR earlier!

mrstegeman:
I just saw! I'll close this one.

PR #1315, opened 2018-08-28
Speed up OTA by deferring shutdown and removing copying

### LINKED ISSUE #761: OTA: Reduce downtime
Right now the OTA update stops the gateway much earlier than it needs to, In theory it can be moved back so far as to be almost seamless

### LINKED ISSUE #1140: OTA Update shouldn't copy over old gateway

hobinjk:
Fixes #761
Fixes #1140

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "things-gateway",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "description": "Web of Things gateway",
   "main": "app.js",
   "scripts": {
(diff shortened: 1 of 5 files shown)


hobinjk:
This works but will only take effect in the release after next due to a combination of an intended behavior with a bug. The OTA uses the old gateway's upgrade script because the new one's doesn't exist, but then it uses the old gateway's post-update script even when the new one's exists.

hobinjk:
:+1:

hobinjk:
Oh and in terms of performance the pre-phase takes 6 minutes during which the gateway is responsive and can perform normal functions then the OTA restarts and upgrades the gateway which takes 2 minutes of downtime.

PR #1317, opened 2018-08-29
Use new intent parser message structure.

mrstegeman:
Requires mozilla-iot/intent-parser#6


mrstegeman:
@hobinjk This should be good now. The failures are because the intent-parser change is not yet merged.

mrstegeman:
Now we just have the failing UI test from #1338

PR #1320, opened 2018-08-30
Refresh add-on list on entering add-on settings.

mrstegeman:


mrstegeman:
That's a good question. It used to be refreshed every time the list was shown. However, it's now cached and is only fetched once while the things UI is open. We could either do it this way or revert to the old behavior.

dhylands:
Is it possible to create a list hash and retrieve that every time and only update the list when the hash changes?

mrstegeman:
Yes, that's another option as well.

mrstegeman:
@hobinjk I changed the commit to essentially revert to old behavior. The add-on list is always updated when the user enters _Settings->Add-ons_, and updated only if necessary when going into an add-on config screen or the discovered add-ons screen, i.e. if they navigate directly there in a new tab for some reason.

PR #1323, opened 2018-08-31
Add titles to Assistant and Things screens.

### LINKED ISSUE #1211: Add page titles to Assistant and Things screens

mrstegeman:
Fixes #1211

### DIFF
--- a/static/css/assistant.css
+++ b/static/css/assistant.css
@@ -5,6 +5,27 @@
   display: none;
 }
 
+#assistant-view > .title-bar {
+  text-align: center;
+  line-height: 9.6rem;
+  height: 9.6rem;
+}
+
+#assistant-header {
+  margin: 0 auto;
+  display: inline-block;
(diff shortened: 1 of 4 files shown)


mrstegeman:
@benfrancis Is this something we want, or no?

benfrancis:
Hmm, consistency vs. clutter.

I think I agree with James, on reflection this makes the things screen seem cluttered. Same with the assistant screen.

I suggest we leave this as it is for now. Sorry for the churn.

mrstegeman:
No problem. I'll close the issue.

PR #1324, opened 2018-08-31
Reopen websocket on close/error.

### LINKED ISSUE #1216: If gateway restarts, websockets wind up being stale
If the gateway is restarted then all of the websockets that the UI is currently holding become stale. and the UI stops reflecting changes to devices.

The UI looks like its still working, because it can still control devices, but changes stop being reflected until refreshing the screen.

mrstegeman:
Fixes #1216


mrstegeman:
@hobinjk Fixed.

PR #1326, opened 2018-08-31
Fix settings subsection scrolling on Chrome

hobinjk:


static/css/settings.css
```diff
@@ -165,6 +165,7 @@
 
 .settings-subsection {
   display: block;
+  overflow-y: scroll;
```
mrstegeman: Why not `auto`?

hobinjk: I thought that scroll was more explicit about what we want it to do but given that a couple lines above it says `auto` I've changed it to be consistent

mrstegeman: `scroll` keeps a persistent scroll bar in a lot of browsers, regardless of whether or not the page is even scrollable, which is annoying.

PR #1337, opened 2018-09-05
V0.5.2 with fast ota

hobinjk:
I also dropped the version bump to 0.5.2 that was erroneously included in the OTA commits.


hobinjk:
Yes, but I figured it could get a better commit message than "fake version bump". I'll add a real one to this PR

PR #1352, opened 2018-09-13
Add notifications

### LINKED ISSUE #1295: Alerts
As a user I want to be notified when a door is opened or motion is detected in my home so that I can act on that information.

hobinjk:
Fixes #1295


hobinjk:
It should work! I'm not sure what it does if the to-be-notified device is asleep/unresponsive

hobinjk:
Yeah, the if/while thing is just a drawback of a different part of the rules engine that I can fix in a different PR

PR #1353, opened 2018-09-18
Send error to UI when add-on fails to start.

### LINKED ISSUE #877: Error reporting mechanism for add-ons
As an add-on developer I want to report an error to the end user when something goes wrong

mrstegeman:
Fixes #877

### DIFF
--- a/src/addon-loader.js
+++ b/src/addon-loader.js
@@ -22,6 +22,10 @@ const path = require('path');
 // Open the database.
 db.open();
 
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 async function loadAddon(addonPath, verbose) {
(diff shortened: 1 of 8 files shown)


mrstegeman:
@hobinjk It would probably be better to fix the individual add-on to do the right thing, which is use `errorCallback()`. What add-on is it?

The problem is that we use the errorCallback to get a proper error message over the IPC bridge, since this happens in a separate process. The errorCallback is supposed to be what kills the process with code 100.

hobinjk:
In my local case the add-on is crashing before it starts up because it happens to be missing a dependency. I think we should report these early fatal errors using the same mechanisms

mrstegeman:
@hobinjk I think I got it handled now. Can you give it a shot?

hobinjk:
Awesome! LGTM

PR #1354, opened 2018-09-18
Implement UI for read-only properties.

### LINKED ISSUE #1261: Read-only properties
As a user I would like to be able to tell whether a property can be modified or whether it is read-only information so that I don't try to set a read-only property

Depends on https://github.com/mozilla-iot/wot/issues/38

mrstegeman:
Fixes #1261


hobinjk:
There appear to be lint errors in this according to Travis

PR #1356, opened 2018-09-19
Cache more things with ServiceWorker.

### LINKED ISSUE #727: Inform user when gateway can't be reached
As a user I want to be informed when my browser can't reach my gateway so I know what's going wrong.

There have been quite a few reported problems where the gateway can not be reached by the cached web interface and it's unclear what has gone wrong. Either the front end loads but can't get API responses or some unhelpful error page is displayed.

It would be great if the front end cached by a Service Worker could somehow detect when the gateway can't be reached and display a more helpful error message. It can actually be quite tricky to tell whether the gateway is actually unreachable or connectivity is just bad, but perhaps there's something we can do.

mrstegeman:
Fixes #727

### DIFF
--- a/static/service-worker.js
+++ b/static/service-worker.js
@@ -32,7 +32,7 @@ self.addEventListener('fetch', function(event) {
 
   event.respondWith((async () => {
     const cache = await caches.open(CACHE);
-    const matching = await cache.match(event.request);
+    const matching = await cache.match(event.request, {ignoreVary: true});
     if (matching) {
       return matching;
     }
--- a/static/service-worker.js
+++ b/static/service-worker.js
@@ -3,7 +3,6 @@
 
 // eslint-disable-next-line no-undef
 const CACHE = `mozilla-iot-cache-${VERSION}`;
-const CACHE_PATH = /^\/bundle|^\/images|^\/optimized-images/;
 
 self.addEventListener('activate', (event) => {
   event.waitUntil(
@@ -22,24 +21,25 @@ self.addEventListener('activate', (event) => {
 self.addEventListener('fetch', function(event) {
   const accept = event.request.headers.get('Accept');
   if (accept === 'application/json' || event.request.method !== 'GET' ||
-      event.request.mode !== 'no-cors') {
+      !['no-cors', 'navigate'].includes(event.request.mode)) {
     return;
   }
+
   const url = new URL(event.request.url);
-  if (url.origin !== location.origin ||
-      !url.pathname.match(CACHE_PATH)) {
+  if (url.origin !== location.origin || url.pathname.endsWith('.map')) {
     return;
   }
+
   event.respondWith((async () => {
     const cache = await caches.open(CACHE);
     const matching = await cache.match(event.request);
     if (matching) {
       return matching;
-    } else {
-      const response = await fetch(event.request);
-      await cache.put(event.request, response.clone());
-      return response;
     }
+
+    const response = await fetch(event.request);
+    await cache.put(event.request, response.clone());
+    return response;
   })());
 });
 
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -80,13 +80,16 @@ const App = {
     this.blockMessages = false;
     this.messageArea = document.getElementById('message-area');
     this.messageTimeout = null;
+
+    this.connectivityOverlay = document.getElementById('connectivity-scrim');
+    this.pingerInterval = null;
+    this.pingerLastStatus = null;
+    this.startPinger();
+
     this.gatewayModel = new GatewayModel();
 
     this.wsBackoff = 1000;
     this.initWebSocket();
-
-    this.connectivityOverlay = document.getElementById('connectivity-scrim');
-    this.startPinger();
 
     Menu.init();
     Router.init();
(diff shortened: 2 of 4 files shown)


sogaani:
I guess we should return cached `index.html` only when request for `index.html` were failed. Because service worker does not update as soon as rebuilding service worker. It means that service worker would return old version `index.html` when rebuild or update the Gateway.
Also, other url which return `index.html`(e.g /thing /setting ....) have same issue.

static/service-worker.js
```diff
@@ -22,24 +21,25 @@ self.addEventListener('activate', (event) => {
 self.addEventListener('fetch', function(event) {
   const accept = event.request.headers.get('Accept');
   if (accept === 'application/json' || event.request.method !== 'GET' ||
-      event.request.mode !== 'no-cors') {
+      !['no-cors', 'navigate'].includes(event.request.mode)) {
```
sogaani: Does this PR still intend to always return cached 'index.html'？
If the browser requests `index.html`, `event.request.mode` would be `'navigate'`. This PR seems to cache 'index.html' and return cache if cached.

static/service-worker.js
```diff
@@ -22,24 +21,25 @@ self.addEventListener('activate', (event) => {
...
   }
+
   const url = new URL(event.request.url);
-  if (url.origin !== location.origin ||
-      !url.pathname.match(CACHE_PATH)) {
+  if (url.origin !== location.origin || url.pathname.endsWith('.map')) {
     return;
   }
+
   event.respondWith((async () => {
```
sogaani: If return cached `index.html` only when request failed, Following code works for me.
```
  event.respondWith((async () => {
    const cache = await caches.open(CACHE);
    if (event.request.mode !== 'navigate') {
      const matching = await cache.match(event.request);
      if (matching) {
        return matching;
      }
    }

    try {
      const response = await fetch(event.request);
      await cache.put(event.request, response.clone());
      return response;
    } catch (e) {
      return cache.match(event.request or cached 'index.html');
    }
  })());
```

static/service-worker.js
mrstegeman: No, it will return whatever the initial request was for. This PR changes what is actually cached, such that HTML, CSS, etc. are also cached, allowing us to serve up the entire page when the user is offline.

mrstegeman: Ah, I see what you mean. Yes, that is a problem during development, but the entire cache is invalidated with every single build during development, so that's not specific to the index.html case. In a release build, the cache version will remain constant until an update.

For testing purposes, I made the cache version constant by changing `VERSION` in the webpack config.

sogaani: > so that's not specific to the index.html case.

Maybe yes. Currently, almost build artifacts have version prefix(change per build) and it referenced from `index.html`, so that with refreshed `index.html` build artifacts autoimatically would be refreshed althought old version are chached.

However, it's OK. I was just wondering.
I understand that sometime we should force-reload to browse new UI when update or rebuild the Gateway.

hobinjk:
Testing this locally I'm still getting "An error occurred during a connection to hobinjk.mozilla-iot.org. Peer received a valid certificate, but access was denied. Error code: SSL_ERROR_ACCESS_DENIED_ALERT" when the gateway is down and I try to reload the page.

hobinjk:
(Firefox, valid ssl cert through tunnel, also happens with invalid cert)

mrstegeman:
What I had to do was:
1. Change the VERSION string to something constant in the webpack config
2. Force refresh a couple times (to make sure the Service worker was updated)
3. Soft refresh a couple times (to make sure content was cached)
4. Kill gateway
5. Refresh

hobinjk:
I followed those steps and still get the Secure Connection Failed error in Firefox and "This site can’t be reached
The webpage at https://hobinjk.mozilla-iot.org/things might be temporarily down or it may have moved permanently to a new web address.
ERR_FAILED" in chrome

hobinjk:
The chrome error has this additional console information
```
An unknown error occurred when fetching the script.
Failed to load resource: net::ERR_SSL_PROTOCOL_ERROR
```

mrstegeman:
You could also try unregistering the old service workers. I was doing that during development as well

hobinjk:
I've tried that and confirmed that the running serviceworker has the correct source according to dev tools. The serviceworker did reload the page after the gateway was done restarting, it just isn't serving index.html

mrstegeman:
Shoot. What I ended up with was the blank things page with the disconnected overlay.

hobinjk:
Hmm, I'll check out net-internals/serviceworker-internals next week to see if I can figure out why it's doing this

sogaani:
@hobinjk Did you reload twice or more before stop the Gateway? Service worker cache artifacts only after it installed. The first time access to the Gateway install service worker in `app.js` then the browsr already get `index.html. So srvice worker does not cache `index.html` at first time access.

mrstegeman:
@sogaani Did you get it to work?

sogaani:
@mrstegeman Yes, it works on Windows10 and latest firefox.

image

mrstegeman:
Yay, I didn't imagine it!!

sogaani:
It does not wok with Chrome.
Chrome does not match root directory request(/things or /settings).
I found defferent between `/things` and other artifactacts in `Vary header`.
Only response to `/things` have `Vary: Accept` header.
`cache.mattch()` seems to consider Vary header. See also https://developer.mozilla.org/en-US/docs/Web/API/Cache/match#Parameters
I try cache matching with ignoreVary option and it work to me.

sogaani:
It works fine for me,
I tested following enviroments.

|OS  |Browser  |`ignoreVary` |offline message  |
|---|---|---|---|
|Windows 10  |Firefox  |false |work |
|Windows 10  |Chrome  |false |not work |
|Windows 10  |Edge |false |work [1] |
|Mac  |Firefox  |false |not work |
|Mac  |Chrome  |false |not work |
|Mac  |Safari |false |false |work [1] |
|Windows 10  |Firefox  |true |work |
|Windows 10  |Chrome  |true |work |
|Windows 10  |Edge |true |work [1] |
|Mac  |Firefox  |true |work [2] |
|Mac  |Chrome  |true |work [2] |
|Mac  |Safari |true |work [1] |

[1] : it works but screm disappear.
[2] : it take several tens of seconds until showing geteway offline message. Some http request seems to block executing scripts.

mrstegeman:
Wow, thanks for all the testing!

PR #1359, opened 2018-09-20
subscription should be unique

sogaani:
Currently, we get duplicated notifications as much as reloaded times.


hobinjk:
Thanks for finding this!

PR #1367, opened 2018-09-24
Add better property value validation.

mrstegeman:


src/controllers/things_controller.js
```diff
@@ -232,17 +232,51 @@ ThingsController.put(
...
-        response.status(500).send(error);
+    Things.getThingDescription(thingId, request.get('Host'), request.secure)
+      .then((thing) => {
+        if (!thing.properties.hasOwnProperty(propertyName)) {
+          response.status(404).end();
+          return;
+        }
+
+        if (thing.properties[propertyName].readOnly) {
+          response.status(400).end();
```
hobinjk: I'd prefer if all of these error codes had an accompanying message so that a developer doesn't have to wonder why they're getting a 400

mrstegeman:
@hobinjk Updated.

PR #1368, opened 2018-09-24
Get rid of warning from dynamicRequire().

mrstegeman:
This seems to confuse a lot of people and make them think that
something is wrong, when everything is fine.

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -32,14 +32,17 @@ const pkg = require('../package.json');
 
 let PluginClient, PluginServer;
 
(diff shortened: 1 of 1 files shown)


hobinjk:
I still see the
```
    WARNING in ./src/addon-manager.js 43:11-18
    Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
     @ ./src/app.js
```
warning when running this. Is there a different warning this is targeting?

mrstegeman:
Ugh, yes. I was looking in the wrong part of the log.

Issue #1273, opened 2018-08-14
Proposal: production shouldn't run webpack at all


hobinjk:
When running in production go directly to `node build/gateway.js`, skipping webpack. On the Pi, only bundle the production `node_modules`

Pros:
- Reduces node_modules size significantly (currently 500M)
- Will boot almost instantly
Cons:
- Makes local development require running `yarn` and some kind of flag to get the systemd service out of production mode

hobinjk:
Another interesting approach is flattening all the dependencies which gets us down to ~250M but will likely cause bugs due to api changes between dependency versions

benfrancis:
This sounds right to me because I wouldn't have thought we should need to build on every boot, but I really don't know enough about webpack to know how it's intended to be used in production.

They do have documentation around production builds but we may already be doing a lot of this https://webpack.js.org/guides/production/

hobinjk:
Unfortunately, the main benefit of not running webpack at all is that we don't have to pull in all of the support modules that webpack needs to function. For example, just the top-level babel dependency we use to turn our fancy JS into JS that even IE could understand is 67MB. We can also take out the test support infrastructure and leave only the smallest set of runtime dependencies.

As far as I can tell the main benefit of production mode is that webpack runs additional optimizations passes on the output files which would have the opposite of the desired effect, but could be a nice way to eke out a couple more percentage points of performance.

mrstegeman:
I believe this is now done. See e66fe3b8136be1c99410d937fa1b7e90f0c0d285.

PR #1369, opened 2018-09-24
Modules toward webpack move to devDependencies.

sogaani:
Related to https://github.com/mozilla-iot/gateway/issues/1273


hobinjk:
Under what conditions would dependencies-check find an error that the normal automated testing wouldn't? I believe that any missing dependency in gateway.js would cause it to crash which would then cause `jest` to report a failure.

sogaani:
We install all dependencies including devDependencies while run jest. Although we wrong to install dependency to `devDependencies`, the Gateway code could use `devDependencies` in test enviroment.
`dependencies-check` would check if dependeincies which is installed by `yarn install --production` and only regards `dependencies`(not include devDependencies) in package.json satisfies to run `gateway.js`.

I guess that if the Gateway only bundle the production node_modules, we need this check.

PR #1370, opened 2018-09-24
ui: Select by class instead of (multiple) id

rzr:
In case of multiple property of same types,
UI events were forwarded to 1st instance of widget.

This change is a follow up of the switch change:

Bug: https://github.com/mozilla-iot/gateway/issues/1148
Relate-to: https://github.com/mozilla-iot/gateway/pull/1249
Change-Id: I2018092168af14eb8f6f1e9e230e04a432490045

### DIFF
--- a/src/test/browser/page-object/thing-detail-page.js
+++ b/src/test/browser/page-object/thing-detail-page.js
@@ -102,7 +102,7 @@ class BrightnessPropertySection extends InputPropertySection {
 class LevelPropertySection extends InputPropertySection {
   constructor(browser, rootElement) {
     super(browser, rootElement);
-    this.defineElement('slider', '#slider');
+    this.defineElement('slider', '#slider-level-6');
   }
 
   async setValue(value) {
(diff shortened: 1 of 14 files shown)


rzr:
Not really this was only needed for CSS rule for switch (for id).

I could use the counter for the slider test, but I guess it thought just add confusion ( count ~= 6 ) and just use the hardcoded one ("slider").

But I can also share an other next change that introduce the counter just for unification of style of templates.

mrstegeman:
@rzr It would make sense to me to be consistent. Can you push that change up to this PR?

src/test/browser/page-object/thing-detail-page.js
```diff
@@ -128,6 +128,7 @@ class LevelPropertySection extends InputPropertySection {
 
   async getValue() {
     const slider = await this.slider();
+    console.log(slider);
```
mrstegeman: Remove log.

PR #1420, opened 2018-10-17
Implement log rotation.

### LINKED ISSUE #1417: We should store (and be able to retrieve) a "previous log"
The gateway currently stores the log for the current run, and the UI has the ability to retrieve this log.

However, if the gateway restarts itself, the cause of the restart will be the last thing in the log and will immediately be overwritten the next time the gateway starts.

I'd like to see run-app.sh modified to keep a previous log and have the UI give a way to retrieve this.

The current user study would benefit from this capability.

mrstegeman:
Fixes #1417

### DIFF
--- a/src/log-timestamps.js
+++ b/src/log-timestamps.js
@@ -115,16 +115,7 @@ if (!console.constructor.hooked) {
 
     for (const func of FUNCS) {
       console[func] = function() {
-        const args = Array.from(arguments);
-        let message = args[0];
-        for (const arg of args.slice(1)) {
-          if (arg.hasOwnProperty('toString')) {
-            message += ` ${arg.toString()}`;
-          } else {
-            message += ` ${arg}`;
-          }
-        }
-        logger[func](message);
+        logger[func](format.apply(null, arguments));
       };
     }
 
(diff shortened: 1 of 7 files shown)


mrstegeman:
@rzr This is much more cross-platform friendly and will work in cases where systemd is not available, like OpenWRT, Windows, macOS, etc.

rzr:
OK it makes sense, so I guess if one want a "linux" profile of gateway, it should be branched and slimmed down ?

mrstegeman:
@rzr That's an option.

mrstegeman:
Rebased.

src/log-timestamps.js
```diff
@@ -84,13 +111,23 @@ if (!console.constructor.hooked) {
...
   } else {
     // This path is for the normal non-jest output
-    const FUNCS = ['log', 'info', 'debug', 'error', 'warn'];
+    const FUNCS = ['info', 'debug', 'error', 'warn', 'verbose', 'silly'];
 
     for (const func of FUNCS) {
-      const realFunc = console[func];
       console[func] = function() {
-        realFunc(logPrefix() + format.apply(null, arguments));
+        const args = Array.from(arguments);
```
hobinjk: This code should be replaced with `format.apply(null, arguments)` which handles objects better and is consistent with the previous code.

mrstegeman: Done.

PR #1421, opened 2018-10-17
Switch EventTrigger to momentary pulse; Retry subscribe on Thing connection

hobinjk:


src/rules-engine/triggers/EventTrigger.js
```diff
@@ -36,6 +38,10 @@ class EventTrigger extends Trigger {
 
   async start() {
     await this.thingConn.start();
+    await this.subscribe();
+  }
+
+  async subscribe() {
```
mrstegeman: `this.subscribed` should probably be set in this function.

hobinjk: Setting this.subscribed in this function would either have no effect (`this.subscribed = false`, it's already false) or be incorrect (`this.subscribed = true`, the Thing model hasn't acknowledged the subscription request)

If there were a way for the Thing model to drop a subscription and require a transition from subscribed to unsubscribed then this would be a different story.

dhylands:
Seems to be working fine for me.

The only thing I noticed (unrelated to this PR) is that there seems to be a debug print here:
https://github.com/mozilla-iot/gateway/blob/2bcc7bf3da1be5e486aa8a387278154ca1d7bd8f/src/rules-engine/Property.js#L109

PR #1424, opened 2018-10-18
Log ping error.

mrstegeman:


hobinjk:
It might also be nice to require a series of ping failures to mark the gateway as disconnected while we're still working on figuring out why this issue is happening

PR #1427, opened 2018-10-19
set new enum-value on property

roderm:
- update enum on doPropertyChange


roderm:
### Bug noticed
if browser cache is disabled, list won't load.

Problem with my plugin or a general issue?

mrstegeman:
Seems unrelated to this change. Could you try disabling your add-on and try again? If you can reproduce, please open an issue so we can take a look.

By the way, thanks for the PR!

PR #1440, opened 2018-10-23
v0.6.1 Candidates

mrstegeman:


mrstegeman:
Not entirely sure why Travis CI is failing, `npm run test` works for me locally.

hobinjk:
Looks like every test that uses the browser is failing with a timeout or "A session id is required for this command but wasn't found in the response payload". This is likely related to the new Firefox release

mrstegeman:
Yep, I just realized the same thing. Working on a fix.

mrstegeman:
I thought that upgrading `webdriverio` and/or `selenium-standalone` would fix it, but apparently not.

PR #1443, opened 2018-10-24
Flash push button icon on events.

### LINKED ISSUE #1442: Tweak PushButton UI
We've found that not all PushButtons provide a PushedProperty.

Here's what I suggest:
* If no PushedProperty is provided, hide the "..."
* PressedEvent makes the icon flash once
* DoublePressed event makes the icon flash twice
* LongPressed event makes the icon flash for a longer period

Timings may require a bit of experimentation.

If we could get this into 0.6.1 that would be great for our MozFest demo, but I wouldn't hold the release on it.

mrstegeman:
Fixes #1442

### DIFF
--- a/static/js/components/capability/push-button.js
+++ b/static/js/components/capability/push-button.js
@@ -66,6 +66,8 @@ class PushButtonCapability extends BaseComponent {
     this._label = this.shadowRoot.querySelector('#label');
 
     this._pushed = false;
+    this._pressTimeout = null;
+    this.addEventListener('press', this._onPress.bind(this));
   }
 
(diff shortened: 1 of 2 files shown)


mrstegeman:
These are the same Travis failures as on #1440. Just due to webdriverio/selenium/??? compatibility issues with Firefox 63.

benfrancis:
Firefox 63 has Custom Elements and Shadow DOM turned on. Coincidence?

hobinjk:
The failing call appears to be `browser.url` which is very strange

PR #1454, opened 2018-10-26
Update (most) dependencies.

mrstegeman:


mrstegeman:
@hobinjk Most of these tests are failing due to `server.address()` returning `null`. Any idea what could be causing this? Websockets are failing to connect in the rules tests as well, which I'm guessing may be related.

hobinjk:
My guess is that `src/app.js` needs the serverStartup to be resolved later when the server is fully setup and listening (right now it's resolved when the server is configured)

It's likely a similar thing for the rules tests, but I'm less sure about that

mrstegeman:
@sogaani Any good suggestions for fixing the schema-form tests? I had to change the UI code to use ES6 modules (`import`/`export`), but `jsdom` doesn't seem to like that much.

src/controllers/oauth_controller.ts
```diff
@@ -175,8 +175,8 @@ function verifyClientAuthorization(client: ClientRegistry,
     return false;
   }
 
-  let clientId = decodeURIComponent(parts[0]);
-  let clientSecret = decodeURIComponent(parts[1]);
+  let clientId = decodeURIComponent(parts[0].replace(/\+/g, '%20'));
```
hobinjk: What behavior does this fix/prevent?

src/test/integration/oauth-test.js
```diff
@@ -103,11 +103,23 @@ describe('oauth/', function() {
...
+      options: {
+        authorizationMethod: 'body',
+      },
+      http: {
+        headers: {
+        },
       },
     }, configProvided || {});
 
+    const clientId = encodeURIComponent(config.client.id).replace(/%20/g, '+');
```
hobinjk: I'm pretty sure this isn't necessary since encodeURIComponent mostly exists as a way of allowing client ids and secrets to contain colons (if I remember correctly)

src/controllers/oauth_controller.ts
mrstegeman: The simple-oauth2 library changed the encoding of URL parameters. This was added to handle that.

src/test/integration/oauth-test.js
mrstegeman: See other comment. This was just done to match watch that library does.

hobinjk:
Another change that would be good is removing the redundant setting of NODE_TLS_REJECT_UNAUTHORIZED in src/test/common.js since it appears that you've switched to specifying it as an environment variable everywhere

mrstegeman:
@hobinjk done.

PR #1463, opened 2018-10-30
Fix error when trying to create duplicate push subscription.

mrstegeman:


src/db.js
```diff
@@ -538,11 +538,28 @@ const Database = {
    * @return {Promise<number>} resolves to sub id
    */
   createPushSubscription: function(desc) {
-    return this.run(
-      'INSERT INTO pushSubscriptions (subscription) VALUES (?)',
-      [JSON.stringify(desc)]
+    const description = JSON.stringify(desc);
```
hobinjk: If we want subscriptions to be unique then it should migrate the database to do a UNIQ constraint or filter everything manually since this code is subject to a race condition

hobinjk: Oh, I misinterpreted what the error you were referring to was, sorry.

PR #1473, opened 2018-11-02
Add --production to gateway build script

hobinjk:
Lets us actually take advantage of the benefits from splitting out dependencies and devDependencies. Reduces node_modules from 525M to 174M.

Will require a bunch of testing in 0.7 or 0.6.2

### DIFF
--- a/image/build-gateway.sh
+++ b/image/build-gateway.sh
@@ -78,7 +78,7 @@ nvm use ${NODE_VERSION}
 npm install -g yarn
 
 # Build the node modules, cross compiling any native code.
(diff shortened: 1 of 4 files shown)


mrstegeman:
Is this going to work? Right now, `run-app.sh` calls `npm start`, which in turn calls `webpack`. `webpack` is a dev dependency.

hobinjk:
Oh right, I forgot that we pre-build everything. That will need a small update then (I think the ota script might also do a similar thing)

hobinjk:
I believe this latest version should work as expected

mrstegeman:
#1476 fixes the Travis errors

PR #1476, opened 2018-11-05
Add filter params when requesting add-on list.

### LINKED ISSUE #1122: Support add-ons with node v10.
The add-on system will require some changes to support both node v8 and node v10.

mrstegeman:
Fixes #1122

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -27,6 +27,7 @@ const tar = require('tar');
 const os = require('os');
 const promisePipe = require('promisepipe');
 const fetch = require('node-fetch');
+const {URLSearchParams} = require('url');
 
 const pkg = require('../package.json');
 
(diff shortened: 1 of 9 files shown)


mrstegeman:
Not entirely sure why the UI test is failing... Everything works properly for me locally.

src/addon-manager.js
```diff
@@ -1095,36 +1096,42 @@ class AddonManager extends EventEmitter {
...
 
     try {
-      const response = await fetch(
-        config.get('addonManager.listUrl'),
-        {headers: {'User-Agent': `mozilla-iot-gateway/${pkg.version}`}});
+      const params = new URLSearchParams();
+      params.set('api', api);
+      params.set('arch', architecture);
+      params.set('version', version);
+      params.set('node', nodeVersion);
```
hobinjk: Should this also set the `test` param?

mrstegeman: Nope, I intentionally left that out here.

1. In test mode, this path is skipped anyway.
2. If someone manually sets the flag in normal (non-test) mode, it probably means they're (we're) testing some stuff, so we shouldn't overwrite test add-ons.

PR #1482, opened 2018-11-09
Use labels in rule display when available.

### LINKED ISSUE #1478: Property in rule engine use 'name' instead of 'label'
A Property has a `name` attribute and a description with a `label`. In Things the property is presented with the `label` but in Rules it is presented with the `name`.

Proposal: Use `label` also in the Rules.

mrstegeman:
Fixes #1478

### DIFF
--- a/src/rules-engine/Property.js
+++ b/src/rules-engine/Property.js
@@ -6,6 +6,7 @@
 
 const assert = require('assert');
 const fetch = require('node-fetch');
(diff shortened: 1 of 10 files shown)


mrstegeman:
There are a couple more changes incoming on this...

mrstegeman:
@hobinjk Should be good now.

PR #1490, opened 2018-11-13
Properly handle minimum/maximum/multipleOf in components.

mrstegeman:


mrstegeman:
@hobinjk Good catch! Fixed.

PR #1493, opened 2018-11-15
Add issuer field to JWT

### LINKED ISSUE #1492: JSON Web Tokens should include issuer field

hobinjk:
Fix #1492


hobinjk:
Thanks for pointing that out! I thought it would work normally but it did not

PR #1497, opened 2018-11-16
Handle enums in rules UI.

### LINKED ISSUE #1457: Rules engine shall handle enums in a better way
Enhancement request
A property with type = enum is possible to use in an input device but the possible values is not selected from a drop down list. Instead the user must type the enum value without any control if the value is an exising enum.

Expected result:
As input device
When a user select an enum properties the selected value should be selected from a drop down list.
As output device
The value to set should be select from a drop down list

(Out of scope of the this issue report but shall enums be handled in an ordered way so it is possible to use comperators < = or >)

mrstegeman:
Fixes #1457

### DIFF
--- a/static/js/rules/PropertySelect.js
+++ b/static/js/rules/PropertySelect.js
@@ -116,6 +116,7 @@ PropertySelect.prototype.addOption = function(name, ruleFragment, selected) {
         valueSelect.appendChild(option);
       }
 
+      valueSelect.addEventListener('click', stopPropagation);
       elt.appendChild(valueSelect);
     } else {
       valueInput = document.createElement('input');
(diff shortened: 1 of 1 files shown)


static/js/rules/PropertySelect.js
```diff
@@ -156,25 +181,47 @@ PropertySelect.prototype.addOption = function(name, ruleFragment, selected) {
...
+      valueSelect = document.createElement('select');
+      valueSelect.classList.add('value-select');
+
+      for (const choice of property.enum) {
+        const option = document.createElement('option');
+        option.value = choice;
+        option.innerText = choice;
+        valueSelect.appendChild(option);
+      }
+
```
hobinjk: valueSelect needs a stopPropagation somewhere so that clicking to select a value doesn't open the rule select menu

mrstegeman: Done.

PR #1502, opened 2018-11-21
Fix local configs.

mrstegeman:
Apparently, these stopped working after the symlink was changed
to a copy.

Fixes mozilla-iot/gateway-docker#13

### DIFF
--- a/src/user-profile.js
+++ b/src/user-profile.js
@@ -194,12 +194,17 @@ const Profile = {
     const oldLocalConfigPath = path.join(this.gatewayDir, 'config', 'local.js');
     const userConfigPath = path.join(this.configDir, 'local.json');
 
+    if (!fs.existsSync(userConfigPath)) {
+      if (fs.existsSync(oldUserConfigPath)) {
+        const oldConfig = config.util.parseFile(oldUserConfigPath);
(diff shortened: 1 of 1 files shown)


src/user-profile.js
```diff
@@ -189,15 +190,25 @@ const Profile = {
...
-    if (!fs.existsSync(userConfigPath)) {
-      fs.writeFileSync(
-        userConfigPath, '\'use strict\';\n\nmodule.exports = {\n};');
+    const oldUserConfigPath = path.join(this.configDir, 'local.js');
+    const oldLocalConfigPath = path.join(this.gatewayDir, 'config', 'local.js');
+    const userConfigPath = path.join(this.configDir, 'local.json');
+
+    if (fs.existsSync(oldUserConfigPath)) {
+      const oldConfig = config.util.parseFile(oldUserConfigPath);
+      fs.writeFileSync(userConfigPath, JSON.stringify(oldConfig, null, 2));
```
hobinjk: This can overwrite `userConfigPath` if the user happens to have one already. I think it would be better to only copy over the old config if userConfigPath doesn't exist and then fall back to `'{\n}'`.

mrstegeman: Nice catch. Fixed.

PR #1524, opened 2018-11-30
Add new /media route.

mrstegeman:
Add-ons can store media here, such as images and videos, and use
an href to this path. The route is authenticated, so nothing is
open to the world.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -87,9 +87,12 @@ const Router = {
     app.use(APP_PREFIX + Constants.OAUTH_PATH, nocache,
             require('./controllers/oauth_controller').default);
 
-    // Web app routes - send index.html and fall back to client side URL router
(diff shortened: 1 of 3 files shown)


mrstegeman:
@hobinjk @benfrancis This serves as an easy solution to our video/image proxying problem.

src/router.js
```diff
@@ -88,6 +88,8 @@ const Router = {
             require('./controllers/oauth_controller').default);
 
     // Web app routes - send index.html and fall back to client side URL router
+    app.use(APP_PREFIX + Constants.MEDIA_PATH, nocache, auth,
```
hobinjk: The comment above should be moved

mrstegeman: Done.

PR #1544, opened 2018-12-14
Add support for kilowatt hours.

mrstegeman:


freaktechnik:
The common SI symbol for them that I know is "kWh", Wikipedia also lists some alternatives, but it doesn't use - at all (that seems like a really weird thing to use in a short symbol for a unit).

mrstegeman:
@freaktechnik I actually pulled the symbol directly from the info box on the right side of the [Wikipedia page](https://en.wikipedia.org/wiki/Kilowatt_hour).

If you look at the derived units on [this page](https://physics.nist.gov/cuu/Units/units.html), the dot is used in several places.

freaktechnik:
oh, right, you used the times dot, missed that. Sorry, I read it as a dash/minus. I guess leaving the dot out would make it easier to use the unit, though.

mrstegeman:
This is just the abbreviation conversion function. In a property description, you can use `kilowatt-hour`.

PR #1548, opened 2018-12-19
Use number input in schema form where applicable.

### LINKED ISSUE #1547: Adapter settings input types
The configuration input for an adapter setting with the type "number" generates a "text" input. See for example the duration setting of the pulse adapter.

mrstegeman:
Fixes #1547

### DIFF
--- a/src/test/schema-form/form-test.js
+++ b/src/test/schema-form/form-test.js
@@ -506,12 +506,15 @@ describe('Form', () => {
           schema,
         });
 
-        const input = node.querySelector('input[type=text]');
+        const input = node.querySelector('input[type=number]');
         input.value = 'not a number';
         fireEvent(input, 'change');
 
         expect(node.querySelector('.error-item').textContent.trim()).toEqual(
-          '.field1 should be number',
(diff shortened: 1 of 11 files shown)


src/test/schema-form/form-test.js
```diff
@@ -506,12 +506,15 @@ describe('Form', () => {
...
 
-        const input = node.querySelector('input[type=text]');
+        const input = node.querySelector('input[type=number]');
         input.value = 'not a number';
         fireEvent(input, 'change');
 
         expect(node.querySelector('.error-item').textContent.trim()).toEqual(
-          '.field1 should be number',
+          // TODO: uncomment this when jsdom's validation is no longer broken.
+          // it doesn't set input.validity.badInput when it should.
```
hobinjk: Is there an upstream issue for this? (or a workaround for displaying the correct error message)

mrstegeman: No, and no. The `validity` object is read-only, so I couldn’t manually set the proper flags.

PR #1550, opened 2018-12-20
Change display of read-only properties.

### LINKED ISSUE #1471: readOnly strings and enums look the same as writable ones
Currently a readOnly enum looks exactly like a writable one. It has the dropdown arrow, and clicking on it does nothing.

Currently a readOnly string looks exactly like a writable one.

I think that a readOnly enum should look like a readOnly string, and that a readOnly string should look something like a voltage (i.e. background color same as the property bubble).

mrstegeman:
Fixes #1471

### DIFF
--- a/static/js/property-detail/enum.js
+++ b/static/js/property-detail/enum.js
@@ -33,13 +33,40 @@ class EnumDetail {
 
   view() {
     const unit = this.unit || '';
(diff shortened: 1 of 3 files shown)


dhylands:
I just realized that "something" is the units and 0 is the value being displayed, so this looks fine. Now I just have to figure out how to change my review to "Approved"

PR #1564, opened 2019-01-03
Improve rules error handling

### LINKED ISSUE #1560: Rule does not always properly activate/deactivate a device
I am using the Pulse add-on to create a timer for a switch and have a Rule like this
'While PumpTimer is on, turn Circulation Pump on'

When testing this rule with a virtual device it always works as expected.  However, when using it with a Eufy smart plug it occasionally fails in various ways.
1. Activating the timer does not activate the plug
1. When the timer deactivates the plug is not turned off

I also tried a pair of `If` rules instead but have had similar issues.  Latest release 0.6.1

screenshot_2019-01-02 things gateway

hobinjk:
Fix #1560

### DIFF
--- a/src/rules-engine/Property.js
+++ b/src/rules-engine/Property.js
@@ -62,15 +62,24 @@ class Property extends EventEmitter {
    * @return {Promise} resolves to property's value
    */
   async get() {
-    return await Things.getThingProperty(this.thing, this.id);
+    try {
+      return await Things.getThingProperty(this.thing, this.id);
+    } catch (e) {
+      console.warn('Rule get failed', e);
+    }
   }
 
   /**
    * @param {any} value
-   * @return {Promise} resolves if property is set to value
+   * @return {Promise} resolves when set is done
    */
-  async set(value) {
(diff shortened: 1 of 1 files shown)


src/rules-engine/Property.js
```diff
@@ -62,15 +62,24 @@ class Property extends EventEmitter {
...
   }
 
   /**
    * @param {any} value
-   * @return {Promise} resolves if property is set to value
+   * @return {Promise} resolves when set is done
    */
-  async set(value) {
-    await Things.setThingProperty(this.thing, this.id, value);
+  set(value) {
```
mrstegeman: Should this ever return a rejected promise which could be handled elsewhere? Just a thought.

hobinjk: At least for now I'm putting the error handling in this class. Otherwise both SetEffect and PulseEffect would need to implement their own error handling. Another place that would make sense is PropertyEffect but if a future rule wants to do things with properties (i.e. the Copy effect) I think this Property utility class is the correct place

PR #1566, opened 2019-01-06
add ewelink - sonoff compatibility 

mcampanha:
add ewelink - sonoff compatibility


mrstegeman:
@mcampanha I don't actually see any commits from you in here. Am I missing something?

PR #1578, opened 2019-01-11
Install ffmpeg in base image.

mrstegeman:
This is needed for video camera support. Also, clean up the prep
scripts after building image.


mrstegeman:
@hobinjk It increases the images size by something like 250 MB, which is really unfortunate.

PR #1588, opened 2019-01-17
add optargs to show usage of LTS for NVM

rhyspowell:
Not being familiar with NVM, didn't know it had to be called. As LTS is recommended, add it to the docs so copy and paste just works

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,31 +66,6 @@ nvm allows you to easily install different versions of node. To install nvm:
 $ curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
 ```
 
-Reinitialize your terminal session.
-
-```
-$ . ~/.bashrc
-```
-
-Use nvm to install and use an LTS version of node and then set the
-default version. The **`.nvmrc`** file will be used by nvm to determine which version of node to install.
-
-```
-$ nvm install
-$ nvm use
-$ nvm alias default $(node -v)
-```
-
-Verify that node and npm have been installed:
-```
-$ node --version
-v8.12.0
-$ npm --version
-6.4.1
-```
-
-Note: these versions might differ from the LTS version installed locally.
-
 ### Install node (if you didn't use nvm)
 
 (If you already installed node via nvm you can skip this step)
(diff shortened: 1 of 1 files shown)


mrstegeman:
Right above that line is the following:

```
The **`.nvmrc`** file will be used by nvm to determine which version of node to install.
```

Was that not actually being used?

rhyspowell:
I could see it was there, but nmv kept showing me the 'help' details. This is on ubuntu 18.04, not sure if its specific to that OS?

mrstegeman:
Can you `cd` into the gateway directory and run `nvm install` and `nvm --version` and post your output?

Here's what I get:
```
pi@gateway:~/mozilla-iot/gateway $ nvm install
Found '/home/pi/mozilla-iot/gateway/.nvmrc' with version <lts/carbon>
v8.15.0 is already installed.
Now using node v8.15.0 (npm v6.4.1)
```
```
pi@gateway:~/mozilla-iot/gateway $ nvm --version
0.33.8
```

rhyspowell:
hmmm

```
bleachin@hass:~$ nvm install
No .nvmrc file found
```

let me just build a fresh vm and run through it again

mrstegeman:
You ran it in your home directory. You need to run `nvm install` from within the gateway directory.

rhyspowell:
I have no gateway directory. There is no request to create one at that point and the git clone for the repo isn't until later.

I can see in the code that the dot file required exists so is it just an ordering of the commands that's the issue?

Also rolled back the second change, shouldn't have tried to do this through the git interface!

mrstegeman:
Ah, there is the issue. Yes, the README should be reordered. You can keep the `nvm` installation piece where it is, but move the set of 3 `nvm` commands down to after cloning the gateway.

rhyspowell:
ok, will get that sorted shortly

README.md
```diff
@@ -66,31 +66,6 @@ nvm allows you to easily install different versions of node. To install nvm:
 $ curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
 ```
 
-Reinitialize your terminal session.
```
mrstegeman: I would keep the `Reinitialize` part here.

README.md
```diff
@@ -275,6 +250,31 @@ python3 -m pip install git+https://github.com/mozilla-iot/gateway-addon-python.g
     $ cd gateway
     ```
 
+* Use nvm to install and use an LTS version of node and then set the default version. The **`.nvmrc`** file will be used by nvm to determine which version of node to install.
```
mrstegeman: Can you change this statement to be:

```
If you chose to install nvm above, use it to install and use an LTS version of node and then set the default version. The **`.nvmrc`** file will be used by nvm to determine which version of node to install.
```

README.md
```diff
@@ -275,6 +250,31 @@ python3 -m pip install git+https://github.com/mozilla-iot/gateway-addon-python.g
...
+    ```
+    $ node --version
+    v8.12.0
+    $ npm --version
+    6.4.1
+    ```
+
+    Note: these versions might differ from the LTS version installed locally.
+
+* Reinitialize your terminal session.
```
mrstegeman: Remove the `Reinitialize` part and keep it where it was.

README.md
rhyspowell: Done

README.md
rhyspowell: Done

PR #1591, opened 2019-01-17
Switch to Shaka Player for video.

mrstegeman:
The reference DASH.js player had issues with the DASH manifest
generated by ffmpeg on the RPi, but Shaka Player handles it just
fine. It also handles HLS video, so we can drop both old players.


static/js/property-detail/video.js
```diff
@@ -96,42 +96,18 @@ class VideoDetail {
...
-      player.extend('RequestModifier', () => {
-        return {
-          modifyRequestHeader: (xhr) => {
-            xhr.setRequestHeader('Authorization', `Bearer ${API.jwt}`);
-            return xhr;
-          },
-          modifyRequestURL: (url) => {
-            return url;
-          },
+    if (shaka.Player.isBrowserSupported() && (this.dashHref || this.hlsHref)) {
```
hobinjk: This wasn't part of the functionality before, but it would be good to have a warning if the browser isn't supported (if there are even that many unsupported browsers)

mrstegeman: Essentially they only browsers that are not supported are IE <= 10 and everything on iOS (which was the case before as well).

mrstegeman:
@hobinjk Added an error message.

hobinjk:
Great, thanks!

PR #1592, opened 2019-01-17
Allow proxying of web thing content.

### LINKED ISSUE #1587: Figure out how to reverse-proxy web thing resources
If a web thing includes links to images, UIs, etc., we need to reverse-proxy those such that they're available through the tunnel.

mrstegeman:
Requires mozilla-iot/gateway-addon-node#26
Fixes #1587

### DIFF
--- a/src/controllers/proxy_controller.js
+++ b/src/controllers/proxy_controller.js
@@ -1,0 +1,46 @@
+/**
+ * Proxy Controller.
+ *
+ * Handles proxied resources.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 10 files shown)


src/controllers/proxy_controller.js
```diff
@@ -0,0 +1,46 @@
...
+ProxyController.addProxyServer = (thingId, server) => {
+  proxies.set(thingId, server);
+};
+
+ProxyController.removeProxyServer = (thingId) => {
+  proxies.delete(thingId);
+};
+
+/**
+ * Log out the user
```
hobinjk: This comment is inaccurate

mrstegeman: Fixed.

Issue #1318, opened 2018-08-29
Default precision of number properties


benfrancis:
How precise should various number properties be in the UI by default?

The current approach appears to be zero decimal points for power, voltage and frequency but one decimal point for current for example.

Would a significant figures approach work better? E.g. 3 significant figures by default?

mrstegeman:
Alternatively, we could switch units as necessary. For example, instead of showing `0.0310 A`, we could show `31.0 mA`.

twobraids:
I support dynamically switching the units for scale so that the display field can generally show 3 digits (or a configurable number) plus a decimal point.

freaktechnik:
I'm not convinced the current default of three decimal digits is a great default, maybe a way for the device to declare accuracy would be a good idea. This would allow the gateway UI to scale to what it can expect from the property. My suggestion would be a property that can have any value between 1 and 0 (including 1, excluding 0). If it is unset, infinite precision is assumed.

Properties with the integer type should get no decimal digits, currently they get the same as a number type. So integers would have a precision of "1".

flatsiedatsie:
> maybe a way for the device to declare accuracy would be a good idea

That would be great to have. That being said, in [some cases](https://github.com/mozilla-iot/gateway/issues/1809#issuecomment-486695028) it's not possible for the device to already suggest how many decimals will be required.


I've spotted some off things around this issue.

First, I have a temperature sensor that has one decimal of precision, but the Gateway always truncates it to remove the decimal? And no, I haven't set it to 'integer'.

```
# Adding temperature property
                    self._type.append('MultiLevelSensor')
                    self.properties[targetPropertyID] = MySensorsProperty(
                        self,
                        targetPropertyID,
                        {
                            '@type': 'TemperatureProperty',
                            'label': new_description,
                            'type': 'number',
                            'unit': 'degree celsius',
                            'readOnly': True,
                        },
                    new_value, new_node_id, new_child_id, new_sub_type)
```

Results in:

![temp-no-decimal](https://user-images.githubusercontent.com/805405/56744476-22d2a280-6768-11e9-9628-407491df0676.png)



On the other side of the spectrum, properties that are initiated without an official type/property (I have a lot of those..) get trailing zeros if they are floats:

![co2-thing](https://user-images.githubusercontent.com/805405/56744822-bb692280-6768-11e9-9d1f-c1a5923ecb27.png)


The code that initiates it:

```
            elif child.type == 22:                      # Air quality
[...]

PR #1602, opened 2019-01-24
Fix precision of readOnly integer properties.

mrstegeman:
References #1318


freaktechnik:
With this PR: 
image

Without:
image

PR #1609, opened 2019-01-25
Store the commands sent to the commands API in the database and display them in the smart assistant page. 

### LINKED ISSUE #1542: Tie Commands API and Virtual Assistant

andrenatal:
Fixes #1542

### DIFF
--- a/src/controllers/commands_controller.js
+++ b/src/controllers/commands_controller.js
@@ -37,6 +37,9 @@ const AddonManager = require('../addon-manager');
 const CommandUtils = require('../command-utils');
 const IntentParser = require('../models/intentparser');
 const Things = require('../models/things');
+const WebSocket = require('ws');
+const db = require('../db');
+const JSONWebToken = require('../models/jsonwebtoken');
 
 const CommandsController = PromiseRouter();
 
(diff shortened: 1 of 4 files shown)


src/controllers/commands_controller.js
```diff
@@ -208,4 +220,33 @@ CommandsController.post('/', async (request, response) => {
   });
 });
 
+const wss = new WebSocket.Server({port: CommandUtils.wsSettings.port});
```
hobinjk: This should use `CommandsController.ws('/', function(ws) {});` so that it doesn't have to do its own JWT checking and connection management. I also think that listening on a port like this doesn't work with the tunneling system.

benfrancis:
I understand why you might want this if you only talk to the gateway through a microphone plugged directly into the gateway, but it has the potential to get confusing if the gateway is used in the way that is actually intended. Imagine you have multiple clients in different rooms (on smart displays, tablets or smartphones) and multiple people interact with the assistant at the same time. The conversation history of separate conversations would all get muddled together and a response to a command issued on one client would appear on another client.

I think the way that Alexa solves this problem is to have the concept of sessions so that a response intended for one session doesn't get sent to another session. I don't think you'd expect the conversation you have with Google Assistant or Alexa on your smartphone to appear on your smart display in the kitchen and vice versa.

If I've understood correctly then I don't think this feature should be landed as currently implemented, if at all. Conversations should probably stay on a per client basis and not get displayed across all clients.

mrstegeman:
I would argue that they should not get stored at all. This is one of my big gripes with current voice assistants, that they store everything, forever. There’s no good reason to do so. That applies to both the text and the audio clips.

hobinjk:
I like the current per-browser-window storage and think that an extension to per-user storage would be useful for some cases. The main way I see this helping with is if a user forgets the available voice commands they can consult what is effectively personalized documentation illustrating exactly the commands they found useful in the past.

kgiori:
I like experimenting with private, local voice commands, even if direct mic/speaker integration with the Things Gateway would not be desirable as an end product. It's useful to try things to prototype possibilities. And developers everywhere I've traveled to recently are excited by the chance to try local, private voice alternatives to cloud-based solutions. The difficulty I have with the current private voice-controller add-on is that when it doesn't interpret what I said correctly, I don't know why. I don't know what it thinks I said. We had that same problem with the original voice interface via the browser, before the Smart Assistant page existed. The smart assistant page has helped tremendously learning why voice commands through the browser have failed. Almost always it is because I spoke too soon -- it captured only the tail end of what I said. The same benefit for seeing local voice commands is what I would expect as a user of this voice-controller add-on. 

I think we should implement this approach, even if temporarily, to learn from it. To avoid confusion between the two options, could we prefix the wake word to the displayed command? For example, "Magic House: turn on the kitchen light" for local voice-controller, "turn on the kitchen light" for browser voice or msg. If we later allow external messaging commands, such as from WhatsApp, those could have pre-pended labels too: "WhatsApp_KG: turn on the kitchen". And for 3rd party smart speakers, something like "Alexa (or Google or Mycroft...): turn on the kitchen lights." 

For storing voice commands, I think it should be up to the user. Just like logging of data in general (e.g., storing timestamps of lights going on/off) can be up to the user. Whether or not to store commands can be a variable on the voice controller add-on configuration  page, or my preference would be to have it be a configuration exposed per web thing (in this case the voice-controller). On the "Edit" page of each web thing where users change the device name, and select the main icon "capability", we could add a check box for whether or not to "store device data".

benfrancis:
Hi @kgiori, there are a lot of separate issues mixed up in there regarding architectural design, UI design and our product roadmap. I'm not exactly sure where to start.

If we were going to store smart assistant commands on the gateway at all then it would at a minimum need to be stored on a per-user basis. The way this voice add-on is implemented (requiring the user to manually generate an access token, with no per-user authentication) is going to make that tricky.

I think we have a consensus among engineers that this pull request is not the right technical approach, and as the designer/product owner I'm not happy with the UI hacks you've proposed just to make it easier to debug.

Apart from that, this general approach is not consistent with our product roadmap and continuing to use it in demos risks misleading the community about what the gateway is intended for and what the final product is going to look like.

This voice add-on is really one big hack which abuses the add-ons system to install a number of software packages (with no way to remove them afterwards) on the gateway in order to turn it into a rudimentary smart speaker, then exposes a web thing in the gateway UI for the sole purpose of triggering the training process. I admire Andre's ingenuity in getting it to work this way, but I'm not really happy about it being in the add-ons directory at all.

We're not building a gateway + smart speaker product, but if we were this is not how we would do it.

I have an alternative approach to speech which I'm in the process of discussing with Andre so I'm going to close this pull request.

PR #1611, opened 2019-01-27
Update generate-release to not expect a .git directory

hobinjk:


tools/generate-release.sh
```diff
@@ -11,15 +11,21 @@ if [ ! -e "package.json" ]; then
...
+
+git init .
+git config user.email "temporary@example.com"
+git config user.name "Temporary"
+git add .
+git commit -m "Temporary"
 git clone ./ gateway
-cd gateway
-git remote set-url origin "${_remote}"
+cd gateway || exit
```
mrstegeman: You may want this and line 28 to be `exit 1`.

hobinjk: Good point! Error conditions should get error codes

PR #1638, opened 2019-02-06
Allow user to rearrange icons.

### LINKED ISSUE #1015: Rearrange icons
It would be very nice to be able to rearrange the things icons on the Things Page with a drag/drop.

mrstegeman:
Fixes #1015

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9633,6 +9633,11 @@
         "minimist": "0.0.8"
       }
     },
+    "mobile-drag-drop": {
+      "version": "2.3.0-rc.1",
+      "resolved": "https://registry.npmjs.org/mobile-drag-drop/-/mobile-drag-drop-2.3.0-rc.1.tgz",
+      "integrity": "sha512-fp2IjJWaZLknjNuzSKjuNDjYARhHVUKJvKOI8DnMLCP9Fe0LaSHbRGCGjNe00yX8gzdSIKpFuP7pUe0jcOD1pQ=="
+    },
     "moment": {
       "version": "2.24.0",
       "resolved": "https://registry.npmjs.org/moment/-/moment-2.24.0.tgz",
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "le-acme-core": "^2.1.4",
     "le-challenge-dns": "^2.3.2",
     "mkdirp": "^0.5.1",
+    "mobile-drag-drop": "^2.3.0-rc.1",
     "nanomsg": "^4.0.2",
     "ncp": "^2.0.0",
     "nocache": "^2.0.0",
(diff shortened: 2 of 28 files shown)


mrstegeman:
> * It's unclear where the dragged thing will end up

This should be clearer now.

> * It's possible to get multiple things stuck in the drag hover state (the dashed border thing) by dragging around too quickly

Fixed.

mrstegeman:
> * The thing icon doesn't appear to be included in the elements that are being dragged so it's less visible. This also makes it hard to realize that you are dragging the thing into a new position

This only seems to be an issue on desktop Firefox, annoyingly.

> * There's no distinct feedback that dragging has started (e.g. the iOS wiggle animation or the android vibration)

Things should be a little better now.

benfrancis:
This is very cool. The vibration is a nice touch.

This is a hard interaction to get right. Some things I noticed:
- Dragging the icon doesn't interact with scrolling in quite the way I would expect. On mobile, sometimes when I try to drag icons vertically it just scrolls the page and I can't move the icon beyond its current row. Also when I drag an icon to the edge of the screen, I would expect the page to then start scrolling so I can continue dragging up/down the page, which it doesn't.
- The dotted line looks a bit strange visually. I'm wondering if having the line only extend to the height of the icon rather than the height of the icon + label might help, if that isn't too hard to achieve?
- On mobile, trying to scroll the page with my finger over a thing icon can immediately trigger a drag. This makes it difficult to scroll the page.
- I am a bit worried about the inconsistency between how you move icons in the things page vs. how you move them in the floorplan page, but currently I can't think of a better design.

I see that you followed the same method of saving the position index as I did for saving the floorplan position (via the REST API). I didn't love that approach at the time because it means exposing that implementation-specific data through the Web Thing API using non-standard object properties, but I couldn't think of anything better. Does anyone have any other suggestions?

mrstegeman:
> This is very cool. The vibration is a nice touch.

I wish it worked on mobile safari.

> * Dragging the icon doesn't interact with scrolling in quite the way I would expect. On mobile, sometimes when I try to drag icons vertically it just scrolls the page and I can't move the icon beyond its current row. Also when I drag an icon to the edge of the screen, I would expect the page to then start scrolling so I can continue dragging up/down the page, which it doesn't.

Should be fixed.

> * The dotted line looks a bit strange visually. I'm wondering if having the line only extend to the height of the icon rather than the height of the icon + label might help, if that isn't too hard to achieve?

Fixed.

> * On mobile, trying to scroll the page with my finger over a thing icon can immediately trigger a drag. This makes it difficult to scroll the page.

Should be fixed. We can adjust the timeout if necessary, I have it set to 500 ms before dragging starts.

> * I am a bit worried about the inconsistency between how you move icons in the things page vs. how you move them in the floorplan page, but currently I can't think of a better design.

I figured it was going to be overly complex to do a full SVG layout like the floorplan, so I went this route.

> I see that you followed the same method of saving the position index as I did for saving the floorplan position (via the REST API). I didn't love that approach at the time because it means exposing that implementation-specific data through the Web Thing API using non-standard object properties, but I couldn't think of anything better. Does anyone have any other suggestions?

Yeah, I just hijacked the existing floorplan handler. Made sense to me.

mrstegeman:
@benfrancis @hobinjk I believe I've addressed everything. Any additional feedback?

PR #1643, opened 2019-02-11
Add initial support for pairing prompts and credentials.

### LINKED ISSUE #1541: Allow adapters to prompt user to take action when pairing
Some adapters, such as Z-Wave, require the user to push a button on the device during the discovery process, but there is currently no way to tell the user about that.

Other adapters can discover devices, but cannot pair unless the user pushes a button or something similar.

mrstegeman:
Fixes #1541

### DIFF
--- a/static/js/new-thing.js
+++ b/static/js/new-thing.js
@@ -24,6 +24,7 @@ class NewThing {
     this.id = id;
     this.description = description;
     this.pinSet = false;
+    this.credentialsSet = false;
     this.iconData = null;
     this.re = null;
     if (description.hasOwnProperty('pin') && description.pin.pattern !== null) {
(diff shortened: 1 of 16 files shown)


static/js/new-thing.js
```diff
@@ -232,6 +266,28 @@ class NewThing {
       this.pinInput.addEventListener('input', this.handlePinInput.bind(this));
       this.cancelButton.addEventListener('click', this.handleCancel.bind(this));
       this.submitButton.addEventListener('click', this.handleSubmit.bind(this));
+    } else if (this.requiresCredentials()) {
+      this.usernameInput =
+        this.element.getElementsByClassName('new-thing-username')[0];
```
hobinjk: Complete nitpick, but this would be a bit cleaner as `this.element.querySelector('.new-thing-username')`

mrstegeman: Agreed. Done.

PR #1646, opened 2019-02-12
Pin gateway-addon library versions.

mrstegeman:


mrstegeman:
This is for the v0.7.0 branch, in case we decide to do a point release. I pinned the libraries to where they were for v0.7.0.

PR #1653, opened 2019-02-13
Update chai-http.

mrstegeman:


mrstegeman:
@hobinjk chai-http now shuts down the server after every request, so some workarounds had to be added in the tests to handle that.

src/test/integration/actions-test.js
```diff
@@ -77,36 +75,36 @@ describe('actions/', function() {
   });
 
   it('should fail to create a new action (empty body)', async () => {
-    const err = await pFinal(chai.request(server)
+    const err = await chai.request(server)
```
hobinjk: I'm so glad this behaves properly now

PR #1662, opened 2019-02-15
Add tests for events and proxied resources.

mrstegeman:


mrstegeman:
The Travis failure is just the normal one we see all the time now, which is super annoying.

PR #1667, opened 2019-02-19
Swap certificates in running server after renewal.

mrstegeman:


mrstegeman:
I know. There's not an official way to do this outside of taking down the server completely and restarting it. I found this solution on the node issue tracker, and everyone seemed satisfied with it.

For reference: https://github.com/nodejs/node/issues/4464

PR #1668, opened 2019-02-19
Add initial OpenWrt network configuration bits.

mrstegeman:


src/network-manager.js
```diff
@@ -0,0 +1,281 @@
...
+          return false;
+        }
+      }
+
+      proc = child_process.spawnSync('uci', ['commit', 'network']);
+      if (proc.status !== 0) {
+        return false;
+      }
+
+      proc = child_process.spawnSync('/etc/init.d/network', ['reload']);
```
dhylands: This should be `restart` not `reload`

src/network-manager.js
```diff
@@ -0,0 +1,281 @@
...
+
+      let device = null;
+      let iface = null;
+      for (const line of proc.stdout.split('\n')) {
+        if (line.indexOf('=') < 0) {
+          continue;
+        }
+
+        const [key, value] = line.split('=', 2).map((s) => stripQuotes(s));
+
```
dhylands: It feels like everything from spawnSync to here could be encapsulated into a common function?
Perhaps something that calls uci show and splits the results into lines?

src/network-manager.js
mrstegeman: The OpenWrt wiki disagrees: https://oldwiki.archive.openwrt.org/doc/uci/network

src/network-manager.js
mrstegeman: Agreed, this could be cleaned up a bit.

src/network-manager.js
dhylands: ok - I take it back. Running `/etc/init.d/network help` shows:
```
Syntax: /etc/init.d/network [command]

Available commands:
        start   Start the service
        stop    Stop the service
        restart Restart the service
        reload  Reload configuration files (or restart if service does not implement reload)
        enable  Enable service autostart
        disable Disable service autostart
```
and if I change the SSID then using reload does in fact change the SSID being advertised.

I guess I've just been used to using restart from the regular system V init scripts.

PR #1671, opened 2019-02-20
Allow configuration of LAN.

mrstegeman:


mrstegeman:
@dhylands Nice catch. I've fixed parsing to handle lists.

PR #1672, opened 2019-02-21
Refactor platform-specific functionality.

mrstegeman:
Things are much more extensible this way.


mrstegeman:
> I think it might be simpler to use `if (Platform.setMdnsStatus) {` instead of having a separate function for `if (Platform.implemented('setMdnsStatus')) {` but I have no objection to the approach you've chosen.

I considered that, but I wanted everything to be defined in case we screwed up, that was we can at least get a descriptive error via the `NotImplementedError`.

PR #1673, opened 2019-02-21
Allow DHCP enable/disable.

mrstegeman:


src/platforms/linux-openwrt.js
```diff
@@ -132,6 +132,39 @@ function uciCommit(key) {
   return proc.status === 0;
 }
 
+/**
+ * Get DHCP server status.
+ *
+ * @returns {boolean} Boolean indicating whether or not DHCP is enabled.
+ */
+function getDhcpStatus() {
```
dhylands: It's probably worth expanding the function name a bit. DHCP will run in 2 contexts: DHCP server for LAN side, and DHCP client for WAN side (or possiblly LAN side if doing an extender setup).

So I think that these should be called getDhcpServerStatus and setDhcpServerStatus

Something for future consideration: It's probably also worthwhile checking to see if we can stop the DHCP server process if it isn't being used. For my turris setup, I did:
```
/etc/init.d/odhcp disable
/etc/init.d/odhcp stop
```
to cause the DHCP server to not even run (which saves memory).

src/platforms/linux-openwrt.js
```diff
@@ -132,6 +132,39 @@ function uciCommit(key) {
...
+
+/**
+ * Set DHCP server status.
+ *
+ * @param {boolean} enabled - Whether or not to enable the DHCP server
+ * @returns {boolean} Boolean indicating success of the command.
+ */
+function setDhcpStatus(enabled) {
+  if (!uciSet('dhcp.lan.ignore', enabled ? '0' : '1')) {
+    return false;
```
dhylands: Do we want to log something in these cases? Otherwise if it fails, we'll have no idea why it failed.

src/platforms/linux-openwrt.js
mrstegeman: I can rename, that's a good point.

According to the OpenWrt docs, odhcp is only used for IPv6, while dnsmasq is used for IPv4.

src/platforms/linux-openwrt.js
mrstegeman: I figured we'd log wherever these functions get called.

src/platforms/linux-openwrt.js
dhylands: So we probably want to stop/start/configure both then. I was following this guide: https://openwrt.org/docs/guide-user/network/wifi/dumbap and that does the dhcp.lan.ignore=1 to disable the IPv4 and also disabled odhcpd

mrstegeman: Done. Can you take a quick look?

src/platforms/linux-openwrt.js
```diff
@@ -152,16 +152,39 @@ function getDhcpStatus() {
...
+  if (!uciSet('dhcp.lan.ra', enabled ? 'server' : 'disabled')) {
+    return false;
+  }
+
   if (!uciCommit('dhcp')) {
     return false;
   }
 
-  const proc = child_process.spawnSync('/etc/init.d/dnsmasq', ['reload']);
+  let proc = child_process.spawnSync('/etc/init.d/dnsmasq', ['stop']);
```
dhylands: Shouldn't this be start if enabled and stop if not?

mrstegeman: ... yes.

mrstegeman: Ok, fixed.

dhylands: LGTM

PR #1677, opened 2019-02-22
Networking improvements.

mrstegeman:
* Fix setHostname on Raspbian and OpenWrt.
* Implement getDhcpServerStatus on Raspbian.
* Implement setDhcpServerStatus on Raspbian.
* Implement getLanMode on Raspbian.
* Implement setLanMode on Raspbian.

### DIFF
--- a/src/platforms/linux-raspbian.js
+++ b/src/platforms/linux-raspbian.js
@@ -12,6 +12,41 @@ const child_process = require('child_process');
 const fs = require('fs');
 
 /**
+ * Get DHCP server status.
+ *
+ * @returns {boolean} Boolean indicating whether or not DHCP is enabled.
+ */
+function getDhcpServerStatus() {
+  const proc = child_process.spawnSync(
+    'sudo',
+    ['systemctl', 'is-active', 'dnsmasq.service']
+  );
+  return proc.status === 0;
+}
+
+/**
+ * Set DHCP server status.
+ *
+ * @param {boolean} enabled - Whether or not to enable the DHCP server
+ * @returns {boolean} Boolean indicating success of the command.
+ */
+function setDhcpServerStatus(enabled) {
+  let proc = child_process.spawnSync(
+    'sudo',
+    ['systemctl', enabled ? 'start' : 'stop', 'dnsmasq.service']
+  );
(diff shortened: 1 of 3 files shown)


src/platforms/linux-raspbian.js
```diff
@@ -137,7 +137,18 @@ function setHostname(hostname) {
...
+    [
+      'sed',
+      '-i',
+      '-E',
+      '-e',
+      `s/(127\\.0\\.0\\.1[ \\t]+)${original}/\\1${hostname}/g`,
+      '/etc/hosts',
+    ]
+  );
+  return proc.status === 0;
```
dhylands: /etc/hostname contains the hostname on the RPi. /etc/hosts should also be updated.
Rather than updating /etc/hostname, it's probably better to call `sudo hostname new-name` (which will update /etc/hostname, but not /etc/hosts).

mrstegeman: This was just an addition. `/etc/hostname` is already updated inside this function, and `hostname` is also called. However, calling `hostname` doesn't actually persist the new hostname to `/etc/hostname`.

dhylands: And In the /etc/hosts I've seen on the RPi, the address was 127.0.1.1 rather than 127.0.0.1

Here's the raspi-config routine that changes the hostname:
https://github.com/RPi-Distro/raspi-config/blob/1960514ea73d3669c370f9cb4d63c1f532c60784/raspi-config#L480

mrstegeman: Ah, that's what I based my code off, but I didn't see they used 127.0.1.1.

mrstegeman:
@dhylands I fixed the 127.0.1.1 thing and also added `getDhcpServerStatus` and `setDhcpServerStatus` for Raspbian.

src/platforms/linux-raspbian.js
dhylands: You're right hostname doesn't persist anything. But it does change something internally since if you do `sudo hostname foo` sudo then complains until /etc/hosts gets updated. I guess that's why raspi-config wants a reboot.

PR #1680, opened 2019-02-25
Update default dnsmasq configuration.

mrstegeman:


mrstegeman:
@dhylands Should we also update the config to serve DHCP on eth0?

dhylands:
Which router are we talking about? This looks like the captive portal IP address. For captive portal I think we should only run the DHCP server on wifi. It probably makes sense to run DHCP client on any LAN ports.

I know on my RPi, I often connect to LAN and setup the WiFi SSID/password over the LAN connection because my desktop doesn't have WiFi

mrstegeman:
I think we'll only ever want to run the DHCP server on the wifi network, given that there's just a single ethernet port. It doesn't make a lot of sense to run it on eth0 as well.

dhylands:
I think once things are setup, if you're running like a "normal" router, i.e. internet access comes in through the WAN port, then you definitely want to run the DHCP server for both WiFi and LAN ports.

If the internet comes through WiFi or LAN, then we wouldn't want to run a DHCP server.

PR #1681, opened 2019-02-25
Implement get/set wireless mode on Raspbian.

mrstegeman:


src/platforms/linux-raspbian.js
```diff
@@ -131,6 +131,293 @@ function setLanMode(mode, options = {}) {
...
+    'sudo',
+    ['systemctl', 'stop', 'hostapd.service']
+  );
+  if (proc.status !== 0) {
+    return false;
+  }
+
+  proc = child_process.spawnSync(
+    'sudo',
+    ['systemctl', 'disable', 'hostapd.service']
```
dhylands: Just a sanity check. Does this make sense here? If the system crashed/rebooted between here and when we rewrite the config file, then hostapd won't start on reboot. Just wanted to make sure that's what we actually want.

mrstegeman: I can move that into the `sta` check.

src/platforms/linux-raspbian.js
```diff
@@ -131,6 +131,293 @@ function setLanMode(mode, options = {}) {
...
+      config += `wpa_passphrase=${options.key}\n`;
+      config += 'wpa_pairwise=CCMP\n';
+    }
+
+    fs.writeFileSync('/tmp/hostapd.conf', config);
+
+    proc = child_process.spawnSync(
+      'sudo',
+      ['mv', '/tmp/hostapd.conf', '/etc/hostapd/hostapd.conf']
+    );
```
dhylands: /tmp and /etc are almost certainly on different filesystems. Which means that the mv will actually be a copy, which isn't atomic. I think it would be better to make the temporary file be /etc/hostspd.conf.new and then do the mv

Then the move will be atomic and you'll get either the entire new file or the entire old file, and not half of the new file if a crash occurs at the wrong time.

dhylands: And then you could use fs.renameSync rather than having to launch a process.

mrstegeman: The problem is that the gateway runs as non-root on Raspbian, but writing to `/etc` requires root access.

mrstegeman: Therefore, we can't directly use the `fs` methods to write to `/etc`.

dhylands: ok - then perhaps do a sudo cp to copy it into /etc and then do a sudo mv to rename it.

dhylands: These could be combined using something like sudo bash -c 'cp blah && mv blah'

mrstegeman: In our Raspbian image, `/etc` and `/tmp` are both on the same device, `/dev/mmcblk0p2`.

dhylands: ok - I'm convinced. I hadn't checked specifically, and I know many small distros use a ram disk for /tmp

PR #1684, opened 2019-02-26
Bump dependencies.

mrstegeman:


package.json
```diff
@@ -111,16 +111,16 @@
...
-    "shaka-player": "^2.5.0-beta2",
+    "selenium-standalone": "^6.15.6",
+    "shaka-player": "^2.5.0-beta3",
     "simple-oauth2": "^2.2.1",
     "sinon": "^7.2.4",
     "speaktome-api": "^0.2.1",
     "style-loader": "^0.23.1",
     "ts-jest": "^24.0.0",
-    "typescript": "^3.3.3",
+    "typescript": "^3.3.3333",
```
hobinjk: I was wondering why typescript jumped 3330 patch versions in one update and it turns out that this version number is not a cause for alarm and instead was just added as a joke by typescript's maintainers

PR #1685, opened 2019-02-27
Generate hotspot name based on MAC address.

mrstegeman:


src/views/hotspot.handlebars
```diff
@@ -8,8 +8,8 @@
      <VersionHigh>2.0</VersionHigh>
      <VersionLow>1.0</VersionLow>
      <AccessProcedure>1.0</AccessProcedure>
-     <AccessLocation>Mozilla IoT Gateway</AccessLocation>
-     <LocationName>Mozilla IoT Gateway</LocationName>
+     <AccessLocation>{{ap_ssid}}</AccessLocation>
+     <LocationName>{{ap_ssid}}</LocationName>
      <LoginURL>http://{{ap_ip}}/wifi-setup</LoginURL>
```
dhylands: Ugh - I don't think it should just be the Mac address. I think it should be something like Mozilla IoT Gateway followed by the last 6 digits of the Mac address.

mrstegeman: It is. It will end up being `Mozilla IoT Gateway 9BC2`, for example.

dhylands: Nevermind - i totally misread this

src/wifi-setup/wifi.js
```diff
@@ -15,6 +15,24 @@ function getStatus() {
...
+ * @returns {string} SSID
+ */
+function getHotspotSsid() {
+  const base = config.get('wifi.ap.ssid_base');
+  const mac = platform.getMacAddress('wlan0');
+  if (!mac) {
+    return base;
+  }
+
+  // Get the last 2 octets of the MAC and create a simple string, e.g. 9E28
```
dhylands: I'd recommend using the last 3 octets, since the first 3 octets are a manufacturer code and the last 3 digits will be unique within a given manufacturer code.

mrstegeman: I did this based on other wifi networks around me. It seems like most OEMs use a string of 4 digits at the end. Probably easier to remember than 6.

src/views/hotspot.handlebars
benfrancis: @mrstegeman  Sorry for the late reply.

I agree with last four digits. Sorry to be a pain but can we make this e.g. "Things Gateway 9BC2" as in the mockups rather than "Mozilla IoT Gateway 9BC2". It was never meant to be "Mozilla IoT Gateway" in the first place because Mozilla IoT is the name of the team, not the software. I realise Mozilla IoT has the benefit that it has Mozilla's brand name in it, but I anticipate changing default again when we do our re-brand.

benfrancis: Oh, you may already have done this in #1686

mrstegeman: Yep, done in the other PR.

PR #1686, opened 2019-02-27
Implement router first-time setup.

### LINKED ISSUE #1622: Router First Time Setup
As a user I want to configure my gateway as a router when I first set it up, so I can use it as the main router for my home rather than connecting to another wireless network.

mrstegeman:
Fixes #1622


mrstegeman:
I've tested the OpenWrt platform functionality bit by bit as I wrote it. I tested the whole flow of the router setup on my Mac by adding `case 'darwin':` here: https://github.com/mozilla-iot/gateway/blob/b5bee4852fb2a09a63299838b27d53c16acc14da/src/app.js#L309

Then, you'll have to fake out a couple calls in `router-setup.js` since the platform functionality isn't implemented for darwin.

You can do the same for wifi-setup, since I made some changes there as well.

PR #1688, opened 2019-02-28
Update dependencies, fix security issues.

mrstegeman:
For future reference:

`npm audit fix` does a great job. Unfortunately, `yarn audit` doesn't
have a `fix` option, so the best workflow is probably:

1. Update packages as necessary.
2. `npm install`
3. `npm audit fix`
4. `rm yarn.lock`
5. `yarn import`

Obviously, skip 1 and 2 if not required.


mrstegeman:
@hobinjk @benfrancis @dhylands On a related note, should we just stop using yarn at this point? It's kind of silly to maintain lock files for both package managers, and I don't believe yarn really gains us much anymore.

benfrancis:
> On a related note, should we just stop using yarn at this point? It's kind of silly to maintain lock files for both package managers, and I don't believe yarn really gains us much anymore.

I have no objections.

rzr:
Maybe we can track a "yarn to npm" task, and then align this in adapters and so on..

mrstegeman:
@rzr It should be irrelevant for adapters. They can do whatever they want, since they're all built independently of the gateway.

dhylands:
> @hobinjk @benfrancis @dhylands On a related note, should we just stop using yarn at this point? It's kind of silly to maintain lock files for both package managers, and I don't believe yarn really gains us much anymore.

We'll need to fix and test the build scripts. We currently do a yarn/npm thing because yarn doesn't do cross compiling.

mrstegeman:
Yeah, that shouldn't be too hard to fix up.

hobinjk:
I agree that we should switch from yarn to npm since npm's lockfile includes everything we wanted from yarn.

PR #1689, opened 2019-02-28
build: Use v8.15.0 until v8.15.1 released for ARMv6

rzr:
Change-Id: I634089f3243e1ccf98a6afe62760b34c86aa79e3
Relate-to: https://social.samsunginter.net/@rzr/101670417948263179

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 [ "$NVM_VERSION" != "" ] || NVM_VERSION="v0.33.8"
-[ "$NODE_VERSION" != "" ] || NODE_VERSION="--lts=carbon"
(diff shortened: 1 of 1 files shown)


mrstegeman:
The `--lts=carbon` flag will automatically get the latest v8 version.

rzr:
If it is released, this patch is a workaround for ARMv6 it may be reversed once file lands there:

Actually the problem could be fixed on nvm which is looking up non existing file:
https://nodejs.org/dist/v8.15.1/#node-v8.15.1-linux-armv6.tar.xz

I shared this patch in case someone is facing this problem

rzr:
No more needed, ARMv6 is released

PR #1708, opened 2019-03-15
Stop using cache-loader for static files.

mrstegeman:
It seems to be broken, causing images and such to stop loading.


hobinjk:
cache-loader can be removed entirely since its main purpose was reducing the webpack build times on the pi in production

mrstegeman:
Even better. Done.

PR #1714, opened 2019-03-18
Remove unnecessary duplicate handlebars dependency.

mrstegeman:


src/router-setup.js
```diff
@@ -2,29 +2,23 @@ const bodyParser = require('body-parser');
...
-  const filename = path.join(Constants.VIEWS_PATH, name);
-  return Handlebars.compile(fs.readFileSync(filename, 'utf8'));
-}
-
-const routerSetupTemplate = getTemplate('router-setup.handlebars');
-const creatingTemplate = getTemplate('creating.handlebars');
-const hotspotTemplate = getTemplate('hotspot.handlebars');
+const hbs = expressHandlebars.create({
+  helpers: {
+    escapeQuotes: (str) => `${str}`.replace(/'/, '\\\''),
```
hobinjk: This looks like it loses whatever additional escaping HandleBars.SafeString was doing (which might be handled elsewhere)

mrstegeman: It actually just converts the argument to a string.

https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/safe-string.js

hobinjk: Wonderful, a truly Safe string incapable of doing any harm

PR #1720, opened 2019-03-26
Add a basic OpenWrt package updater.

mrstegeman:


mrstegeman:
The weird condition of updating node and having an add-on restart is definitely possible. Otherwise, though, the gateway + node should already be loaded into memory, so this shouldn't really be a problem.

As far as restarting the gateway, that's due to personal lack of knowledge about OpenWrt packages. With Debian packages, for instance, you can add post-installation triggers to do things like restarting services. I'm not sure if that's possible with OpenWrt or not. If not, we'll definitely want to restart the gateway ourselves.

mrstegeman:
@dhylands How much do you know about OpenWrt packaging? Can we add a post-install script/hook?

hobinjk:
An edge case I thought of the other day that might be worth worrying about is that we'll need the gateway package to schedule a restart instead of restarting immediately or the `opkg upgrade` will get killed as a child of the gateway

mrstegeman:
Hmmm... That's a good point. Maybe we should just leave out the post-install trigger and instead restart that gateway manually if it (or node) is updated?

hobinjk:
Yeah, I think checking if the packages upgraded includes dependencies that require a restart then restarting should cover most of the potential failure conditions and we can figure out the rest when testing on-device.

mrstegeman:
@hobinjk https://github.com/mozilla-iot/gateway/pull/1720/commits/299242da050969a44368200a9045a17cfa7d7b4a

PR #1721, opened 2019-03-26
Add 'use strict' everywhere.

mrstegeman:


mrstegeman:
`eslint` freaks out when I do that.

```'use strict' is unnecessary inside of modules``` x240

Issue #1624, opened 2019-02-05
Mycroft Skill


benfrancis:
As a user I want to control my smart home using the [Mycroft](https://mycroft.ai) voice assistant.

flatsiedatsie:
Would this be about installing MyCroft on the Pi? If so, there is already a local voice control option?
https://github.com/andrenatal/voice-addon

As far as I can tell this add-on is built by someone at Mozilla. Perhaps its wise to just focus on making that work well for version 1.0?

jaller94:
In my mind, this issue is best addressed by creating a new repository for a Mycroft Skill which is generic and can handle any Web Thing. The gateway can then expose connected things (as in the name "gateway").
A local authorization in the gateway should allow Mycroft to connect to the things.

hobinjk:
Sorry for not responding to this earlier. I typed up a reply but it's lost somewhere in my forest of tabs.

A skill for generic webthing control would be very compelling and interesting but would also require reimplementing the discovery and WoT API parts of the gateway in python. At least for now the goal for this skill is just to demonstrate that the gateway can be easily controlled by a third party service.

mrstegeman:
@hobinjk @benfrancis @kgiori Do we have any status on this? It's been a while.

hobinjk:
90% sure the only remaining item for the first review is for me to update the Mycroft skill's recommended commands list then we only need two more reviews

hobinjk:
https://github.com/MycroftAI/mycroft-skills/pull/1086 is where the work for this is taking place

mrstegeman:
Thanks for the update!

mrstegeman:
Actually, it looks like that PR was just merged. Does that mean we officially have a Mycroft skill now??

benfrancis:
Good news: The skill is now [live](https://market.mycroft.ai/skills/5081cacb-b54b-43b7-994f-83c6fc376ead) and I successfully managed to authenticate it with my gateway using the configuration UI

Bad news: Currently all voice commands I try cause the spoken response "An error occurred when processing request in Mozilla IoT gateway"

benfrancis:
@hobinjk Do you have any tips on debugging this?

hobinjk:
I think a more verbose error should show up in mycroft-cli-client

benfrancis:
Here is the string of exceptions I see in mycroft-cli-client:

```
[...]

PR #1733, opened 2019-04-03
Add Mycroft skill and skill testing client ids

hobinjk:
Note that http://localhost:8888 should be removed in the future and is
only there for testing the initial development of the skill.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -74,4 +74,5 @@ module.exports = {
       ssid_base: 'WebThings Gateway',
     },
   },
+  oauthTestClients: false,
 };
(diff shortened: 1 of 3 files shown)


hobinjk:
The use case for this is letting a Mycroft engineer test the oauth flow locally using w3c-interop. How do you feel about giving it a non-Mycroft name like "Localhost Testing Probably a Virus" (or just "Test OAuth Client")

Come to think of it, all of the addition of the 127.0.01:31338 clients for the automated tests should be guarded in a test flag

mrstegeman:
I like the virus name haha.

How about we just add a flag to the config file to enable all the test clients. I can toggle that flag on the w3c-interop instance.

hobinjk:
:+1: Sounds like a good plan

mrstegeman:
References #1624

PR #1735, opened 2019-04-04
Split out config files from image scripts.

### LINKED ISSUE #1510: Split systemd service files out of image builder script
This would allow for easier packaging and such outside of the Raspberry Pi image.

mrstegeman:
Fixes #1510

### DIFF
--- a/image/etc/init.d/resize2fs_once
+++ b/image/etc/init.d/resize2fs_once
@@ -1,0 +1,25 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          resize2fs_once
+# Required-Start:
+# Required-Stop:
+# Default-Start: 3
+# Default-Stop:
+# Short-Description: Resize the root filesystem to fill partition
+# Description:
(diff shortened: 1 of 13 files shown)


image/etc/init.d/resize2fs_once
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
dhylands: I think that this has to go someplace else.

The way it was coded before, it would get executed on the boot after prepare-base.sh was run. With this restructuring it will get executed on the boot before prepare-base.sh is run

mrstegeman: So, I thought about that, but I think this is ok. It's copied into the filesystem in the `prepare-base.sh` script, but it's not enabled. That's still done by the `add-gateway.sh` script, as before.

dhylands: Actually, I think it should be ok. The symlink is created in add-gateway.sh which is run after prepare-base.sh

PR #1742, opened 2019-04-04
Fixed copy/paste bug in etc/init.d/gateway-iptables

dhylands:


mrstegeman:
@dhylands this is still not right. Hang on.

mrstegeman:
Nevermind, good to go.

PR #1749, opened 2019-04-08
Fix router setup.

mrstegeman:
The startup logic was flawed in that the code would wait for an AP
to start up for 60 seconds, and then start the default AP after
that. However, other parts of the code did not distinguish between
a user-configured AP and the default AP.


mrstegeman:
Right. The network configuration gets stored in uci config files, so we should be ok there.

PR #1755, opened 2019-04-08
Update graph during drag; Misc. log improvements

### LINKED ISSUE #1746: Logs: Scroll in real time
First implementation wasn't fancy enough (~5fps on a Pixel) so it's going to take something fancier

Likely target at this point is to render the entire graph (or sections of the graph) and represent it as a single layer that's CSS transformed upon scrolling

hobinjk:
I made it fast enough to do in almost real time on a pixel. Kind of sad I didn't get to do the statically rendered infinite scroll tiles approach but I'm glad the one weird trick of converting everything to integers got a 4x speed up on its own.

Close #1746

### DIFF
--- a/static/js/logs/log.js
+++ b/static/js/logs/log.js
@@ -200,9 +200,13 @@ class Log {
 
   makePath(points, close) {
     const path = document.createElementNS('http://www.w3.org/2000/svg', 'path');
-    const text = ['M', points[0].x, points[0].y];
+    const text = [
+      'M', Math.round(points[0].x).toString(),
+      Math.round(points[0].y).toString(),
+    ];
     for (let i = 1; i < points.length; i++) {
-      text.push('L', points[i].x, points[i].y);
+      text.push('L', Math.round(points[i].x).toString(),
+                Math.round(points[i].y).toString());
     }
     if (close) {
       text.push('Z');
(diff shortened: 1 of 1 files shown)


static/js/logs/log.js
```diff
@@ -304,8 +301,12 @@ class Log {
     }
 
     if (this.rawPoints.length === 0) {
-      return {min: 0, max: 1};
+      return {
+        min: this.property.minimum || 0,
+        max: this.property.maximum || 1,
```
mrstegeman: There's probably a case where maximum could be 0 that would cause this to be incorrect.

static/js/logs/log.js
```diff
@@ -317,6 +318,23 @@ class Log {
...
     }
+
+    if (this.property.hasOwnProperty('minimum') &&
+        this.property.hasOwnProperty('maximum')) {
+      const propMin = this.property.minimum;
+      const propMax = this.property.maximum;
+      // If the description's min and max aren't ridiculously out of proportion
+      // use them since they likely have good properties
+      if ((propMax - propMin) / (max - min + 0.001) < 3 &&
+         min >= propMin &&
```
mrstegeman: nit: small alignment issue.

static/js/logs/log.js
hobinjk: :+1:, updated accordingly

PR #1765, opened 2019-04-09
Sort installed add-ons by display name.

mrstegeman:
This is already done on the discovered add-ons page.

### DIFF
--- a/static/js/views/settings.js
+++ b/static/js/views/settings.js
@@ -1387,9 +1387,9 @@ const SettingsScreen = {
         noAdapters.classList.add('hidden');
         adaptersList.classList.remove('hidden');
 
-        for (const metadata of adapters) {
-          new Adapter(metadata);
-        }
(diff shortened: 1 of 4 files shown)


static/js/views/settings.js
```diff
@@ -1520,13 +1520,14 @@ const SettingsScreen = {
...
-        components.set(
-          name,
-          new InstalledAddon(addon, this.installedAddons, this.availableAddons)
-        );
-      }
+      Array.from(this.installedAddons.entries())
+        .sort((a, b) => a[1].display_name.localeCompare(b[1].display_name))
+        .forEach((x) => {
+          components.set(
+            x[0],
```
hobinjk: nit: x could be named addon to make this a bit clearer

mrstegeman: `x` is actually a 2 element array, which are `name` and `addon`.

PR #1783, opened 2019-04-10
Make main menu scrollable.

### LINKED ISSUE #1781: Make main menu scrollable
As we are adding more entries to the main menu, if they're all enabled they may not all be accessible on a small screen (e.g. landscape on a smartphone). The main menu needs to be scrollable to reach the later entries.

mrstegeman:
Fixes #1781

### DIFF
--- a/static/css/menu.css
+++ b/static/css/menu.css
@@ -20,8 +20,11 @@
 #main-menu ul {
   list-style-type: none;
   margin: 0;
+  margin-top: 9.6rem;
   padding: 0;
-  padding-top: 9.6rem;
+  height: calc(100% - 9.6rem);
+  max-height: calc(100% - 9.6rem);
+  overflow: auto;
 }
 
 #main-menu a, .log-out-button {


hobinjk:
It would be nice to have a conditional border or drop-shadow on the bottom of the wordmark section so that the scrolling area is better distinguished. Otherwise it's just

benfrancis:
> It would be nice to have a conditional border or drop-shadow on the bottom of the wordmark section so that the scrolling area is better distinguished.

Oh sorry, I didn't spot this. Good point. Not actually sure how to fix this without re-designing the menu. Maybe a follow-up task? (At least it's now useable).

hobinjk:
Yeah, it's definitely good enough as-is and the bar does go a ways to setting off the scrollable vs non-scrollable content

mrstegeman:
@benfrancis @hobinjk Let me know what you think about the border I just added. I can back it out if you dislike.

hobinjk:
:+1: LGTM

benfrancis:
Thanks for trying that but honestly I think it looks better without.

@hobinjk's idea of a dropshadow might work better but would need applying to the top element.

Honestly, it's probably fine as it was for now as this is an edge case.

Issue #1793, opened 2019-04-15
"Undefined" message after 0.8 -> 0.7 -> 0.8 upgrade path


benfrancis:
STR:
* Upgrade from 0.7 to 0.8
* Load gateway web interface

Expected:
* Everything looks normal

Actual:
* "Undefined" message appears as a toast and can't be dismissed
* Source map errors in the console

This can be fixed with a hard reload, but it would be great to get the bottom of the cause because every user will probably see this and not everyone will be able to figure out how to fix it.

hobinjk:
I haven't been able to reproduce this yet but I believe that the most likely cause is https://github.com/mozilla-iot/gateway/blob/v0.7.0/static/js/app.js#L128 which means that I just need to add a no-op websocket listener to the Logs controller

hobinjk:
or rather get it to be a no-op if it's 0.7 listening instead of 0.8

hobinjk:
Thinking this through, this behavior will only happen if someone enables logging, has log data, then downgrades to 0.7 so it won't actually happen to normal users. Any objection to closing this?

hobinjk:
After #1794 this will only happen if someone does 0.8 -> 0.7 -> 0.8 and will not happen during the normal 0.7 -> 0.8 so I'm closing this for now

PR #1794, opened 2019-04-16
Guard against sending empty arrays when logs don't exist

hobinjk:
Related to #1793, most likely fixes it for normal use cases. If someone is using the 0.7 UI to look at the 0.8 backend this prevents 3 "undefined" messages from showing up. The "undefined" messages shouldn't be stuck on screen though so it's likely that the second part of #1793 is the proposed 0.8 -> 0.7 -> 0.8 transition

### DIFF
--- a/src/controllers/logs_controller.js
+++ b/src/controllers/logs_controller.js
@@ -165,6 +165,9 @@ LogsController.ws('/', (websocket) => {
   }, 30 * 1000);
 
   function streamMetric(metrics) {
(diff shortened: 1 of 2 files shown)


mrstegeman:
We should probably also make sure that `message.message` is not undefined when the UI receives it: https://github.com/mozilla-iot/gateway/blob/v0.7.0/static/js/app.js#L128

hobinjk:
That's true, future-proofing is a good idea. Updated accordingly

PR #1802, opened 2019-04-18
Copy over all config files

hobinjk:


hobinjk:
Notably this doesn't restart things to make sure that changes in config files are propagated. At least for dnsmasq the gateway handles starting/stopping so that isn't a concern.

tools/pre-upgrade.sh
```diff
@@ -28,3 +28,5 @@ sudo pip3 install -U "$_url"
 # Upgrade adapt-parser Python package
 _url="git+https://github.com/mycroftai/adapt#egg=adapt-parser"
 sudo pip3 install -U "$_url"
+
+sudo cp -r image/etc/ /
```
mrstegeman: We should probably do some proper permissions.

```bash
sudo chown -R root:root "image/etc"
sudo cp -r "image/etc" /
sudo chown -R $(id -u):$(id -g) "image/etc"
```

hobinjk: :+1:, feel free to continue your PR and obsolete this one :)

mrstegeman: Nah, already deleted it. These were my changes exactly.

hobinjk: Okay, updated this PR with those changes

PR #1803, opened 2019-04-18
Bump version to 0.8.1

hobinjk:
I tested the OTA from 0.7.0 to 0.8.1 and it fixed the captive portal issue. I'll be testing 0.8.0 -> 0.8.1 right after this.

@mrstegeman Any objections to rolling out an OTA tonight?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webthings-gateway",
-  "version": "0.8.0",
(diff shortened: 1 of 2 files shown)


mrstegeman:
It seems like we may still have another issue with the AP starting up when it shouldn't, right?

hobinjk:
Yeah, we can delay the OTA until that underlying issue is fixed

PR #1816, opened 2019-04-22
Handle readOnly property in JSON schema forms.

mrstegeman:


static/js/schema-form/number-field.js
```diff
@@ -108,8 +108,7 @@ class NumberField {
         id="${id}"
         class="form-control"
         ${this.required ? 'required' : ''}
-        ${this.readonly ? 'readonly' : ''}
-        ${this.disabled ? 'disabled' : ''}
+        ${this.disabled || this.readOnly ? 'disabled' : ''}
```
hobinjk: This might be the Monday brain talking, but I'd like parentheses around `this.disabled || this.readOnly` for a bit of clarity

mrstegeman: That's fair. Done.

PR #1819, opened 2019-04-23
Now gets to the end of first time setup for OpenWRT

dhylands:


src/platforms/linux-openwrt.js
```diff
@@ -100,26 +144,27 @@ function uciGet(key) {
...
-
   if (Array.isArray(value)) {
-    let proc = child_process.spawnSync('uci', ['delete', key]);
-    if (proc.status === 0) {
-      for (const v of value) {
-        proc = child_process.spawnSync('uci', ['add_list', `${key}=${v}`]);
-        if (proc.status !== 0) {
-          break;
-        }
+    // Failing to delete a non-existant key is fine.
```
mrstegeman: nit: s/non-existant/nonexistent/ (2 typos)

Ditto on 158.

src/platforms/linux-openwrt.js
```diff
@@ -130,7 +175,196 @@ function uciSet(key, value) {
...
+    '--set-mark', '1',
+  ];
+  let proc = child_process.spawnSync(cmd, args);
+  if (proc.status !== 0) {
+    console.error(`redirectTcpPort: ${cmd} ${args} failed:`,
+                  proc.status);
+    return false;
+  }
+
+  cmd = 'iptables';
```
mrstegeman: `cmd` is not actually changing. Ditto on 229.

src/platforms/linux-openwrt.js
```diff
@@ -130,7 +175,196 @@ function uciSet(key, value) {
...
+  const proc = child_process.spawnSync(cmd, args, {encoding: 'utf8'});
+  if (proc.status !== 0) {
+    console.error(`isRedirectedTcpPort: ${cmd} ${args} failed:`, proc.status);
+    return false;
+  }
+
+  // The line we're looking for should look like this:
+  //  -A PREROUTING -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 8080
+
+  const dst = `--dst ${ipaddr}`;
```
mrstegeman: The `--dst ${ipaddr}` bit isn't in the comment above.

src/platforms/linux-openwrt.js
```diff
@@ -130,7 +175,196 @@ function uciSet(key, value) {
...
+
+function getCaptivePortalStatus() {
+  const result = uciGetAsArray('dhcp.@dnsmasq[0].address');
+  if (!result.success) {
+    return false;
+  }
+
+  for (const value of result.value) {
+    if (value.startsWith('/#/')) {
+      // The pattern /#/IP-ADDRESS is the make all host names resolve
```
mrstegeman: nit: s/is the/is to/

src/platforms/linux-openwrt.js
```diff
@@ -153,7 +387,7 @@ function getDhcpServerStatus() {
  * @param {boolean} enabled - Whether or not to enable the DHCP server
```
mrstegeman: Add the `options` param to the comment.

src/platforms/linux-openwrt.js
```diff
@@ -166,39 +400,47 @@ function setDhcpServerStatus(enabled) {
...
   if (proc.status !== 0) {
+    console.error(`setDhcpServerStatus: ${cmd} ${args} failed`);
     return false;
   }
 
-  proc = child_process.spawnSync(
-    '/etc/init.d/dnsmasq',
-    [enabled ? 'enable' : 'disable']
-  );
+  cmd = '/etc/init.d/dnsmasq';
```
mrstegeman: cmd didn't change.

src/platforms/linux-openwrt.js
```diff
@@ -166,39 +400,47 @@ function setDhcpServerStatus(enabled) {
...
+    console.error(`setDhcpServerStatus: ${cmd} ${args} failed`);
     return false;
   }
 
-  proc = child_process.spawnSync(
-    '/etc/init.d/odhcpd',
-    [enabled ? 'enable' : 'disable']
-  );
-  return proc.status === 0;
+  cmd = '/etc/init.d/odhcpd';
```
mrstegeman: cmd didn't change.

src/router-setup.js
```diff
@@ -160,6 +152,7 @@ function handleCreating(request, response) {
           );
         } else {
           return waitForWiFi(20, 3000).then(() => {
+            DEBUG && console.log('WiFi setup complete');
```
mrstegeman: This log should be something like "router-setup: setup complete"

src/router-setup.js
```diff
@@ -203,18 +196,45 @@ function getHotspotSsid() {
...
   const ssid = getHotspotSsid();
-  return platform.setWirelessMode(true, 'ap', {ssid, ipaddr});
+
+  // We need the LAN to be configured to have a static IP address,
+  // and run dnsmasq (DHCP server and DNS server).
+
+  const ifname = null;
+  const netmask = '255.255.255.0';
+
+  if (!platform.setLanMode('static', {ipaddr, ifname, netmask})) {
```
mrstegeman: Can you prefix all the logs in this function with `router-setup: ` for consistency?

src/router-setup.js
```diff
@@ -226,6 +246,12 @@ function stopAP() {
  *                    the command.
  */
 function defineNetwork(ssid, password) {
+  DEBUG && console.log('router-setup: defineNetwork ssid:', ssid);
+
+  // TODO: eth0 is RPi specific and will be different for each potential
```
mrstegeman: In my opinion, the platform code should handle this itself without having to pass in the interface name. The function is called `setWanMode`, so that code should probably be able to figure out for itself what the WAN interface is.

dhylands: On the turris, there is eth0, eth1, and eth2. eth1 is the WAN. linux-openwrt.js would need to know that it's running on a turris omnia

mrstegeman: Maybe we put that into one of the config files and let the platform code read it.

dhylands:
> When we get things going on the Turris, I suspect this will need to be reworked a bit due to their web interface running on port 80.

Luci also runs on port 80, but it can be configured to run on a different port.

PR #1838, opened 2019-04-26
Updated to work on openwrt master

dhylands:
- added lots more debugging prints (enable using DEBUG = true)
- Changed dhcp servers from start to restart in case they happened
  to already be running
- sets the wan mode to DHCP earlier. This give the NTP server more time
  to sync the time
- Added checkConnection function to the platform code so that non-static
  initialization (i.e. iptables) can be run on every run of the program.

### DIFF
--- a/src/platforms/linux-openwrt.js
+++ b/src/platforms/linux-openwrt.js
@@ -205,9 +205,10 @@ function uciCommit(label, key) {
  * @returns {boolean} Whether or not the command was successful
  */
 function redirectTcpPort(add, ipaddr, fromPort, toPort) {
+  const action = add ? 'Redirecting' : 'Removing redirection';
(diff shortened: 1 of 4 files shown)


src/platforms/linux-openwrt.js
```diff
@@ -483,17 +525,19 @@ function setLanMode(mode, options = {}) {
...
+  const result = uciShow(label, 'network.wan', {errOk: true});
   if (!result.success) {
+    DEBUG && console.log(`${label}: returning`, {mode, options});
     return {mode, options};
   }
 
   for (const [key, value] of Object.entries(result.pairs)) {
     // discard network.wan=interface
-    if (!key.startsWith('network.wan.')) {
+    if (!key.startsWith('network.wana.')) {
```
mrstegeman: typo?

PR #1845, opened 2019-04-29
Allow platform to decide architecture

dhylands:


src/platforms/linux-openwrt.js
```diff
@@ -292,6 +292,17 @@ function isRedirectedTcpPort(ipaddr, fromPort, toPort) {
   return false;
 }
 
+/**
+ * Get Architecture
+ *
+ * @param {string} default architecture detcted by the gateway.
```
mrstegeman: nit: s/detcted/detected/

PR #1854, opened 2019-05-01
Add linear backoff to plugin restarting

hobinjk:
Delays up to 30 seconds before restarting and resets the delay if a crash hasn't occurred within 60 seconds

### DIFF
--- a/src/plugin/plugin.js
+++ b/src/plugin/plugin.js
@@ -50,6 +50,7 @@ class Plugin {
     this.restart = true;
     this.restartDelay = 0;
     this.lastRestart = 0;
+    this.pendingRestart = null;
     this.unloadCompletedPromise = null;
     this.unloadedRcvdPromise = null;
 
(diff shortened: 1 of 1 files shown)


src/plugin/plugin.js
```diff
@@ -553,9 +555,18 @@ class Plugin {
...
+          if (this.lastRestart + 60 * 1000 < Date.now()) {
+            this.restartDelay = 0;
+          }
           console.log('Plugin:', this.pluginId, 'died, code =', code,
-                      'restarting...');
-          this.start();
+                      'restarting after', this.restartDelay);
+          this.lastRestart = Date.now();
+          setTimeout(() => {
+            this.start();
```
mrstegeman: `this.lastRestart` should probably be moved inside the timeout.

hobinjk: :+1:

hobinjk:
I had to switch around a ton of code to make the restarted process available to the test

Issue #1758, opened 2019-04-09
Notifier Add-ons


mrstegeman:
As a developer, I want to write an add-on which can be used to notify a user of something via a rule.

Right now, we have several add-ons that represent notifiers as things, such as:
* activitypub-adapter
* email-sender-adapter
* twilio-adapter

While this is functional, it seems a bit clunky, since these are not really _things_.

Instead, I propose a new type of add-on that presents a simple "Notifier", rather than an "Adapter". The Notifier would probably just need one primary method, `notify`, which would accept a string.

Add-ons would be able to create multiple notifiers, as well. For instance, a user might want to configure a "Slack Notifier Add-on" with multiple channels. Each notifier instance could then be used for different rules.

The existing add-on configuration UI could still be used.

kgiori:
Would the user then be offered a list of configurable notification "channels" in the configuration page of this over-arching notifier add-on? And then the rules engine "notification" block would send the alert to all configured channels? I'm trying to imagine the user experience.

benfrancis:
I imagine it would be a type of add-on, there would be multiple Notifier add-ons which expose a particular kind of notification mechanism (e.g. An SMS add-on or an email add-on). It would make sense for those notification mechanisms to then be used as outputs in the rules engine, yes.

mrstegeman:
The initial back-end for this is done now. We'll need some UI design for the rules engine.

mrstegeman:
This issue is essentially done. Follow up work is in #1973 and #1974.

PR #1857, opened 2019-05-03
Add initial notifier support.

mrstegeman:
References #1758
Requires mozilla-iot/gateway-addon-node#39

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -44,7 +44,9 @@ class AddonManager extends EventEmitter {
   constructor() {
     super();
     this.adapters = new Map();
+    this.notifiers = new Map();
     this.devices = {};
+    this.outlets = {};
     this.deferredAdd = null;
     this.deferredRemove = null;
     this.addonsLoaded = false;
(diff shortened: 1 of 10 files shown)


mrstegeman:
This requires that the gateway-addon changes be pulled in manually.

Also, there are no tests yet. Wanted to make sure the API was ok before doing those.

src/addon-manager.js
```diff
@@ -828,13 +971,22 @@ class AddonManager extends EventEmitter {
...
+    const notifierIds = notifiers.map((n) => n.id);
+
     const unloadPromises = [];
     if (adapters.length > 0) {
       for (const a of adapters) {
         console.log('Unloading', a.name);
         unloadPromises.push(a.unload());
         this.adapters.delete(a.id);
       }
+    } else if (notifiers.length > 0) {
```
hobinjk: I think a single package could expose both an add-on and a notifier if it wanted to

mrstegeman: Yes, it could. I was considering doing that for the email adapter for backwards compat.

mrstegeman:
The rules test failure is occurring on master as well, so it's not due to this PR.

PR #1866, opened 2019-05-07
WIP: Switch to Jasmine test runner from Jest

hobinjk:
We disable 90% of what jest does so we should just use a normal test
runner that doesn't randomly fail to run tests.

This still needs to replace `jest --coverage` but is at least green on travis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6454,6 +6454,22 @@
         "handlebars": "^4.1.0"
       }
     },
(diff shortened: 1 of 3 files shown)


hobinjk:
The tests being green twice in a row when this was a standalone branch was just a really weird fluke since this PR doesn't actually switch Travis to Jasmine

PR #1878, opened 2019-05-14
Add button to control rule animation

hobinjk:
recording

### DIFF
--- a/static/css/rule.css
+++ b/static/css/rule.css
@@ -43,6 +43,10 @@
   background: #777;
 }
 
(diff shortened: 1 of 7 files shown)


hobinjk:
That gif compressed even worse than I thought possible

mrstegeman:
My only other gripe is that the animation doesn't show any difference when changing the if/while selector.

hobinjk:
Thanks for the feedback!  Fixed both concerns:
updated recording

PR #1893, opened 2019-05-17
Simplify control flow

tim-hellhake:


src/addon-loader.js
```diff
@@ -105,32 +80,47 @@ async function loadAddon(addonPath, verbose) {
...
+        }
+      } catch (e) {
+        console.error(e);
+        const message =
         `Failed to start ${manifest.display_name} add-on: ${
           e.toString().replace(/^Error:\s+/, '')}`;
-      return fail(message);
-    }
-  });
+        return fail(addonManagerProxy, message);
```
dhylands: nit: fail doesn't return (it exits the process), so the return statement is never used.

tim-hellhake: I've wondered about that too, but I didn't want to change the existing logic.

tim-hellhake: @dhylands removed it

Issue #1917, opened 2019-05-28
IFTTT Maker PUT Webhook 404


damooooooooooh:
I am trying to set the property of a thing using the api endpoint from an ifttt Webhook using a PUT request.  I can successfully issue the PUT request from Fiddler using the headers

Accept: application/json
Content-Type: application/json

When I remove the Accept header, it fails with 404 and message:

> Cannot PUT /things/thingid/properties/on

I tracked the issue down to router.js line 86

      else if (!request.accepts('html') && request.accepts('json') ||
                 request.get('Upgrade') === 'websocket' ||
                 request.is('multipart/form-data') ||
                 request.path.startsWith(Constants.ADDONS_PATH) ||
                 request.path.startsWith(Constants.INTERNAL_LOGS_PATH)) {
        request.url = API_PREFIX + request.url;
        next();

      // Otherwise treat it as an app request
      } else {
        request.url = APP_PREFIX + request.url;
        next();
      }
    });

It looks like when ifttt sends the request, the request defaults to accept: "*/*" and the code is not handling this use case correctly.  Maybe the else if statement should examine the requested content-type for application/json && put or post request and treat that as an api request rather than falling through to an app html request?

![iftttExample](https://user-images.githubusercontent.com/1326094/58803853-55de4080-8608-11e9-9c32-4be18a7507ea.PNG)

damooooooooooh:
Adjusted router.js as follows, can someone in the know let me know if this approach will cause issues?  Now iftttt webhook functions properly.

`      // If request won't accept HTML but will accept JSON,
      // or has a content-type of json
      // or is a WebSocket request, or is multipart/form-data
      // treat it as an API request
      } else if (!request.accepts('html') && request.accepts('json') ||
                 request.headers['content-type'] === 'application/json' ||
                 request.get('Upgrade') === 'websocket' ||
                 request.is('multipart/form-data') ||
                 request.path.startsWith(Constants.ADDONS_PATH) ||
                 request.path.startsWith(Constants.INTERNAL_LOGS_PATH)) {`

damooooooooooh:
63810604267c283b040b47e14f73d2be89c8bb54

damooooooooooh:
Please see 63810604267c283b040b47e14f73d2be89c8bb54 on my fork that resolves the issue for incoming IFTTT webhook calls. Can someone review and I can issue a pull request?

tim-hellhake:
Just open a PR 😄
Someone will look at it.

[...]

PR #1924, opened 2019-06-03
Fix Handling of JSON PUT without accept header

damooooooooooh:
Fix Handling of JSON PUT without accept header.  Use case linked to IFTTT webhook push failed as IFTT webhook does not add the Accepts: application/json header and sends Accepts: */* which results in improper handling of incoming api json PUT as gui PUT.


damooooooooooh:
https://github.com/mozilla-iot/gateway/issues/1917

PR #1929, opened 2019-06-07
Add basic query-type commands with scaffolding for future work

hobinjk:
I still want to add capability-based queries and log-based queries but thought I'd start the PR process early to see if people think this is a good approach/direction.


src/command-utils.js
```diff
@@ -8,6 +8,66 @@
 
 'use strict';
 
+const booleans = {
+  on: {
+    value: true,
+    '@type': null,
```
mrstegeman: `'@type': 'OnOffProperty'`

Ditto for `off`.

src/controllers/commands_controller.js
```diff
@@ -40,6 +43,42 @@ const Things = require('../models/things');
...
 
+function internalError(response) {
+  response.status(400).json({
+    message: 'Sorry, something went wrong.',
+  });
+}
+
+function thingNotFound(response, name) {
+  response.status(400).json({
+    message: `Sorry, the thing ${name} wasn't found.`,
```
mrstegeman: This sentence is a little awkward.

src/controllers/commands_controller.js
```diff
@@ -40,6 +43,42 @@ const Things = require('../models/things');
...
+
+function invalidForDevice(response) {
+  response.status(400).json({
+    message: 'Sorry, I\'m afraid I can\'t do that.',
+  });
+}
+
+function propertyNotFound(response, thing, property) {
+  response.status(400).json({
+    message: `Sorry, the thing ${thing} has no property ${property}.`,
```
mrstegeman: A bit awkward again. Maybe "Sorry, that thing has no property named ${property}." would make more sense.

src/controllers/commands_controller.js
```diff
@@ -124,17 +132,39 @@ CommandsController.post('/', async (request, response) => {
                                           'level'),
   };
 
+  for (const name in CommandUtils.booleans) {
+    const bool = CommandUtils.booleans[name];
+    if (!bool.value || !bool['@type']) {
```
mrstegeman: I'm guessing you want `!bool.hasOwnProperty('value')`

src/controllers/commands_controller.js
```diff
@@ -203,9 +233,43 @@ CommandsController.post('/', async (request, response) => {
...
+    const verb = bool.verb || 'is';
+    const answer = value ? 'Yes' : 'No';
+    response.status(201).json({
+      message: `${answer}, the ${payload.thing} ${verb} ${valueStr}.`,
+      payload: payload,
+    });
+    return;
+  }
+  response.status(400).json({
+    message: `I don't feel like dealing with thing ${payload.thing} property ${payload.value} today`,
```
mrstegeman: 🤣

flatsiedatsie:
To my untrained eye, this code seems to add support for booleans. Would it be an idea to add support for something like "what level is .."? That would make it easier to query sensors.

hobinjk:
Yes, the idea is to support both boolean and numeric properties which is why I haven't merged this yet. I'd also like to throw in support for string properties and for selecting properties by name (a way of improving the behavior in #1984) but that might be out of scope of this initial implementation.

PR #1946, opened 2019-06-19
Use min/max/unit on MultiLevelSwitch bubble.

### LINKED ISSUE #1905: LevelProperty assumes range between 0 and 100 when highlighted.
weird-thermostat

There are two oddities in the GIF above:
- While the property on the left is a 'degree celsius' item that can go between 0 and 150, the center bubble shows a percentage.
- If the property on the left is set to above 100, the central bubble does not respond.

It could be that I'm using MultiLevel wrong. Here's the initialising code from the MySensors adapter for the item on the left:

```
            elif child.type == 14:                      # Heater
                if new_sub_type == 0: # V_TEMP
                    self._type.append('TemperatureSensor')
                    self.properties[targetPropertyID] = MySensorsProperty(
                        self,
                        targetPropertyID,
                        {
                            '@type': 'TemperatureProperty',
                            'label': new_description,
                            'type': 'number',
                            'unit': 'degree celsius',
                            'readOnly': True,
                        },
                    new_value, new_node_id, new_child_id, new_sub_type)
                if new_sub_type == 45: # V_HVAC_SETPOINT_HEAT
                    self._type.append('MultiLevelSwitch')
                    self.properties[targetPropertyID] = MySensorsProperty(
                        self,
                        targetPropertyID,
                        {
                            '@type': 'LevelProperty',
                            'label': new_description,
                            'minimum':0,
                            'maximum':150,
                            'type': 'number',
                            'unit': 'degree celsius',
                        },
                    new_value, new_node_id, new_child_id, new_sub_type)

                    # TODO: V_HVAC_FLOW_STATE. Mozilla Gateway has no support for a multiple-buttons type yet?

```

[Looking at the schema](https://iot.mozilla.org/schemas/#LevelProperty) I found no indication that a LevelProperty is clamped between 0 and 100.

mrstegeman:
Fixes #1905

### DIFF
--- a/static/js/components/capability/multi-level-switch.js
+++ b/static/js/components/capability/multi-level-switch.js
@@ -100,7 +100,7 @@ class MultiLevelSwitchCapability extends BaseComponent {
     this.level =
       typeof this.dataset.level !== 'undefined' ? this.dataset.level : this.min;
     this.unit =
-      typeof this.dataset.unit !== 'undefined' ? this.dataset.unit : '%';
+      typeof this.dataset.unit !== 'undefined' ? this.dataset.unit : '';
 
     if (typeof this._min === 'undefined') {
       this._min = 0;
--- a/static/js/components/capability/multi-level-switch.js
+++ b/static/js/components/capability/multi-level-switch.js
@@ -92,15 +92,45 @@ class MultiLevelSwitchCapability extends BaseComponent {
   }
 
   connectedCallback() {
+    this._upgradeProperty('min');
+    this._upgradeProperty('max');
+
     // Default to on=true to display level
     this.on = typeof this.dataset.on !== 'undefined' ? this.dataset.on : true;
     this.level =
-      typeof this.dataset.level !== 'undefined' ? this.dataset.level : false;
+      typeof this.dataset.level !== 'undefined' ? this.dataset.level : this.min;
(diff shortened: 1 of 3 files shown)


static/js/components/capability/multi-level-switch.js
```diff
@@ -92,17 +92,47 @@ class MultiLevelSwitchCapability extends BaseComponent {
...
+    this._upgradeProperty('min');
+    this._upgradeProperty('max');
+
     // Default to on=true to display level
     this.on = typeof this.dataset.on !== 'undefined' ? this.dataset.on : true;
     this.level =
-      typeof this.dataset.level !== 'undefined' ? this.dataset.level : false;
+      typeof this.dataset.level !== 'undefined' ? this.dataset.level : this.min;
+    this.unit =
+      typeof this.dataset.unit !== 'undefined' ? this.dataset.unit : '%';
```
hobinjk: I think this should only default to percent if min is 0/undefined and max is 100/undefined. If a developer decides their thing has min 0 max 150 and no unit then the property could display 140%

mrstegeman: Fair point. I was trying to keep backwards compatibility with how things functioned before, but this is the proper thing to do.

PR #1952, opened 2019-06-20
Use title instead of name for devices.

### LINKED ISSUE #1920: "Add by URL..." requires type or @type property
I noticed that when adding a device by manually entering the URL I got an "Invalid thing description" error that I didn't get when the device was discovered automatically. I found that when entering a URL the gateway checks whether a `type` or `@type` property is present in the description, and throws the error if neither is there (see [here](https://github.com/mozilla-iot/gateway/blob/9398e60b6e0f22ac7ef2a11a5c3a7a774f973131/src/controllers/new_things_controller.js#L76)). Adding a `type` or `@type` property to my description got rid of the error as expected based on the code.

However, the current Web Thing API doc says that the "@type member is an optional annotation" and the `type` member no longer exists, so this seems to be a bug.

mrstegeman:
Fixes #1920

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5335,7 +5335,7 @@
       "dev": true
     },
     "gateway-addon": {
-      "version": "git+https://github.com/mozilla-iot/gateway-addon-node.git#714ee0baf0e56917d3383b652a1a53cb5d0cb589",
+      "version": "git+https://github.com/mozilla-iot/gateway-addon-node.git#26464d416f13e2cf936f521287a37865cba9cf20",
       "from": "git+https://github.com/mozilla-iot/gateway-addon-node.git",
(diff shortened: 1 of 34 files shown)


mrstegeman:
Done.

tim-hellhake:
@mrstegeman Why did you changed it from `name` to `title`?
According to the [spec](https://iot.mozilla.org/wot/#thing-resource) a thing has a `name`.
Is this going to be changed?

mrstegeman:
We're trying to converge a bit on the W3C spec, which now uses `title` rather than `name`. Our spec will change soon in accordance.

https://w3c.github.io/wot-thing-description/#sec-td-serialization

PR #1971, opened 2019-06-26
Handle precision via multipleOf for properties.

### LINKED ISSUE #1318: Default precision of number properties

mrstegeman:
Fixes #1318

### DIFF
--- a/static/js/schema-impl/capability/energy-monitor.js
+++ b/static/js/schema-impl/capability/energy-monitor.js
@@ -11,6 +11,7 @@
 'use strict';
 
 const Thing = require('./thing');
+const Utils = require('../../utils');
 
 class EnergyMonitor extends Thing {
   /**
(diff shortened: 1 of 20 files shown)


static/js/schema-impl/capability/energy-monitor.js
```diff
@@ -51,6 +52,22 @@ class EnergyMonitor extends Thing {
...
+
+    this.precision = 0;
+    this.unit = 'watt';
+
+    if (this.powerProperty) {
+      const property = this.displayedProperties[this.powerProperty].property;
+
+      if (property.hasOwnProperty('multipleOf') &&
+          `${property.multipleOf}`.includes('.')) {
+        this.precision = `${property.multipleOf}`.split('.')[1].length;
```
hobinjk: Hmm, nice way to do it!

PR #1990, opened 2019-07-02
Update most dependencies.

mrstegeman:


webpack.config.js
```diff
@@ -61,6 +60,9 @@ const webpackNode = {
 };
 
 const pluginsWeb = [
+  new CleanWebpackPlugin({
+    cleanOnceBeforeBuildPatterns: ['**/*', '!service-worker.js*'],
```
hobinjk: Nice catch! Good to have these ordered properly too

PR #2000, opened 2019-07-08
Add Notifiers controller

hobinjk:
Supports the following paths:
- `/notifiers`: List of all notifiers
- `/notifiers/notifier-id/outlets`: List of notifier's outlets
- POST `/notifiers/notifier-id/outlets/outlet-id/notify`: Calls outlet.notify() with title, message, and level from body

### DIFF
--- a/src/controllers/notifiers_controller.js
+++ b/src/controllers/notifiers_controller.js
@@ -1,0 +1,77 @@
+'use strict';
+
+const PromiseRouter = require('express-promise-router');
+const AddonManager = require('../addon-manager');
+const Constants = require('../constants');
+
+const NotifiersController = PromiseRouter();
+
+NotifiersController.get('/', async (request, response) => {
+  const notifiers = AddonManager.getNotifiers();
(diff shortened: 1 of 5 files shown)


hobinjk:
The rules engine only needs the first two routes but the third one seemed nice to have

Also #2000 \o/

src/controllers/notifiers_controller.js
```diff
@@ -0,0 +1,77 @@
...
+  response.status(200).json(outletList);
+});
+
+/**
+ * Create a new notification with the title, message, and level contained in
+ * the request body
+ */
+NotifiersController.post(`/:notifierId/outlets/:outletId/notify`, async (request, response) => {
+  const notifierId = request.params.notifierId;
+  const outletId = request.params.notifierId;
```
mrstegeman: `request.params.outletId`

tim-hellhake:
> POST /notifiers/notifier-id/outlets/outlet-id/notify

As POST always creates a new resource it should be a noun.
How about `/notifiers/notifier-id/outlets/outlet-id/notification`?

PR #2008, opened 2019-07-11
Add better openwrt platform names

dhylands:


src/platforms/linux-openwrt.js
```diff
@@ -300,7 +328,10 @@ function isRedirectedTcpPort(ipaddr, fromPort, toPort) {
  */
 
 function getPlatformArchitecture(defaultArchitecture) {
-  return `openwrt-${defaultArchitecture}`;
+  if (DISTRIB_ARCH) {
+    return `openwrt-linux-${DISTRIB_ARCH}`;
+  }
+  return `openwrt-unknown-${defaultArchitecture}`;
```
mrstegeman: The fallback should probably be `openwrt-linux-unknown`.

PR #2035, opened 2019-07-23
Remove secondary captive portal

### LINKED ISSUE #2019: Unable to reclaim my tunnel id
Once the Wifi is setup, and you get reconnected, you get automcatically redirected to the tunnel setup.

I then tried to use an existing name, and clicked on the "Click here to reclaim it" link. On my iPhone as soon as I switch away from the captive portal login, it disconnects me from the Wifi and reconnects to my regular Wifi. Of course, once I get the code and go back thru the Wifi Setup, I have to click the "Click Here" link again which generates a new code and the cycle continues.

### LINKED ISSUE #2028: [OpenWrt] Error redirecting after first time setup
Environment:
* Turris Omnia as router
* Mac OS as client

STR:
* Flash 0.9.0-4 Turris Omnia image
* Connect to Wi-Fi access point
* Configure Wi-Fi network
* Re-connect to Wi-Fi network to start 2nd captive portal
* Configure subdomain (I reclaimed an existing one)

Expected:
* Re-directs to subdomain, ideally in a browser window rather than the captive portal window...

Actual:
* "A problem occurred"
Screenshot 2019-07-22 at 17 46 57

* Only option is to click cancel
* gateway.local and 192.168.1.2 do not resolve (Foris is accessible via 192.168.1.2:81)
* Was able to access web server after a reboot, had to do the domain registration step again

Anyone able to reproduce?

### LINKED ISSUE #2034: Disable second captive portal during router first time setup
In #1798 we added a secondary captive portal during router first time setup, after configuring the Wi-Fi network, so that the user doesn't have to manually type gateway.local or 192.168.2.1 into their browser.

Unfortunately we failed to address the second part of Dave's original suggestion around re-claiming domains, which has come back to bite us.

Until we figure out that UX I suggest backing out #1993 or just disabling the second captive portal for the 0.9 release so that this doesn't block the release.

mrstegeman:
Fixes #2019 
Fixes #2028
Fixes #2034

### DIFF
--- a/src/views/creating.handlebars
+++ b/src/views/creating.handlebars
@@ -97,12 +97,13 @@
     <div id="maindiv">
       <h1 id="title">Creating Wi-Fi network...</h1>
       <h3 id="content">
-        Please connect to {{ap_ssid}} with the password you just created.
+        Please connect to {{ap_ssid}} with the password you just created, then
+        navigate to
+        <a class="gateway-link" href="http://{{domain}}.local" target="_blank" rel="noopener">{{domain}}.local</a>
+        or
+        <a class="gateway-link" href="http://{{ap_ip}}" target="_blank" rel="noopener">{{ap_ip}}</a>
+        in your web browser.
       </h3>
-      <div id="warning">
-        Note: If setup doesn't continue automatically, navigate to
-        {{domain}}.local or {{ap_ip}} in your web browser.
-      </div>
     </div>
   </body>
 </html>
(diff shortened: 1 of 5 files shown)


src/views/creating.handlebars
```diff
@@ -97,12 +97,9 @@
     <div id="maindiv">
       <h1 id="title">Creating Wi-Fi network...</h1>
       <h3 id="content">
-        Please connect to {{ap_ssid}} with the password you just created.
+        Please connect to {{ap_ssid}} with the password you just created, then
+        navigate to {{domain}}.local or {{ap_ip}} in your web browser.
```
hobinjk: A minor QoL improvement here would be to make these links. Although I think many implementations of captive portal support close the page faster than a user can click them

mrstegeman: Force pushed an update with links.

PR #2040, opened 2019-07-24
Add default route on Raspbian in AP mode.

mrstegeman:


flatsiedatsie:
What does this do? Will this make it easier to use the Gateway in a permanent non-internet-connected hotspot mode?

mrstegeman:
No, this just restored network connectivity. The Raspberry Pi Zero W came up with no routes at all when testing the release.

PR #2081, opened 2019-08-15
Install mosquitto in base image and after OTA.

### LINKED ISSUE #2077: Install Mosquitto by default
MQTT has become a popular standard for communication in the smart home.

Already multiple add-ons have been created that use it in some capacity.

To avoid add-ons al starting their own MQTT server, it would be great if Mosquitto came pre-installed on the WebThings Gateway.

See:
https://github.com/mozilla-iot/addon-list/issues/388

mrstegeman:
Fixes #2077

### DIFF
--- a/image/prepare-base.sh
+++ b/image/prepare-base.sh
@@ -56,6 +56,7 @@ sudo apt install -y \
   libtool \
   libudev-dev \
   libusb-1.0-0-dev \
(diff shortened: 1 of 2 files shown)


mrstegeman:
@createcandle @flatsiedatsie is the single package sufficient?

PR #2085, opened 2019-08-19
WebSocket /things endpoint

hobinjk:
Allows opening a websocket on /things to get messages from every thing

### DIFF
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -463,21 +463,29 @@ function websocketHandler(websocket, request) {
     });
   }
 
+  let thingCleanups = {};
   function addThing(thing) {
     thing.addEventSubscription(onEvent);
     const onConn = onConnected.bind(this, thing.id);
     thing.addConnectedSubscription(onConn);
 
-    const onRem = () => {
+    const thingCleanup = () => {
       thing.removeEventSubscription(onEvent);
       thing.removeConnectedSubscription(onConn);
+      thing.removeRemovedSubscription(onRem);
+    };
+    thingCleanups[thing.id] = thingCleanup;
+    const onRem = () => {
+      if (thingCleanups[thing.id]) {
+        thingCleanups[thing.id]();
+        delete thingCleanups[thing.id];
+      }
       if (typeof thingId !== 'undefined' &&
           (websocket.readyState === WebSocket.OPEN ||
            websocket.readyState === WebSocket.CONNECTING)) {
         websocket.close();
       }
     };
-    websocket.on('close', onRem);
     thing.addRemovedSubscription(onRem);
   }
 
@@ -516,6 +524,10 @@ function websocketHandler(websocket, request) {
   const cleanup = () => {
     AddonManager.removeListener(Constants.PROPERTY_CHANGED, onPropertyChanged);
     Actions.removeListener(Constants.ACTION_STATUS, onActionStatus);
+    for (const id in thingCleanups) {
+      thingCleanups[id]();
+    }
+    thingCleanups = {};
     clearInterval(heartbeatInterval);
   };
 
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -617,11 +617,11 @@ function websocketHandler(websocket, request) {
       case Constants.REQUEST_ACTION: {
         for (const actionName in request.data) {
           const actionParams = request.data[actionName].input;
-          Things.getThing(thingId).then((thing) => {
+          Things.getThing(id).then((thing) => {
             const action = new Action(actionName, actionParams, thing);
             return Actions.add(action).then(() => {
               return AddonManager.requestAction(
-                thingId, action.id, actionName, actionParams);
+                id, action.id, actionName, actionParams);
             });
           }).catch((err) => {
             sendMessage({
--- a/src/controllers/things_controller.js
+++ b/src/controllers/things_controller.js
@@ -513,6 +513,7 @@ function websocketHandler(websocket, request) {
       sendMessage({
         messageType: Constants.ERROR,
         data: {
+          code: 404,
           status: '404 Not Found',
           message: `Thing ${thingId} not found`,
         },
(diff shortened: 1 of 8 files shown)


src/controllers/things_controller.js
```diff
@@ -395,74 +400,126 @@ ThingsController.ws('/:thingId/', (websocket, request) => {
   const thingId = request.params.thingId;
   const subscribedEventNames = {};
 
-  function sendMessage(message) {
-    websocket.send(message, (err) => {
+  async function sendMessage(message) {
+    // Convert from internal id to external id
```
mrstegeman: I don't understand the "internal" vs "external" ID bit. Can you explain?

src/controllers/things_controller.js
```diff
@@ -395,74 +400,126 @@ ThingsController.ws('/:thingId/', (websocket, request) => {
...
+    }
+
+    const message = {
       messageType: Constants.ACTION_STATUS,
       data: {
         [action.name]: action.getDescription(),
       },
-    }));
+    };
+    if (action.hasOwnProperty('thingId')) {
```
mrstegeman: Is there a reason that actions don't always have a thingId, but events do?

src/controllers/things_controller.js
```diff
@@ -490,26 +551,39 @@ ThingsController.ws('/:thingId/', (websocket, request) => {
...
         data: {
           status: '400 Bad Request',
           message: 'Parsing request failed',
         },
-      }));
+      });
+      return;
+    }
+
+    const id = request.id || thingId;
```
mrstegeman: Line 620 below should probably use `id`, rather than `thingId`.

static/js/models/reopening-web-socket.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  onMessage(event) {
+    for (const listener of this.listeners.message) {
+      listener(event);
+    }
+  }
+
+  reopen() {
+    this.ws.removeEventListener('message', this.onMessage);
```
mrstegeman: Is there ever a case that this would be hit before `onOpen()` is called? If so, you should also remove the `open` listener.

static/js/models/reopening-web-socket.js
```diff
@@ -0,0 +1,110 @@
...
+  send(msg) {
+    return this.ws.send(msg);
+  }
+
+  closePermanently() {
+    this.closing = true;
+    this.listeners = {};
+
+    if (this.ws) {
+      this.ws.removeEventListener('message', this.onMessage);
```
mrstegeman: Remove the `open` listener?

static/js/models/thing-model.js
```diff
@@ -127,39 +127,41 @@ class ThingModel extends Model {
...
       return;
     }
 
-    const wsHref = this.href.href.replace(/^http/, 'ws');
-    this.ws = new WebSocket(`${wsHref}?jwt=${API.jwt}`);
+    if (globalWs) {
+      this.ws = globalWs;
+      this.personalWebSocket = false;
+    } else {
+      const wsHref = this.href.href.replace(/^http/, 'ws');
```
mrstegeman: I don't think this non-global logic is ever used.

src/controllers/things_controller.js
hobinjk: It's described in more detail in later comments but the gist is that all the models/internal things use `virtual-thing-1` as the id and the external thing description uses `https://cool-gateway.mozilla-iot.org/things/virtual-thing-1`

src/controllers/things_controller.js
hobinjk: The two thingId-less actions are 'pair' which is always global and 'unpair' which includes the thingId as part of its input. It's a bit more complex than it should be, unpair definitely could just include the thingId normally

src/controllers/things_controller.js
hobinjk: Nice catch, thanks!

static/js/models/thing-model.js
hobinjk: I thought it would be nice to leave in case the gateway ever connects directly to things but I'll just tear it out and simplify

static/js/models/reopening-web-socket.js
hobinjk: Yeah, I'll improve this logic since I haven't seen anything saying the readyState will always pass through open during some point of the WS's lifecycle

static/js/models/thing-model.js
hobinjk: Although now closePermanently is never called

src/controllers/things_controller.js
```diff
@@ -395,74 +400,126 @@ ThingsController.ws('/:thingId/', (websocket, request) => {
...
+
+  if (typeof thingId !== 'undefined') {
+    Things.getThing(thingId).then((thing) => {
+      addThing(thing);
+    }).catch(() => {
+      console.error('WebSocket opened on nonexistent thing', thingId);
+      sendMessage({
+        messageType: Constants.ERROR,
+        data: {
+          status: '404 Not Found',
```
tim-hellhake: A property like `code` containing only the http code would make it easier to handle errors.

mrstegeman:
I noticed that when I add a new thing (virtual thing), it shows up in a disconnected state and won't change to connected until a browser refresh.

src/controllers/things_controller.js
hobinjk: Nice idea! I've added a `code` property to all the error messages

hobinjk:
I've added support for new things, thanks for pointing that out!

mrstegeman:
This seems to be working much better!

One last question. Would it be straightforward to push added/removed thing messages to _all_ clients? For instance, if I have 2 browsers open and I add a new thing in one of them, I currently have to refresh before things change.

We can definitely do that separately if it's not a small task.

hobinjk:
Unfortunately that's unrelated to the WS connection as far as I can tell. Both windows receive the thing connected message but only the one where the thing was added has the thing ui element

PR #2090, opened 2019-08-21
Allow .tar.gz files to be passed in

dhylands:
This is useful for uploading the medkit files.


dhylands:
Reworked things a bit

PR #2123, opened 2019-09-02
Remove Tunnel Token and allow re-registration to the tunelling service.

abhijitnathwani:
Added helper scripts to change the tunnel registration settings without reflashing the SD card. The `redo-tunnel-registration.sh` is for the time when the user skipped the `tunnel-registration`. And `remove-tunnel-token.sh` if the user wants to re-registration to a new tunnel subdomain.
Based on: https://discourse.mozilla.org/t/need-assistance-to-configure-remote-access-after-initial-setup/44011/2 
and 
https://discourse.mozilla.org/t/remote-tunnel-access-denied/43772/3

Let me know if any changes are required.

### DIFF
--- a/tools/redo-tunnel-registration.sh
+++ b/tools/redo-tunnel-registration.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+MOZIOT_HOME="${MOZIOT_HOME:=${HOME}/.mozilla-iot}"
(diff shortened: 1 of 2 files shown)


mrstegeman:
It would make the most sense to me to just merge these two scripts and remove both keys from the database. Otherwise, looks good.

abhijitnathwani:
Alright. Thanks for the input @mrstegeman . I kept them separate for the use case. I'll merge them. I was also skeptical about deleting the keys which may not exist. I just checked, it does not cause any trouble.

Issue #1831, opened 2019-04-26
NAS ready to use package


raidnet-ms:
I see the work to implement iot in open wrt gone further with the latest relase. I was wondering why not to make a package ready to install on most widespread NAS devices like Synology and Qnap.

They already run node server and the package will be provided on their "app stores" with one click installation.

[Making Synology packages](https://www.synology.com/en-global/support/developer)
[Packages for Qnap app center](https://www.qnap.com/en-au/how-to/knowledge-base/article/how-to-install-qnap-applications-qpkg-files-manually/)

Anyway seems Qnap it's building its own solution called [QIot Suite](https://www.qnap.com/solution/qiot-suite/en-uk/), would be interesting to compare its features.

Anybody ever tryied to make a package for Nas devices?

mrstegeman:
If someone wants to attempt this, that would be great. Our core team just doesn't have the time to do it.

There are lots of ways we could package this:
* Snap
* .deb
* .rpm
* PKGBUILD (Arch Linux)
* Qnap
* Synology
* ..... the list goes on

Right now, we're focused on our Raspbian and OpenWrt images, and we also maintain a [Docker image](https://hub.docker.com/r/mozillaiot/gateway) that should allow the gateway to run essentially anywhere.

popey:
@mrstegeman I'd be happy to take a look at making a PoC snap of the gateway. I work on the Snap Advocacy Team at Canonical, and would love to spend some time to boostrap an easily install-able and distributable package of the Mozilla IoT Gatweway.

mrstegeman:
Please do! Let us know if you need assistance.

raidnet-ms:
On synology you should apply with your company and tgen assign a developer. We cannot apply with your name https://www.synology.com/en-global/support/developer#apply

raidnet-ms:
The reason why i think a NAS package has priority is due to the visibility this project can get from their "appstore".

flatsiedatsie:
I originally thought I wanted this to run on a NAS. The NAS would be the server for all my home things.

I ran Domoticz package on Synology for a while an quickly realized the problem with this approach:

You want your NAS drives to spin down and sleep. This preserves their life span. Using a NAS as a server made my drives wear out quicker.

My advice: run it on a Pi Zero. $20 hardware with flash storage that will save your $200 hard drives.

raidnet-ms:
[...]

Issue #2113, opened 2019-08-27
Transition to Node 10 or 12


mrstegeman:
Node 8 is end-of-life on 2019-12-31, so we should probably move to something newer.

* Node 10
    * EOL on 2021-04-01
* Node 12
    * EOL on 2022-04-01
    * Bluetooth-based adapters will not work yet due to an [issue](https://github.com/abandonware/noble/issues/21)

PR #2124, opened 2019-09-02
snap

mikeroyal:
Added snapcraft.yaml to build [snap](https://snapcraft.io/). This PR addresses issue #1831, though, I still need to make a few tweaks.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -10,21 +10,14 @@ base: core18
 apps:
   gateway:
     command: env TMPDIR=$XDG_RUNTIME_DIR desktop-launch $SNAP/gateway
-    plugs:
-      - desktop
-      - desktop-legacy
+    plugs: 
       - home
-      - x11
       - browser-support
-      - network
-      - gsettings
-      - pulseaudio
-      - opengl
-
+      
 parts:
   gateway:
     plugin: nodejs
-    node-engine: "12.4.0"
+    nodejs-version: "8.16.0"
     build-packages:
       - pkg-config
       - curl 
@@ -50,6 +43,6 @@ parts:
       - libxss1
       - libxtst6
 
-    source: dist/linux-unpacked
+    source: 
     after:
       - desktop-gtk3
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,55 @@
+name: gateway
+version: git
+summary: Things Gateway.
+description: Web of Things gateway.
+
+confinement: strict
+grade: stable
+base: core18
+
+apps:
+  gateway:
+    command: env TMPDIR=$XDG_RUNTIME_DIR desktop-launch $SNAP/gateway
+    plugs:
+      - desktop
+      - desktop-legacy
+      - home
+      - x11
+      - browser-support
+      - network
+      - gsettings
+      - pulseaudio
+      - opengl
+
+parts:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
+name: gateway
```
benfrancis: webthings-gateway

snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+version: git
+summary: Things Gateway.
```
benfrancis: WebThings Gateway

snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+description: Web of Things gateway.
```
benfrancis: Web of Things gateway by Mozilla

snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+    plugs:
+      - desktop
+      - desktop-legacy
+      - home
+      - x11
+      - browser-support
+      - network
+      - gsettings
+      - pulseaudio
+      - opengl
```
benfrancis: What are all these desktop plugs needed for? The gateway is a headless command line application. X11 shouldn't be necessary.

benfrancis:
@popey might also have feedback?

popey:
I couldn't get it to build here, looks like the yaml is unfinished. `node-engine` should be `nodejs-version`, and the `source: dist/linux-unpacked` assumes something exists in that folder, which for a clean install, is empty. @mikeroyal have you tested building this in a clean VM?

mrstegeman:
I had the same question about all of the desktop dependencies.

What is the build process for this?

@benfrancis If the build is straightforward, it shouldn't be difficult to put out a snap release, in the same way we do Docker releases. The main thing we'll be lacking is the extra test time that the regular gateway images get.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+parts:
+  gateway:
+    plugin: nodejs
+    node-engine: "12.4.0"
```
mrstegeman: None of our add-ons are currently built for Node 12. Our images right now use Node 8, so it'd be great if we could stick with that for the time being. That can be changed when #2113 is resolved.

mikeroyal: I was not sure what of node version was being used so thank you for me know I will that fix :+1:

snap/snapcraft.yaml
mikeroyal: Yeah, that's my bad I thought I had removed in a early version. So thank you for catching that I will remove the plus. :+1:

mikeroyal:
hi @popey, I'm still making changes to the .yaml file. So it's still not complete but I will build it in a clean Ubuntu VM when I've made the necessary corrections. :+1:

mrstegeman:
@mikeroyal Curious if you're still working on this?

mikeroyal:
Yes, @mrstegeman I'm still working on this just been busy with my day job. I would be making some updates this week to the snap. :)

mrstegeman:
@mikeroyal Thanks for the update! Looking forward to getting this merged soon.

mrstegeman:
We've got a separate repo going for this now: https://github.com/mozilla-iot/gateway-snap

PR #2126, opened 2019-09-04
Fix a few fluent messages.

mrstegeman:
References https://discourse.mozilla.org/t/subdomain-reclamation-not-working/44658


mrstegeman:
No worries! I'm not sure what's going on yet. I'm wondering if fluent is only doing its thing on first load or something. This at least fixes things in the interim.

PR #2127, opened 2019-09-04
Comment fluent translations and organize.

mrstegeman:


mrstegeman:
I was planning on falling back to the en-US messages when not found in the desired translation, so I think we can leave it in place for now as a starting point.

PR #2128, opened 2019-09-04
Localize templates and other HTML pages.

mrstegeman:


mrstegeman:
> @mrstegeman Just curious - why are the English strings for page titles included in the HTML, but other strings aren't?

I'm not entirely sure. @hobinjk would know better. If they _are_ required to be there, my guess would be that it's because they're part of the `<head>` block, so not part of the DOM.

PR #2130, opened 2019-09-05
recognize pantavisor as a container runtime in run-app.sh

asac:


asac:
@mrstegeman any ETA when you will roll an update to dockerhub so that webthings can work out of the box on pantavisor enabled systems? to your question: checkout docs.pantahub.com the downloads section i guess...

mrstegeman:
We're hoping to get our 0.10 release out next week, so I'll push new Docker images at that point.

PR #2136, opened 2019-09-06
Fix add things icon on mobile Safari.

mrstegeman:
Similar to #2055, just missed a button.

### DIFF
--- a/static/css/things.css
+++ b/static/css/things.css
@@ -22,7 +22,7 @@
 #add-button {
   bottom: 2rem;
   right: 2rem;
-  background-image: url('../optimized-images/add.svg');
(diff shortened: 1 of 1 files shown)


static/css/things.css
```diff
@@ -22,7 +22,7 @@
 #add-button {
   bottom: 2rem;
   right: 2rem;
-  background-image: url('../optimized-images/add.svg');
+  background: no-repeat center/100% url('../optimized-images/add.svg');
```
benfrancis: This is strange, is it working around a bug in Safari? If so, may be worth leaving a comment.

mrstegeman: Yeah, there's some WebKit issue around this. I wish I had an exact pointer for a bug.

PR #2138, opened 2019-09-09
Split manifest parsing out of add-on manager.

mrstegeman:


src/addon-loader.js
```diff
@@ -21,56 +21,17 @@ const PluginClient = require('./plugin/plugin-client');
...
-    console.error(
-      `API mismatch for package: ${manifest.name}\n` +
-      `Current: ${apiVersion} ` +
-      `Supported: ${manifest.moziot.api.min}-${manifest.moziot.api.max}`);
-
-    const err =
-      `Failed to start ${manifest.display_name} add-on: API version mismatch`;
-    return Promise.reject(err);
-  }
+  const packageName = path.basename(addonPath);
```
dhylands: I see you factored the above into addon-utils. Shouldn't there be a call to validate the manifest?

mrstegeman: My thought was this:
1. The add-on loader wasn't really validating anything but the API version. It was primarily reading the manifest in order to get the name and display name.
2. Add-on manager is what ends up creating the addon-loader process.
3. Add-on manager already does full validation.

So, it seemed unnecessary to validate the manifest twice.

dhylands: Seems fair. I think that the important validation is the one done by the addon manager.

PR #2152, opened 2019-09-13
Fixes an issue to enable the add-on on based on the manifest.

abhijitnathwani:
Also fixes an issue where the add-on is loaded as per the saved settings after an **addon-update**.


mrstegeman:
Thanks for the fix!

PR #2155, opened 2019-09-13
Add basic extension add-on framework.

### LINKED ISSUE #2143: Package custom UI in extension add-ons
As an add-on developer I want to package HTML, CSS & JavaScript as part of my add-on which extends the gateway UI

### LINKED ISSUE #2144: Define custom menu entries in extension add-ons
As an add-on developer I want to add an item to the gateway's main menu which can be used to access my custom UI

mrstegeman:
There are a lot of things we can do to create more extension points,
but this at least provides a starting point which already opens up a
lot of possiblities.

Fixes #2143
Fixes #2144

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5688,6 +5688,11 @@
         }
       }
     },
+    "glob-to-regexp": {
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/glob-to-regexp/-/glob-to-regexp-0.4.1.tgz",
+      "integrity": "sha512-lkX1HJXwyMcprw/5YUZc2s7DrpAiHB21/V+E1rHUrVNokkvB6bqMzT0VfV6/86ZNabt1k14YOIaT7nDvOX3Iiw=="
+    },
     "global-modules": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/global-modules/-/global-modules-2.0.0.tgz",
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "express-ws": "^4.0.0",
     "find": "^0.3.0",
     "gateway-addon": "https://github.com/mozilla-iot/gateway-addon-node",
+    "glob-to-regexp": "^0.4.1",
     "http-proxy": "^1.17.0",
     "ip-regex": "^4.1.0",
     "jsonwebtoken": "^8.5.1",
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -48,6 +48,7 @@ class AddonManager extends EventEmitter {
     this.notifiers = new Map();
     this.devices = {};
     this.outlets = {};
+    this.extensions = {};
     this.deferredAdd = null;
     this.deferredRemove = null;
     this.addonsLoaded = false;
@@ -224,6 +225,27 @@ class AddonManager extends EventEmitter {
   getNotifiersByPackageId(packageId) {
     return Array.from(this.notifiers.values()).filter(
       (n) => n.getPackageName() === packageId);
+  }
+
+  /**
+   * @method getExtensions
+   * @returns Returns a Map of the loaded extensions. The dictionary
+   *          key corresponds to the extension ID.
+   */
+  getExtensions() {
+    return this.extensions;
+  }
+
+  /**
+   * @method getExtensionsByPackageId
+   * @returns Returns a Map of loaded extensions with the given package ID.
+   */
+  getExtensionsByPackageId(packageId) {
+    if (this.extensions.hasOwnProperty(packageId)) {
+      return this.extensions[packageId];
+    }
+
+    return {};
   }
 
   /**
@@ -602,6 +624,17 @@ class AddonManager extends EventEmitter {
       throw new Error(`Add-on not enabled: ${manifest.id}`);
     }
 
+    if (manifest.content_scripts && manifest.web_accessible_resources) {
+      this.extensions[manifest.id] = {
+        extensions: manifest.content_scripts,
+        resources: manifest.web_accessible_resources,
+      };
+    }
+
+    if (!manifest.exec && config.get('ipc.protocol') !== 'inproc') {
+      return;
+    }
+
     const errorCallback = (packageId, err) => {
       console.error(`Failed to load add-on ${packageId}:`, err);
     };
@@ -841,6 +874,10 @@ class AddonManager extends EventEmitter {
     if (!this.addonsLoaded) {
       // The add-ons are not currently loaded, no need to unload.
       return Promise.resolve();
+    }
+
+    if (this.extensions.hasOwnProperty(packageId)) {
+      delete this.extensions[packageId];
     }
 
     const plugin = this.getPlugin(packageId);
--- a/src/addon-utils.js
+++ b/src/addon-utils.js
@@ -129,9 +129,11 @@ function validateManifestJson(manifest) {
     version: '',
   };
 
-  if (config.get('ipc.protocol') !== 'inproc') {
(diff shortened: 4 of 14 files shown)


mrstegeman:
This can be tested with: https://github.com/mozilla-iot/example-extension

mrstegeman:
I think that comparing to WebExtensions is a bit tricky in this case. WebExtensions are essentially given a standalone `document` to draw into, so they can insert a full HTML page.

In our case, a menu entry essentially just references a `<section>` element on the existing page. I'm unaware of a good way to insert a full HTML document into a page, other than using an `<iframe>` (which I really didn't want to do).

Also, I'm unsure what the background page is actually used for in a WebExtension. MDN's example just shows it loading scripts, not having content.

One possible workaround would be:
1. Write your `<section>` content in an HTML file.
2. Load that file via JS.
3. Insert content via JS.

benfrancis:
@mrstegeman wrote:
> I think that comparing to WebExtensions is a bit tricky in this case. WebExtensions are essentially given a standalone document to draw into, so they can insert a full HTML page.
> In our case, a menu entry essentially just references a `<section>` element on the existing page. I'm unaware of a good way to insert a full HTML document into a page, other than using an `<iframe>` (which I really didn't want to do).

That's a good point. You could just copy the content of the `<body>` into the content of the `<section>`, but link relations in the `<head>` (e.g. for CSS style sheets) would also have to be manually moved over, which could get messy.

I had the same instinct to avoid using an iframe, but I can't think of any specific problems with that. Can you?

> I'm unsure what the background page is actually used for in a WebExtension. MDN's example just shows it loading scripts, not having content.

A better example is probably a [sidebar_action](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/sidebar_action). [Here](https://github.com/mdn/webextensions-examples/tree/master/menu-demo) is an example browser extension which demonstrates a sidebar provided as an HTML document (it also demonstrates the use of the context [menu API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus) which is interesting).

What if we took a similar approach to sidebars where a custom section is defined in the manifest including its icon, title and HTML content:

**manifest.json**
```
  "gateway_section": {
    "icon": "images/my_section_icon.png",
    "title" : "My Section",
    "content": "my_section.html"
  }
```

HTML content for a custom section could be provided as an HTML file referenced in the manifest. The HTML file could then load its own CSS and JavaScript files, rather than using the `content_scripts` member of the manifest which in browser extensions is really intended for modifying web content from specific domains.

**my_section.html**
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <link rel="stylesheet" href="css/foo.css" />
    <script src="js/bar.js"></script>
  </head>
  <body>
    <h1>My Section</h1>
  </body>
</html>
```

One problem I can think of with using an iframe to embed this content is that no CSS or JavaScript from the wider gateway application would be loaded into the page by default. This might make it tricky to provide APIs via global variables in the style of Web Extension APIs. The extension would have to include a script from the gateway itself.

mrstegeman:
> @mrstegeman wrote:
> 
> > I think that comparing to WebExtensions is a bit tricky in this case. WebExtensions are essentially given a standalone document to draw into, so they can insert a full HTML page.
> > In our case, a menu entry essentially just references a `<section>` element on the existing page. I'm unaware of a good way to insert a full HTML document into a page, other than using an `<iframe>` (which I really didn't want to do).
> 
> That's a good point. You could just copy the content of the `<body>` into the content of the `<section>`, but link relations in the `<head>` (e.g. for CSS style sheets) would also have to be manually moved over, which could get messy.

Yes, exactly. Manually moving scripts and links out of the `<head>` seemed... unpleasant.

> I had the same instinct to avoid using an iframe, but I can't think of any specific problems with that. Can you?

The biggest problem is the inability to access the global gateway APIs that we expose in the UI, and especially the JWT.

Beyond that, I would say that as we add more extensions points, things may get really tricky if everything is isolated in an `<iframe>`. For example, what if we want to have an extension that draws new Web Components? Or what if we want to handle events from the WebSocket -- should each iframe open its own WebSocket?

> > I'm unsure what the background page is actually used for in a WebExtension. MDN's example just shows it loading scripts, not having content.
> 
> A better example is probably a [sidebar_action](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/sidebar_action). [Here](https://github.com/mdn/webextensions-examples/tree/master/menu-demo) is an example browser extension which demonstrates a sidebar provided as an HTML document (it also demonstrates the use of the context [menu API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus) which is interesting).

That makes sense. But in this case, again, the extension gets an entire `document` to draw to.

> What if we took a similar approach to sidebars where a custom section is defined in the manifest including its icon, title and HTML content:
> 
> **manifest.json**
> 
> ```
>   "gateway_section": {
>     "icon": "images/my_section_icon.png",
>     "title" : "My Section",
>     "content": "my_section.html"
>   }
> ```

I'm not necessarily opposed to the icon/title approach here (content discussed below). My biggest concern was localization, but I see that WebExtensions provide a way to localize strings in the manifest.

> HTML content for a custom section could be provided as an HTML file referenced in the manifest. The HTML file could then load its own CSS and JavaScript files, rather than using the `content_scripts` member of the manifest which in browser extensions is really intended for modifying web content from specific domains.
> 
> **my_section.html**
> 
> ```
> <!DOCTYPE html>
> <html>
>   <head>
>     <meta charset="utf-8">
>     <link rel="stylesheet" href="css/foo.css" />
>     <script src="js/bar.js"></script>
>   </head>
>   <body>
>     <h1>My Section</h1>
>   </body>
> </html>
> ```
> 
> One problem I can think of with using an iframe to embed this content is that no CSS or JavaScript from the wider gateway application would be loaded into the page by default. This might make it tricky to provide APIs via global variables in the style of Web Extension APIs. The extension would have to include a script from the gateway itself.

Yes, this is exactly the issue with using the iframe. I really like the idea of having all of the extension APIs pre-loaded into the browser, rather than having each extension load the same JS file. For example, if you have 10 extensions, you're going to load `extension.js` 10 separate times.

flatsiedatsie:
It's just my two cents, but being able to have a fully working html page, with css and js in it, sounds like it would make development a lot easier.

mrstegeman:
@flatsiedatsie I don't disagree with you. How would you recommend we do that, though, without using an `<iframe>`? You can already use arbitrary CSS and JS, and you can inject HTML. The only thing missing is the ability to load an HTML file directly.

mrstegeman:
@benfrancis Check this out and see what you think: https://github.com/mozilla-iot/example-extension/pull/1

mrstegeman:
Yeah, I think it's fundamentally different. If you look at Voice Fill, which does draw content directly to web pages (Google, Yahoo, etc.), it does so through JS, not by loading HTML. I think something like that is a better analogy here.

flatsiedatsie:
I wasn't advocating for an iframe option (although one advantage of that might be security? A rogue add-on could make less of a mess?).

Mostly just that loading in a html file would indeed be nice. Perhaps bake that in, as was suggested. I could imagine, for example, a separate development support tool that merges the html, css and js for quick previews.

I was also thinking that it could be nice to have the example add-on come with some example html built in. A basic container, some buttons. It would help developers get started quickly, and it would help new add-ons conform to the Gateway's style.

mrstegeman:
The example-extension has some HTML and fetches it asynchronously. See [here](https://github.com/mozilla-iot/example-extension).

PR #2166, opened 2019-09-20
Add ability for add-ons to create REST API handlers.

mrstegeman:
Also, allow all extensions to have a generic view to draw to, not
just those that add menu entries.


tim-hellhake:
That's awesome!
I'm really looking forward to using this.

PR #2169, opened 2019-09-23
Add a data directory for add-ons.

mrstegeman:
Instead of creating arbitrary data files all over the user profile
directory, or using the gateway's database, each add-on should use
its own data directory.

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -635,6 +635,16 @@ class AddonManager extends EventEmitter {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


flatsiedatsie:
Will existing add-ons have to change?

mrstegeman:
Ideally, yes. I won't enforce it for a while, though.

I've been working through our add-ons slowly, so please make updates when you have time. Things should go in `~/.mozilla-iot/data/your-addon-name`, rather than `~/.mozilla-iot/config`.

PR #2180, opened 2019-09-27
Wait for add-on to be disabled when toggling.

### LINKED ISSUE #2168: When reloading gpio-adapter, devices are not present
I haven't figured out if this is a problem with the gpio-adapter or the gateway yet, so I thought I would record the bug here to start with.

When the gateway starts up, all of the configured gpio devices seem to work fine. If I go into settings->addons and then disable and re-enable the gpio adapter, then the devices don't seem to get re-registered. They all show as grayed out in the UI and I see errors in the log about non-existant gpio devices. Restarting the gateway makes things work again.

I observed this with both the 0.6.1 and 0.7.0 versions of the gpio-adapter.

mrstegeman:
Fixes #2168

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -594,7 +594,7 @@ class AddonManager extends EventEmitter {
     await this.loadAddon(packageId);
   }
 
-  async disableAddon(packageId) {
+  async disableAddon(packageId, wait = false) {
     if (!this.installedAddons.has(packageId)) {
       throw new Error('Package not installed.');
     }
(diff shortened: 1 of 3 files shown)


src/addon-manager.js
```diff
@@ -594,15 +594,15 @@ class AddonManager extends EventEmitter {
     await this.loadAddon(packageId);
   }
 
-  async disableAddon(packageId) {
+  async disableAddon(packageId, wait = false) {
```
dhylands: Why default to false? There is only one place which this is being called, and thats below where we pass in true.

mrstegeman: Just to retain compatibility with the other function, in case we decide to use disableAddon elsewhere.

PR #2206, opened 2019-10-03
Jokes don't belong in widely-used error messages

hobinjk:
This also highlights the secondary problem that the commands controller doesn't use `@type`. Also the second argument isn't used by propertyNotFound for brevity but might also be nice to include as an improvement

### DIFF
--- a/src/controllers/commands_controller.js
+++ b/src/controllers/commands_controller.js
@@ -55,12 +55,6 @@ function thingNotFound(response, name) {
   });
 }
 
-function invalidForDevice(response) {
-  response.status(400).json({
-    message: 'Sorry, I\'m afraid I can\'t do that.',
-  });
-}
-
 function propertyNotFound(response, thing, property) {
   response.status(400).json({
     message: `Sorry, that thing has no property named ${property}.`,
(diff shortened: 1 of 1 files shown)


src/controllers/commands_controller.js
```diff
@@ -55,12 +55,6 @@ function thingNotFound(response, name) {
   });
 }
 
-function invalidForDevice(response) {
-  response.status(400).json({
-    message: 'Sorry, I\'m afraid I can\'t do that.',
-  });
-}
-
```
benfrancis: R-

I suggest instead changing the message to "Sorry Dave, I\'m afraid I can\'t do that." This has the dual benefits of:
1. Being a more accurate cultural reference
2. Freaking out at least two users I know of

I am half serious.

P.S. Hi James!

mrstegeman:
> This also highlights the secondary problem that the commands controller doesn't use `@type`.

Isn't that what [this code](https://github.com/mozilla-iot/gateway/blob/master/src/controllers/commands_controller.js#L117-L137) is doing?

hobinjk:
that would require remembering the code I wrote, which is demonstrably impossible

src/controllers/commands_controller.js
hobinjk: 🤔 
Greetings!

PR #2215, opened 2019-10-09
Handle add-on update failures more gracefully.

### LINKED ISSUE #2214: Extensions get deleted if update fails
Hi,

I was tweaking around with [extensions](https://github.com/mozilla-iot/example-extension) and I noticed that the update mechanism for the addons does not work quite as expected.

Scenario:
* Load the example-extension
* Start a local `addon-proxy` with a updated version of `extension`.
* To reproduce the issue, I added an invalid `download_url` in the `list.json`, but this would fail even if we have some network issue while downloading the update package.
* Restart the `mozilla-iot-gateway` service to trigger `Checking for updates`.

Result:
* the package is removed first and then the new package is replaced. In the scenario the new package is not downloaded, we lose the `extension`.
* there is no rollback sort of functionalities.

Expected Result:
* If the update package has not been downloaded due to any reason (network issue or incorrect `download_url`) the existing package should not be removed and the functionality should not break.

The scenario works as expected for the [example-adapter](https://github.com/mozilla-iot/example-adapter).

example-extension-update-failure

mrstegeman:
Fixes #2214

### DIFF
--- a/src/addon-manager.js
+++ b/src/addon-manager.js
@@ -31,6 +31,7 @@ const os = require('os');
 const promisePipe = require('promisepipe');
 const fetch = require('node-fetch');
 const {URLSearchParams} = require('url');
+const {ncp} = require('ncp');
 
 const pkg = require('../package.json');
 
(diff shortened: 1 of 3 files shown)


mrstegeman:
They started being generated by `eslint` a couple months ago. I just haven't had time to address them. Most are within the test code.

Issue #1523, opened 2018-11-30
Settings - Unit Preferences


benfrancis:
As a user I want to choose the units I would prefer to see values displayed in.

E.g. degree Celsius vs. Fahrenheit

PR #2227, opened 2019-10-18
Implement UI and backend for localization settings.

### LINKED ISSUE #808: Start/end times of rules changed since summertime
Since summertime started here in The Netherlands, the rules end/start times unexpectedly changed with +1, which is not correct.

Original rule: 

If time is 22:00, turn on lamp

Same rule after summertime started:

If time is 23:00, turn on lamp

Rules should remain the same, regardless if it’s summertime or not

Kind regards,
Gerrit Haan

### LINKED ISSUE #1113: Settings - Location
For #562 we need a section of the settings to allow the user to specify their location and optionally timezone. What I have in mind is:
"Location: [around here](link to external map) [Get]()"
"Timezone: EDT" (with drop-down for manually setting an alternative)
@benfrancis Do you have the time to mock this up? If not I can just implement it and post screenshots on the PR.

### LINKED ISSUE #1530: Settings - Time
It seems the gateway currently [supplies the wrong time](https://github.com/tsathishkumar/MySController-rs/issues/8) (if I understand correctly).

There is no option in the UX for setting the time / clock.

I don't know if [setting a location](https://github.com/mozilla-iot/gateway/issues/1113) / timezone would be a better / enough of a solution.

The issue of Europe stepping away from summer/wintertime is also coming in the median future.

See also:
https://github.com/mozilla-iot/gateway/issues/1113
https://github.com/mozilla-iot/gateway/issues/1428

### LINKED ISSUE #2125: Settings - Language
As a user I want to choose a language so I can view the UI in my native language.

See also:
* https://github.com/mozilla-iot/gateway/issues/1647
* https://github.com/mozilla-iot/gateway/issues/1113

mrstegeman:
Setting the country, timezone, and language are fully implemented,
but the unit settings will require some additional effort to plumb
through the whole UI.

References #1523

Fixes #808
Fixes #1113
Fixes #1530
Fixes #2125


static/js/views/settings.js
```diff
@@ -1517,6 +1575,73 @@ const SettingsScreen = {
...
   },
 
+  showLocalizationSettings: function() {
+    this.localizationSettings.classList.remove('hidden');
+
+    const countrySelect = this.elements.localization.country;
+    countrySelect.innerHTML = '';
+    countrySelect.disabled = true;
+
+    API.getCountry().then((response) => {
```
benfrancis: It would be nice to have a visible disabled state for the dropdown box (perhaps with `background-color: #48779a;`) if this comes back with an error or an empty response.

static/css/settings.css
```diff
@@ -268,6 +274,7 @@
 
 .developer-checkbox-item,
 .experiment-item,
+.localization-item,
```
benfrancis: Nit: This background colour is too dark to form a good contrast with the dropdown. Should be `#5288af`, as per:

```
.network-item,
.wireless-network-item,
.user-item,
.adapter-item,
.addon-item,
.authorization-item,
.domain-item {
  background-color: #5288af;
}
```

static/js/views/settings.js
```diff
@@ -108,8 +113,57 @@ const SettingsScreen = {
...
+      units: {
+        temperature: document.getElementById('localization-settings-units-temperature'),
+      },
+    };
+
+    this.elements.localization.country.addEventListener(
+      'change',
+      (ev) => {
+        API.setCountry(ev.target.value)
+          .catch(console.error);
```
benfrancis: Something we could do a better job of in general is providing user feedback (e.g. via a toast) when an operation like this fails. Not necessarily to address in this patch, but the kind of thing I should really be including in a more full UX specification.

PR #2238, opened 2019-10-26
Use the compat build of Fluent.js to support Microsoft Edge

gucci-on-fleek:
Currently, the current version of Microsoft Edge doesn't work with the gateway. The [readme.md states that Edge 14+ is supported](https://github.com/mozilla-iot/gateway#browser-support). Currently, the regular build of Fluent.js doesn't support Microsoft Edge, and [requires the `compat` build to be used](https://github.com/projectfluent/fluent.js/issues/431). This PR should fix the Edge support.


mrstegeman:
Thanks!

PR #2243, opened 2019-10-30
Send message to add-ons when a device has been saved.

### LINKED ISSUE #2109: Device Paired Callback
Maybe I’ve missed it, but a pairedDevice callback to the adapter after a device has been paired would be convenient.

mrstegeman:
Fixes #2109

### DIFF
--- a/src/plugin/plugin.js
+++ b/src/plugin/plugin.js
@@ -15,6 +15,7 @@ const AdapterProxy = require('./adapter-proxy');
 const APIHandlerProxy = require('./api-handler-proxy');
 const appInstance = require('../app-instance');
 const config = require('config');
+const Constants = require('../constants');
 const db = require('../db');
 const Deferred = require('../deferred');
 const DeviceProxy = require('./device-proxy');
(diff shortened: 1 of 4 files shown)


src/plugin/plugin.js
```diff
@@ -237,16 +239,32 @@ class Plugin {
...
+            MessageType.DEVICE_SAVED_NOTIFICATION,
+            {
+              deviceId: thing.id,
+              device: thing.getDescription(),
+            }
+          );
+        };
+
+        Things.getThings().then((things) => {
+          things.forEach(send);
```
dhylands: Should this filter the things and only send the things which are associated with the adapter? This looks like every adapter will get every thing.

mrstegeman: Ideally, yes, but there's no way to tell when the gateway starts up which things belong to which adapter. Can you think of a way?

For instance, take this scenario:
* User adds a device, `device-1`, via some add-on which doesn't do auto-discovery (like, say, zigbee).
* User restarts gateway.
* Gateway loads `device-1` from database, but add-on never reports it, since there's no autodiscovery.

How could the add-on find out about the saved device, unless we send all devices to every adapter?

I'd also argue that it's trivial for any add-on to just read the database (or use the REST API) and get the list of things, so this doesn't really change the status quo.

PR #2259, opened 2019-11-05
Display a warning if clock is not yet synchronized.

mrstegeman:


src/controllers/settings_controller.js
```diff
@@ -313,6 +314,32 @@ SettingsController.post('/system/actions', auth, (request, response) => {
...
+    synchronized = Platform.getNtpStatus();
+  }
+
+  response.json({
+    statusImplemented,
+    synchronized,
+  });
+});
+
+SettingsController.put('/system/ntp', (request, response) => {
```
benfrancis: nit: I probably would have used a POST here rather than a PUT because PUT implies you're overwriting the ntp resource with this empty content whereas POST can modify the resource in some server defined way.

static/js/setup-subdomain.js
```diff
@@ -34,6 +35,36 @@ function setupForm() {
...
+
+          if (!haveRestartedNtp) {
+            API.restartNtpSync().catch((e) => {
+              console.error('Failed to restart NTP sync:', e);
+            });
+            haveRestartedNtp = true;
+          }
+
+          // poll again in 5 seconds
+          setTimeout(pollNtp, 5000);
```
benfrancis: It's a shame we have to poll here because 5 seconds may actually feel like quite a long time, but I can't think of an obvious alternative given how the underlying system service works.

src/controllers/settings_controller.js
mrstegeman: Fair enough.

static/js/setup-subdomain.js
mrstegeman: I know... I considered making a WebSocket, but the server was just going to be polling anyway.

PR #2290, opened 2019-11-15
Add config file for Pontoon.

mrstegeman:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1596535

### DIFF
--- a/l10n.toml
+++ b/l10n.toml
@@ -1,0 +1,13 @@
+basepath = "."
+
+locales = [
+  "it",
+]
+
+branches = [
+  "master",
+]
+
+[[paths]]
+  reference = "static/fluent/en-US/main.ftl"
+  l10n = "static/fluent/{locale}/main.ftl"


mrstegeman:
@Pike r?

stefanreimers:
Hi @mrstegeman , do I read the bugzilla discussion correctly, the voice component might come as an add-on and eventually support other languages like German (depends on available Common Voice datasets)?

mrstegeman:
@stefanreimers The situation is this: we're trying to decide if the voice/assistant bit should be a part of the gateway itself or pulled out into an add-on, or even a separate application. There are already 2 other voice add-ons.

As for the bit about other languages, my point in the other discussion was just that the API we're using allows for other languages. We're not currently using that feature, though, and our intent parser may or may not be able to handle languages other than English.

So, there are a lot of unknowns around voice right now.

PR #2300, opened 2019-11-19
Add French (fr-FR) localization

lissyx:


mrstegeman:
@lissyx @gucci-on-fleek you can now find this project in Pontoon: https://pontoon.mozilla.org/projects/webthings-gateway/

PR #2309, opened 2019-11-22
Dutch translation fixes from @Fjoerfoks.

mrstegeman:


mrstegeman:
@Fjoerfoks are you happy with the current set of changes? We can do more later, but are these ok?

Fjoerfoks:
> @Fjoerfoks are you happy with the current set of changes? We can do more later, but are these ok?
Yes these are fine for now. Will work on them later on.

PR #2324, opened 2019-11-27
Beaglebone debian dev

cdelston0:
I'm running gateway on a BeagleBone Enhanced board running a stock Debian image, and made these commits to support that.

I realise that the renaming of linux-raspbian -> linux-debian is potentially contentious and may need discussion on whether this is the preferred way to resolve this.

### DIFF
--- a/src/platform.js
+++ b/src/platform.js
@@ -45,8 +45,11 @@ function getOS() {
   }
 
   const proc = child_process.spawnSync('lsb_release', ['-i', '-s']);
-  if (proc.status === 0 && proc.stdout.toString().trim() === 'Raspbian') {
-    return 'linux-raspbian';
+  if (proc.status === 0) {
+    const lsb_release = proc.stdout.toString().trim();
+    if ((lsb_release === 'Debian') || (lsb_release === 'Raspbian')) {
+      return 'linux-debian';
+    }
   }
 
   if (fs.existsSync('/etc/openwrt_release')) {
--- a/src/platform.js
+++ b/src/platform.js
@@ -47,7 +47,7 @@ function getOS() {
   const proc = child_process.spawnSync('lsb_release', ['-i', '-s']);
   if (proc.status === 0) {
     const lsb_release = proc.stdout.toString().trim();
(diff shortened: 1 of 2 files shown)


src/platforms/linux-debian.js
```diff
@@ -158,7 +158,13 @@ function getWirelessMode() {
     mode = 'ap';
     enabled = true;
 
-    const data = fs.readFileSync('/etc/hostapd/hostapd.conf', 'utf8');
+    const data = null;
+    try {
+      data = fs.readFileSync('/etc/hostapd/hostapd.conf', 'utf8');
+    } catch (err) {
+      if (err.code !== 'ENOENT') throw err;
```
mrstegeman: Please use a full block:
```js
if (...) {
  ...
}
```

src/platforms/linux-debian.js
```diff
@@ -158,7 +158,13 @@ function getWirelessMode() {
     mode = 'ap';
     enabled = true;
 
-    const data = fs.readFileSync('/etc/hostapd/hostapd.conf', 'utf8');
+    const data = null;
```
mrstegeman: This should be `let data`, not `const data`.

src/platform.js
```diff
@@ -45,8 +45,10 @@ function getOS() {
   }
 
   const proc = child_process.spawnSync('lsb_release', ['-i', '-s']);
-  if (proc.status === 0 && proc.stdout.toString().trim() === 'Raspbian') {
-    return 'linux-raspbian';
+  if (proc.status === 0) {
+    var lsb_release = proc.stdout.toString().trim();
```
mrstegeman: `const`, not `var`.

src/platform.js
```diff
@@ -45,8 +45,10 @@ function getOS() {
...
+  if ((lsb_release === 'Debian') || (lsb_release === 'Raspbian')) {
```
mrstegeman: Please indent this, and you're missing the closing brace.

cdelston0:
Thankyou for your feedback. Javascript is not my mother tongue!

I believe I have implemented the requested changes, and I hope I haven't mangled the git history too much.

Issue #2326, opened 2019-11-28
en-US localization review


flodolo:
# Major Issues

## Concatenations

This is not OK for localization.

```
add-thing-add-adapters-hint = No new things found. Try
add-thing-add-adapters-hint-anchor = adding some add-ons.
```

```
<p id="add-adapters-hint" class="hidden">
  <span data-l10n-id="add-thing-add-adapters-hint"></span>
  <a id="add-adapters-hint-anchor" href="/settings/addons"
     data-l10n-id="add-thing-add-adapters-hint-anchor"></a>
</p>
```		

You're concatenating two strings with an hard-coded space in-between, one of them is a link, and you're forcing the English structure to all other languages. There should be a way to avoid that, depending on the version of Fluent you're using.

This one was obvious, I'm not sure if I missed others in the file. For example, `new-web-thing-from` seems suspect, but I couldn't figure out how it's used in the code. The `Local Token Page` section is also suspicious, but I can't find any reference in the code.

## Content Exposed for Localization

```
thing-icons-thing-src = /optimized-images/thing-icons/thing.svg
```

Why is this file path exposed for localization?

## Branding

At the beginning of the file there are two messages

```
# Main Title
webthings-gateway = WebThings Gateway

# Wordmark
wordmark =
  .alt = Mozilla IoT
```

Is it correct to assume that they should not be localizable? If that's the case, it would be better to define them as terms.

```
## The following terms must be treated as brand, and kept in English.
##
## They cannot be:
## - Transliterated.
## - Translated.
##
## Declension should be avoided where possible.
##
## Reference: https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark

-webthings-gateway-brand = WebThings Gateway
-mozilla-iot-brand = Mozilla IoT

# Main Title
webthings-gateway = { -webthings-gateway-brand }

# Wordmark
wordmark =
  .alt = { -mozilla-iot-brand }
```

Even if they're not localizable, you can still use them later to make sure the product is localized consistently (only a couple of examples, there are more).

```
login-title = Login — { -webthings-gateway-brand }
signup-title = Create User — { -webthings-gateway-brand }
```

# Medium Issues

## Lack of Comments

The rules section definitely needs some comments. "and" and "or" are lowercase, why? How is "Not" used, given it's uppercase?

The *Authorize Page* section is a mistery to me. How are all those strings used together?

# Minor Issues

## Message IDs

[...]

PR #2335, opened 2019-12-02
Use terms for branding.

mrstegeman:
References #2326

### DIFF
--- a/static/fluent/en-US/main.ftl
+++ b/static/fluent/en-US/main.ftl
@@ -9,14 +9,13 @@
 ## Reference: https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark
 
 -webthings-gateway-brand = WebThings Gateway
--mozilla-iot-brand = Mozilla IoT
 
 # Main Title
 webthings-gateway = { -webthings-gateway-brand }
 
 # Wordmark
 wordmark =
-  .alt = { -mozilla-iot-brand }
+  .alt = { -webthings-gateway-brand }
 
 ## Menu Items
 
--- a/static/fluent/en-US/main.ftl
+++ b/static/fluent/en-US/main.ftl
@@ -1,9 +1,22 @@
+## The following terms must be treated as brand, and kept in English.
+##
+## They cannot be:
+## - Transliterated.
+## - Translated.
+##
+## Declension should be avoided where possible.
+##
+## Reference: https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark
+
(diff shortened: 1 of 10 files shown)


flodolo:
Here's one problem: localization has already started via GitHub, which means some locales already have these strings translated.

In the future, contributions should only happen via Pontoon, and to make similar changes you'll need to update the string ID. 

For this PR, you can either do that (change IDs), or you need to go through all languages and replace all occurrences of `WebThings Gateway` with the corresponding term.

mrstegeman:
@flodolo I already did the latter and replaced the term in all of the translations.

flodolo:
> For this PR, you can either do that (change IDs), or you need to go through all languages and replace all occurrences of `WebThings Gateway` with the corresponding term.

Sorry, tunnel vision at 6AM. That's exactly what you did and that's perfect ;-)

static/fluent/en-US/main.ftl
```diff
@@ -1,9 +1,22 @@
...
+-mozilla-iot-brand = Mozilla IoT
+
 # Main Title
-webthings-gateway = WebThings Gateway
+webthings-gateway = { -webthings-gateway-brand }
 
 # Wordmark
 wordmark =
-  .alt = Mozilla IoT
+  .alt = { -mozilla-iot-brand }
```
benfrancis: Where in the UI is this being used?

mrstegeman: It's the `alt` text for the wordmark image.

mrstegeman: I suppose we should just be using the WebThings Gateway brand.

benfrancis: Yeah I think this should have been changed when the logo was changed

benfrancis: (probably my bad)

mrstegeman: Fixed.

static/fluent/sr/main.ftl
```diff
@@ -1,9 +1,21 @@
...
+##
+## They cannot be:
+## - Transliterated.
+## - Translated.
+##
+## Declension should be avoided where possible.
+##
+## Reference: https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark
+
+-webthings-gateway-brand = { -webthings-gateway-brand }
```
benfrancis: Should this be `-webthings-gateway-brand = WebThings Gateway` ?

mrstegeman: Yes, great catch! Fixed.

PR #2336, opened 2019-12-03
Fix translation issues with some concatentations.

mrstegeman:
References #2326


static/fluent/en-US/main.ftl
```diff
@@ -294,8 +294,7 @@ active = Active
 inactive = Inactive
 
 ## Domain Setup
-reclaim-prompt = It looks like you've already registered that subdomain. To reclaim it
-click-here = click here
+tunnel-setup-reclaim-domain = It looks like you've already registered that subdomain. To reclaim it <a data-l10n-name="tunnel-setup-reclaim-domain-click-here">click here</a>.
```
flodolo: Apostrophes ;-)

mrstegeman: That fix is in a separate PR.

flodolo: But this a string that it's not available in that PR yet, isn't it?

mrstegeman: No, the strings here are not new, just concatenation properly.

flodolo: Right, but merging the PR with the apostrophes fix won't have effect on this one. Rebasing will remove the two existing strings, you'll still need to manually update this string with the correct apostrophe.

mrstegeman: Updated.

PR #2342, opened 2019-12-03
Fix translation issues with complex concatentations.

mrstegeman:
This fix is only for en-US right now. Translators will need to fix
existing translations.

References #2326

### DIFF
--- a/src/views/authorize.handlebars
+++ b/src/views/authorize.handlebars
@@ -25,15 +25,20 @@
       </div>
 
       <form id="authorize-form" method="get" action="allow">
-        <p><b>{{name}}</b> <span data-l10n-id="authorize-would-like"></span>
-          <span class="authorize-select-container">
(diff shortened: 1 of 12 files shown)


mrstegeman:
I believe we're ok to merge this right now, as there's no better alternative for our current design.

The Fluent team suggested that we instead redesign this page such that the &lt;select> is pulled out of the middle of the sentence, since the translation could change based on the selected option.

benfrancis:
> The Fluent team suggested that we instead redesign this page such that the <select> is pulled out of the middle of the sentence

I'm not really sure how this would work...

benfrancis:
I'm not really sure how this would work...

mrstegeman:
I think we’d have a normal prompt, and then below the prompt the user could select monitor or monitor and control. Something of that nature.

benfrancis:
Oh, I see.

PR #2348, opened 2019-12-04
Add basic Darwin platform utilities.

mrstegeman:
There are a lot more things which are possible, but most require
sudo access.

### DIFF
--- a/src/platforms/darwin.js
+++ b/src/platforms/darwin.js
@@ -1,0 +1,147 @@
+/**
+ * Darwin (macOS) platform interface.
+ *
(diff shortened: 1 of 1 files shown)


mrstegeman:
My primary goal was to just make the UI more functional on Mac, especially with regard to the new settings pages. This will at least show your current time zone, it just won't let you set it.

I had no intention of pull in any network settings. Right now, that only makes sense for Raspbian and OpenWrt.

mrstegeman:
It also has the added benefit of getting rid of the "failed to import platform utilities" log line. 😄

PR #2358, opened 2019-12-09
Allow updates to be disabled.

### LINKED ISSUE #565: Allow disabling OTA updates
As a privacy-conscious user I want to make sure that my gateway never updates by disabling OTA updates.

gucci-on-fleek:
Some platforms (Docker, package managers) expect to handle all updates. If the gateway updates itself, it could lead to inconsistencies. I've added a config flag to disable updates so that the native package manager (Docker, apt, pacman, yum, etc.) can handle them itself.

Fixes #565


static/js/views/settings.js
```diff
@@ -1742,7 +1745,11 @@ const SettingsScreen = {
         statusText += fluent.getMessage('never');
       }
 
+      if (support.support) {
```
mrstegeman: nit: indent line 1749

gucci-on-fleek:
Alright, now the default is to disable updates, and updates are only enabled on Raspbian.

mrstegeman:
Please merge https://github.com/gucci-on-fleek/gateway/pull/1

gucci-on-fleek:
Merged

PR #2381, opened 2019-12-20
Format rule time based on locale.

### LINKED ISSUE #675: Format rule time based on locale
The time input in en_US is 1:00 PM but will show up in the rule's summary as 13:00. The summary should do the same locale checking as the time input.

mrstegeman:
Fixes #675

### DIFF
--- a/static/js/rules/Rule.js
+++ b/static/js/rules/Rule.js
@@ -5,6 +5,7 @@
  */
 
 const API = require('../api');
+const App = require('../app');
 const RuleUtils = require('./RuleUtils');
 const Units = require('../units');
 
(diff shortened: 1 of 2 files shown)


madb1lly:
Hi @benfrancis,
I'm not sure exactly what will be used as a result of this PR, but I often access my gateway from a different timezone and I want all times etc to be displayed based on where the gateway is, not where the browser is (or more correctly what timezone/locale they are set to, as this may not be the same as where they are geographically). Showing browser time instead of gateway time would be confusing for me.
Cheers :)

mrstegeman:
@benfrancis That's a great point. I updated the PR to use the currently set language.

@madb1lly The times will still be displayed according to the gateway's time zone. This only changes the display format.

PR #2387, opened 2019-12-26
Route JSON data to API endpoints.

### LINKED ISSUE #1917: IFTTT Maker PUT Webhook 404

mrstegeman:
Fixes #1917

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -86,11 +86,12 @@ const Router = {
         next();
 
       // If request won't accept HTML but will accept JSON,
-      // or is a WebSocket request, or is multipart/form-data
+      // or is a WebSocket request, or is multipart/form-data or JSON,
       // treat it as an API request
       } else if (!request.accepts('html') && request.accepts('json') ||
                  request.headers['content-type'] === 'application/json' ||
                  request.get('Upgrade') === 'websocket' ||
+                 request.is('json') ||
                  request.is('multipart/form-data') ||
                  request.path.startsWith(Constants.ADDONS_PATH) ||
                  request.path.startsWith(Constants.INTERNAL_LOGS_PATH)) {


src/router.js
```diff
@@ -86,11 +86,12 @@ const Router = {
         next();
 
       // If request won't accept HTML but will accept JSON,
-      // or is a WebSocket request, or is multipart/form-data
+      // or is a WebSocket request, or is multipart/form-data or JSON,
       // treat it as an API request
       } else if (!request.accepts('html') && request.accepts('json') ||
                  request.headers['content-type'] === 'application/json' ||
                  request.get('Upgrade') === 'websocket' ||
+                 request.is('json') ||
```
benfrancis: Is this not equivalent to line 92?

mrstegeman: ... yes. guess that issue was already fixed at some point.

PR #2395, opened 2020-01-07
Allow insecure connections on the local network.

### LINKED ISSUE #2394: Allow plain HTTP access via local network
Following on from the [discussion](https://discourse.mozilla.org/t/local-non-ssl-connection-to-gateway/51262/7) on Discourse. Please correct me if I've misunderstood any of this.

Currently, once a *.mozilla-iot.org subdomain has been configured all HTTP traffic is redirected from port 80 to port 443 via HTTPS using the TLS certificate generated for that domain by LetsEncrypt.

For fully qualified domain names via the tunnelling service this makes sense and is what we want because users should always use HTTPS when available, but when accessing the gateway via gateway.local or a local IP address the gateway will always return the certificate for the *.mozilla-iot.org subdomain so the user sees a certificate error and has to set an exception in their browser in order to proceed.

Given that this certificate will never be valid for gateway.local and will always generate an exception, is this actually what we want? Would it be better to special case the configured .local domain (and local IP address if that can be detected) and just allow access over plain HTTP? Access via the fully qualified domain name would still be redirected as normal and the user could still manually type https://gateway.local if they want to, which could either return a self-signed certificate or a *.mozilla-iot.org certificate and display the certificate warning as before.

mrstegeman:
Fixes #2394


mrstegeman:

tim-hellhake:
Looks good to me.
Just one question.
Why Did you remove the 307 redirect?

mrstegeman:
Because the hosts matched in that case are no longer redirected to https. They're just allowed on http.

PR #2401, opened 2020-01-13
Ensure that all WebSockets are closed when unloading.

### LINKED ISSUE #1895: Icons often show as dimmed even though everything works fine
We saw this numerous times at maker faire. All of the icons the yo works show as dimmed. While in this state we could intact with them fine.

mrstegeman:
Fixes #1895


mrstegeman:

PR #2403, opened 2020-01-13
Send property status when WebSocket connects.

mrstegeman:
Amongst other things, this fixes an issue where:
* A PWA (or the page in a mobile browser) is put to sleep for a while
* The page is resumed
* Some/all property statuses are invalid due to missed propertyStatus
  messages
* In the PWA, it's impossible to refresh, so the app has to be killed
  and reopened

References mozilla-iot/wot#95


mrstegeman:

PR #2411, opened 2020-01-30
Drop commands API.

mrstegeman:
This includes the intent server and the API itself.


mrstegeman:

kgiori:
wait, what? does this mean that the voice controller (using Snips) will no longer work? please don't break that until we have a solid history of an alternative that works just as well.

mrstegeman:
@lissyx and @andrenatal are working to combine their DeepSpeech and Snips add-ons into one, with an option to select which back-end to use. They will pull the intent parsing and such into the add-on. That way, functionality can be added in a much easier fashion. This is similar to how the Voco add-on works.

kgiori:
ah, I see. happy to hear about the collaboration!

mrstegeman:
@andrenatal FYI

flatsiedatsie:
Wait what? Voco no longer works in 0.11??

PR #2417, opened 2020-02-03
Add wiringpi dependency.

mrstegeman:


mrstegeman:
@tim-hellhake this is an easy resolution to the rf433-adapter issue.

PR #2444, opened 2020-02-26
[doc] use `npm ci` to restrict dependencies to lockfile

Swaagie:
Rather than installing dependencies that are allowed within semver ranges of dependencies specified in `package.json`. Use `npm ci` to restrict installation to the dependencies specified in `package-lock.json`. This should prevent unexpected issues arising from installs/builds from source for dependencies that are known to work.


mrstegeman:
Thanks!

PR #2475, opened 2020-04-07
Enable maskable icons for PWA.

### LINKED ISSUE #2474: Maskable webmanifest icon
Please add the `"purpose":  "maskable" ` property to the icons in the webmanifest to enable maskable icons aka. adaptive icons on android.

mrstegeman:
Fixes #2474

### DIFF
--- a/static/app.webmanifest
+++ b/static/app.webmanifest
@@ -8,12 +8,14 @@
     {
       "src": "/images/icon.png",
       "sizes": "256x256",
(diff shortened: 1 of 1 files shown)


lazylazyllama:
Actually: If you test the current icons with https://maskable.app/

You get this as your safe area:
image

So probably it would be better to scale the maskable icons a bit down.

PR #2486, opened 2020-04-26
fix(addon-loader) Print an error when `npm-link` fails

Hywan:
When installing the gateway on my setup, `npm link` failed because of a permission issue. The error was silent. With this patch, the error appears, and it could help the user to fix the issue.


Hywan:
Done.

PR #2488, opened 2020-04-27
Allow extension pages to be reloaded

tim-hellhake:


static/js/extension.js
```diff
@@ -32,6 +32,18 @@ class Extension {
...
   }
 
+  /**
+   * Load your html page here
+   *
+   * @returns {Promise} Promise which should resolve as soon as the extension
+   *                    is ready to be shown
+   */
+  load() {
+    return new Promise((resolve) => {
```
mrstegeman, line 42: `return Promise.resolve()` would be simpler.

static/js/app.js
```diff
@@ -288,14 +299,28 @@ const App = {
...
+      // wait until the extensions is loaded before showing it
+      extension.load()
+        .then(() => {
+          extension.show(context);
+          this.extensions[extensionId].show(context);
+          this.selectView(
+            `extension-${Utils.escapeHtmlForIdClass(extensionId)}-view`
+          );
+          this.displayedExtension = extensionId;
+        });
```
mrstegeman, line 312: You should add a `catch()` that will just redirect to `/things`.

static/js/app.js
```diff
@@ -288,14 +299,28 @@ const App = {
...
-      this.extensions[extensionId].show(context);
-      this.selectView(
-        `extension-${Utils.escapeHtmlForIdClass(extensionId)}-view`
-      );
-      this.displayedExtension = extensionId;
+      const extension = this.extensions[extensionId];
+      // wait until the extensions is loaded before showing it
+      extension.load()
+        .then(() => {
+          extension.show(context);
```
mrstegeman, line 306: Why the double `show()`?

tim-hellhake: I removed the second `show` call.

static/js/app.js
tim-hellhake: I added the `catch` accordingly.

static/js/extension.js
tim-hellhake: Done

mrstegeman:
For anyone seeing this, here's an example for compatibility before and after this change: https://github.com/mozilla-iot/example-extension/commit/97f57a26de4a67738ed3f83919e56a90dfebdfe8

PR #2515, opened 2020-06-09
Add udev rule for the co2-monitor-adapter

tim-hellhake:


image/stage3/02-systemd-units/files/etc/udev/98-co2mon.rules
```diff
@@ -0,0 +1 @@
+SUBSYSTEM=="usb", ATTRS{idVendor}=="04d9", ATTRS{idProduct}=="a052", GROUP="plugdev", MODE="0666"
```
mrstegeman, line 1: `MODE="0660"` should be sufficient, since `pi` is already part of the `plugdev` group by default.

PR #2519, opened 2020-06-12
Enable SPI

tim-hellhake:


tim-hellhake:
The option just activates the SPI device.
Unless someone is relying on an error to occur while trying to open an SPI device there should be no side effects.

PR #2520, opened 2020-06-14
Add search box for available add-ons

### LINKED ISSUE #1932: Add-ons Search
As a user I'd like to search for an add-on (by its name and description) so that I can more quickly find the add-on I'm looking for.

The add-ons directory is starting to get quite long, we are soon going to need a way to search it!

tim-hellhake:
Adds a search box for case insensitive client-side filtering of installable add-ons.
The supplied input is matched against the name and the description of the add-on.

fixes #1932

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -386,6 +386,9 @@
         <section id="addon-config-settings" class="hidden settings-subsection">
         </section>
         <section id="addon-discovery-settings" class="hidden settings-subsection">
+          <div class="section-title-container">
+            Search: <input id="discover-addons-search"></input>
+          </div>
           <ul id="discovered-addons-list">
           </ul>
         </section>
--- a/static/index.html
+++ b/static/index.html
@@ -387,7 +387,7 @@
         </section>
         <section id="addon-discovery-settings" class="hidden settings-subsection">
           <div class="section-title-container">
-            Search: <input id="discover-addons-search"></input>
+            <input id="discover-addons-search" data-l10n-id="addon-search"></input>
           </div>
           <ul id="discovered-addons-list">
           </ul>
--- a/static/fluent/en-US/main.ftl
+++ b/static/fluent/en-US/main.ftl
@@ -458,6 +458,8 @@ addon-discovery-added = Added
 addon-discovery-add = Add
 addon-discovery-installing = Installing…
 addon-discovery-failed = Failed
+addon-search =
+    .placeholder = Search
 
 ## Page Titles
 
--- a/static/js/views/settings.js
+++ b/static/js/views/settings.js
@@ -1676,7 +1676,8 @@ const SettingsScreen = {
 
     const searchText = this.discoverAddonsSearch.value.toLowerCase();
 
-    const matches = (x) => x.name.toLowerCase().indexOf(searchText) > -1 ||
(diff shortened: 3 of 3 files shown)


tim-hellhake:
The input element looks quite bad right now.
I focused on the functionality as there seems to be a good proposal in #1932.

mrstegeman:
Thoughts:
1. The search should probably be cleared each time the page is shown.
2. What do you think about adding a type selector, e.g. All/Adapter/Notifier/Extension? This could also be done by supporting keywords in the search field, e.g. "type:adapter".
3. Could you also let it search the `id` field? That comes in handy sometimes. Maybe also `author`, but that's less important.
4. The [addon-proxy](https://github.com/mozilla-iot/addon-proxy) actually supports server-side searching. However, it's probably not super important to use that at this time, when there's still a relatively small number of add-ons.

tim-hellhake:
> The search should probably be cleared each time the page is shown.

Indeed.

> What do you think about adding a type selector, e.g. All/Adapter/Notifier/Extension? This could also be done by supporting keywords in the search field, e.g. "type:adapter".

Definitely a good idea but do we really need a keyword?
If we include the addon type in the search, you could simply type in "notifier" to find all notifiers.
It's not 100% accurate, but it _just works_.

> Could you also let it search the id field? That comes in handy sometimes. Maybe also author, but that's less important.

Sure, I'll add that.

> The addon-proxy actually supports server-side searching. However, it's probably not super important to use that at this time, when there's still a relatively small number of add-ons.

What I like about client-side filtering is the absence of latency.
As long as the complete list is shown when opening the page, I see no immediate advantage of service-side filtering.

mrstegeman:
> Definitely a good idea but do we really need a keyword?
> If we include the addon type in the search, you could simply type in "notifier" to find all notifiers.
> It's not 100% accurate, but it _just works_.

Fine by me.

> What I like about client-side filtering is the absence of latency.
> As long as the complete list is shown when opening the page, I see no immediate advantage of service-side filtering.

Yeah, that's totally fine. If we get to a point where the add-on list is massive, we can consider switching to server-side filtering.

mrstegeman:
I like it! Are you ready to move this out of draft status? I can do the CSS work separately.

tim-hellhake:
> Are you ready to move this out of draft status?

Except for the UI, it's finished.

> I can do the CSS work separately.

That would be awesome!

static/index.html
```diff
@@ -386,6 +386,9 @@
         <section id="addon-config-settings" class="hidden settings-subsection">
         </section>
         <section id="addon-discovery-settings" class="hidden settings-subsection">
+          <div class="section-title-container">
+            Search: <input id="discover-addons-search"></input>
```
mrstegeman, line 390: The "Search" string should probably just be a placeholder for the input, and it needs to be pulled into the Fluent translation. So, the HTML should be this:

```html
<input id="discover-addons-search" data-l10n-id="addon-search"></input>
```

Then, in `static/fluent/en-US/main.ftl`, in the "Add-on Settings" section, you should have:

```
addon-search =
    .placeholder = Search
```

tim-hellhake: Done

PR #2548, opened 2020-07-24
Open license in same window

flatsiedatsie:
For touch screen kiosk users, opening a new tab can lead to complications. The 'back' button no longer works. And since the license is shown as a text file, there is no 'close' button to close the new tab.

If the license if shown in the same tab, then the 'swipe to go back' funtionality still works, and users can get back to the UI.

Perhaps an even more optimal solution would be to load the license in an overlay.

### DIFF
--- a/static/js/views/installed-addon.js
+++ b/static/js/views/installed-addon.js
@@ -80,7 +80,7 @@ class InstalledAddon {
             ${fluent.getMessage('by')} <a href="${this.homepageUrl}" target="_blank" rel="noopener">${Utils.escapeHtml(this.author)}</a>
           </span>
           <span class="addon-settings-license">
-            (<a href="${this.licenseUrl}" target="_blank" rel="noopener">license</a>)
+            (<a href="${this.licenseUrl}" rel="noopener">license</a>)
           </span>
         </div>
         <div class="addon-settings-controls">


flatsiedatsie:
(In kiosk mode, the external links to github are simply blocked. The user sees a small alert at the top of the window that this is not permitted.)

mrstegeman:
We specifically did this to keep users from navigating out of the "normal" gateway UI in a tab, so all external things are opened in new tabs, including:
* Internal logs
* Licenses for installed and available add-ons
* The Wi-Fi setup page when it posts a link to `gateway.local`
* Links to "more information" sent by add-ons during pairing/unpairing
* External thing links, e.g. those added by the Spotify, Weather, Speed Test, and other add-ons.

Since this is for a very specific use-case, maybe it would make more sense to create a "kiosk" extension that fixed up these links to open in the same tab?

flatsiedatsie:
That's an option, sure.

How about I reframe it. Forget the kiosk thing. 

The current solution to open raw license and log text files in a new window is, in my opinion, a sub-optimal design. It feels..  unfinished. It would be nicer ("prettier") if the gateway could open license files an overlay inside the UI itself. It would also be less confusing to users, as they may not expect a new window to be opened, and may not understand how to navigate between the two. Similarly, if the internal logs were part of a page in the gateway, it would become easy to add extra features op top. For example, to create an extension addon that changes the UI of that page to add filtering of the internal logs, so you can quickly find messages from a particular addon. Or to only show the latest 20 lines, so you can check that's going on 'live' instead of having to refresh a large raw file. It would also make it easier to implement some level of access control to a log that can contain a lot of sensitive information.

mrstegeman:
> The current solution to open raw license and log text files in a new window is, in my opinion, a sub-optimal design. It feels.. unfinished.

While I don't disagree, this is actually pretty common with most products I own which have a web interface. The open source licenses open in a standalone page.

> It would be nicer ("prettier") if the gateway could open license files an overlay inside the UI itself.

I agree.

> It would also be less confusing to users, as they may not expect a new window to be opened, and may not understand how to navigate between the two.

I would say that pretty much all users (especially users of this software) know how to operate a web browser at this point.

> Similarly, if the internal logs were part of a page in the gateway, it would become easy to add extra features op top. For example, to create an extension addon that changes the UI of that page to add filtering of the internal logs, so you can quickly find messages from a particular addon. Or to only show the latest 20 lines, so you can check that's going on 'live' instead of having to refresh a large raw file. It would also make it easier to implement some level of access control to a log that can contain a lot of sensitive information.

I agree.

To sum everything up, I don't like the current approach in these 2 PRs, but would welcome a change which opens things in an overlay or something else more integrated with the UI.

flatsiedatsie:
> I would say that pretty much all users (especially users of this software) know how to operate a web browser at this point.

You should meet my parents :-)

To me, the WebThings Gateway's strength is that it's so simple and user friendly. My parents could get used to it if I gave them a kiosk-type device that runs it.

>  don't like the current approach in these 2 PRs, but would welcome a change which opens things in an overlay or something else more integrated with the UI.

I agree that would be optimal. I'll look into it.

mrstegeman:
I'll close these PRs and let you open another if you get to it.

PR #2567, opened 2020-08-03
Simplify capability code

tim-hellhake:


static/js/components/capability/capabilities.js
```diff
@@ -0,0 +1,56 @@
+/**
```
mrstegeman, line 1: The `components` directory seems like a bit of a strange place for this.

static/js/views/new-thing.js
```diff
@@ -67,71 +69,13 @@ class NewThing {
...
-          break;
-        case 'Custom':
-        default:
-          cls = cls || (capabilities.length > 1 ? '' : 'custom-thing');
-          break;
+
+      if (!cls) {
+        cls = getClassFromCapability(capability);
+
+        if (!cls && capabilities.length == 1) {
```
mrstegeman, line 76: This should be `<= 1` to match the previous logic.

static/js/views/new-thing.js
```diff
@@ -373,74 +317,11 @@ class NewThing {
...
-      case 'Lock':
-        cls = 'lock';
-        break;
-      case 'Custom':
-        this.customIconLabel.classList.remove('hidden');
-        this.customIcon.classList.remove('hidden');
-        break;
-      default:
-        break;
+    const cls = getClassFromCapability();
```
mrstegeman, line 320: You should probably pass something to the function. 😄

static/js/views/new-web-thing.js
```diff
@@ -111,74 +113,11 @@ class NewWebThing {
...
-      case 'Lock':
-        cls = 'lock';
-        break;
-      case 'Custom':
-        this.customIconLabel.classList.remove('hidden');
-        this.customIcon.classList.remove('hidden');
-        break;
-      default:
-        break;
+    const cls = getClassFromCapability();
```
mrstegeman, line 116: Missing function argument.

static/js/views/new-web-thing.js
```diff
@@ -309,73 +248,13 @@ class NewWebThing {
...
-            cls = cls || (capabilities.length > 1 ? '' : 'custom-thing');
-            break;
-          default:
-            cls = cls || (capabilities.length > 1 ? '' : 'custom-thing');
-            break;
+
+        if (!cls) {
+          cls = getClassFromCapability(capability);
+
+          if (!cls && capabilities.length == 1) {
```
mrstegeman, line 255: `<= 1`

static/js/components/capability/capabilities.js
mrstegeman: Maybe this code should just be merged with the other new file?

static/js/views/new-thing.js
tim-hellhake: If assumed that if the length is `< 1`, the for loop will not be entered.
I omitted the condition because I considered it dead code.

tim-hellhake: I assumed that if the length is < 1, the for loop will not be entered.
I omitted the condition because I considered it dead code.

static/js/views/new-thing.js
tim-hellhake: Damn, I miss my compiler :sweat_smile:

static/js/components/capability/capabilities.js
tim-hellhake: I placed it there because all the custom element definitions are located there.
But I think it makes sense to merge it with the other file as it also translates the capability into something else.

PR #2570, opened 2020-08-05
Fixed list_prots.js to match latest SerialPort

dhylands:


tools/list_ports.js
```diff
@@ -15,11 +15,7 @@ function extraInfo(port) {
   return output;
 }
 
-SerialPort.list((error, ports) => {
-  if (error) {
-    console.log('Error:', error);
-    return;
-  }
+SerialPort.list().then((ports) => {
```
mrstegeman, line 18: The `comName` stuff below should be changed to `path` as well.

davehylands:
I now get this output when run on an RPi 3 with no dongles plugged in:
```
Serial Device found @ /dev/ttyAMA0
Serial Device found @ /dev/ttyS0
```

PR #2573, opened 2020-08-22
Add properly sized maskable icons

### LINKED ISSUE #2474: Maskable webmanifest icon

lazylazyllama:
Fixes #2474


benfrancis:
Nit: These icons are not as high quality as the originals. @lazylazyllama How did you generate them? Would you like me to generate some new ones from the SVG source? (We could also add an SVG icon with a size of "any").

lazylazyllama:
> Nit: These icons are not as high quality as the originals. @lazylazyllama How did you generate them? Would you like me to generate some new ones from the SVG source? (We could also add an SVG icon with a size of "any").

I see your point I will create some new ones with inkscape then and also add a svg :)

benfrancis:
@lazylazyllama No need to create a new one, here is the original SVG https://drive.google.com/file/d/1MLENWcjol2WS5YGRNV8elda5fjIlxGN9/view?usp=sharing

lazylazyllama:
> @lazylazyllama No need to create a new one, here is the original SVG https://drive.google.com/file/d/1MLENWcjol2WS5YGRNV8elda5fjIlxGN9/view?usp=sharing

Thank you, sorry for confusing you, what I meant is that I change it to a new one with properly padding around so that it can be used as a maskable icon just like the new ones etc.

lazylazyllama:
@benfrancis I added your icons + the new maskable files in #2578

PR #2578, opened 2020-08-24
Add higher quality app icons and svgs

lazylazyllama:
Followup of #2573


lazylazyllama:
The maskable svg is a bit smaller then the original one, because I generated it with inkscape, but I decided to keep the original one as is

benfrancis:
Thanks @lazylazyllama, much better :)

PR #2587, opened 2020-09-07
Unhide delete button on rule page

### LINKED ISSUE #564: Allow deleting Rule in Rule page
As a user, I want to be able to delete the rule I'm working on from the edit page instead of having to return to the rules list.

Etzos:
The existing codes seems to be handling this correctly, so this simply
removes the CSS hiding the delete button and makes sure that the display
does not hang around for the next rule.

I saw this was listed as a good first bug and jumped in. If this was an inappropriate task to pick up or there's something else I should have worked on let me know, I'm more than happy to shift gears.

resolves #564

### DIFF
--- a/static/css/rule.css
+++ b/static/css/rule.css
@@ -30,7 +30,6 @@
 }
 
 #delete-button {
-  display: none; /* TODO */
   background: none;
   border: 0;
   width: 9.6rem;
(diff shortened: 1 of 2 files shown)


tim-hellhake:
I just tried out your branch, but I can't find the delete button.
I hope I didn't misunderstand the purpose of the PR :sweat_smile:

image

tim-hellhake (approved):
I just tested it locally.
Looks awesome!

peiying2:
@mrstegeman please send me an email through Chat or regular moz email. We have a followup conversation on WTG. Thank you!

Issue #1826, opened 2019-04-25
Allow user to select which property gets highlighted


flatsiedatsie:
As mentioned in [this issue](https://github.com/mozilla-iot/gateway/issues/1809), it would be great if the user can select which property gets to take the 'center stage' spot in a device.

![co2-settings](https://user-images.githubusercontent.com/805405/56756659-c54b4f80-6781-11e9-9605-bb48eff70767.png)

One reason this came up is because of the MySensors add-on. The MySensors Arduino library allows users to create any combination of properties in a new device.

In practise, it was hard to match all MySensors possibilities to existing properties from the Mozilla schema. The result is that 
- most devices don't show data in the central position. They show a generic 'thing' instead, which feels like a waste.
- If a schema can be followed, a user might end up with an undesired highlighted item.

USER STORY
As a user I'd like to manually select which property is highlighted in the things overview. I'd like to make sure there a value is always highlighted, and I avoid the generic 'thing' display.

SOLUTIONS
There is already a dropdown in the thing edit menu. If I'm honest I had no idea what it did, because whenever I switched to another MULTI_LEVEL_SENSOR item in the selector nothing really changed. I was hoping it could be that selector.

Another option that was discussed is to create a new more universal type/property pair that can act as a 'backup' property in case a more perfect choice is not available.

To see the issue in code, have a look at [this file from the adapter](https://github.com/createcandle/Webthings-mysensors-adapter/blob/master/pkg/mysensors_device.py#L92). There you can see that the strict Mozilla schema could only be followed in a small number of cases.

flatsiedatsie:
Here are some sketches:

This sketch shows how a user can select which property should be highlighted:
![select-highlighted-property](https://user-images.githubusercontent.com/805405/56797406-274c9900-6804-11e9-86e1-59fadc09a88b.png)

This sketch is a little more far-out. It explores the idea that users can also select which icon they'd want to represent their device.
![co2-settings-and-icon-select](https://user-images.githubusercontent.com/805405/56797486-53681a00-6804-11e9-8d7c-a254a628f977.png)

kgiori:
I really like this idea. Even for completely custom things, show the list of properties of that thing and let the user pick one to be the default value to display with the icon. What about a thing with actions and events? When would it be valuable for the Edit menu to also include a listing of the things' actions and events?

flatsiedatsie:
Sorry to beat this horse, but it's still a little counter intuitive that this doesn't work.

[...]

PR #2598, opened 2020-10-05
Add support for AirQualitySensor

### LINKED ISSUE #2574: Add support for AirQualitySensor
This includes ConcentrationProperty and DensityProperty

tim-hellhake:
resolves #2574

### DIFF
--- a/static/js/components/capability/air-quality-sensor.js
+++ b/static/js/components/capability/air-quality-sensor.js
@@ -1,0 +1,19 @@
+/**
+ * AirQualityCapability
+ *
+ * A bubble showing a air quality sensor icon.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+'use strict';
+
+const LabelCapability = require('./label');
+
+class AirQualityCapability extends LabelCapability {
+}
+
+window.customElements.define('webthing-air-quality-sensor-capability',
+                             AirQualityCapability);
+module.exports = AirQualityCapability;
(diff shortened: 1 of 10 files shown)


static/js/components/capability/air-quality-sensor.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * AirQualityCapability
+ *
+ * A bubble showing a air quality sensor icon.
```
mrstegeman, line 4: nit: "an air quality"

static/js/schema-impl/capability/air-quality-sensor.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * AirQuality sensor.
+ *
+ * UI element representing a airquality sensor.
```
mrstegeman, line 4: nit: "an air quality"

static/js/schema-impl/capability/air-quality-sensor.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  findProperties() {
+    this.airqualityProperty = null;
+
+    // Look for properties by type first.
+    for (const name in this.displayedProperties) {
+      const type = this.displayedProperties[name].property['@type'];
+
+      if (type === 'ConcentrationProperty') {
+        this.airqualityProperty = name;
```
mrstegeman, line 46: nit: `airQualityProperty`

mrstegeman (changes requested):
I think my biggest issue is that we just pick the first concentration/density property to display, but that's what we do for other things as well. Just wish there was a better approach.

mrstegeman (approved):

tim-hellhake:
> I think my biggest issue is that we just pick the first concentration/density property to display, but that's what we do for other things as well. Just wish there was a better approach.

Yeah, me too.
But as long as one device is tied to one _buble_, there is not much we can do.
Maybe we need something like the rule editor where you first pick the device and then the property you want to display.

mrstegeman:
You mean like in #1826?

PR #2603, opened 2020-10-10
fix: detect container when running via podman

### LINKED ISSUE #2602: Running via podman is not recognized as "inside container"
If I want to run [the official Docker image](https://hub.docker.com/r/mozillaiot/gateway) via podman instead of Docker, the app doesn't recognize, that it's running inside a container.

Basically [`is_container() in run-app.sh`](https://github.com/WebThingsIO/gateway/blob/d84c0561385876269d703445f3944911c3870624/run-app.sh#L9-L23) returns `false` instead of `true`.

foodev:
According to https://docs.podman.io/en/latest/markdown/podman-run.1.html
podman creates the file `/run/.containerenv` to indicate to programs
they are running in a container.

resolves #2602

### DIFF
--- a/run-app.sh
+++ b/run-app.sh
@@ -7,7 +7,7 @@ args=""
 start_task="run-only"
 
 is_container() {
-  if [ -f /.dockerenv ]; then
+  if [ -f /.dockerenv ] || [ -f /run/.containerenv ]; then
     return 0
   fi
 
--- a/run-app.sh
+++ b/run-app.sh
@@ -7,7 +7,11 @@ args=""
 start_task="run-only"
 
 is_container() {
-  if [ -f /.dockerenv ] || [ -f /run/.containerenv ]; then
+  if [ -f /.dockerenv ]; then
+    return 0
+  fi
+
+  if [ -f /run/.containerenv ]; then
     return 0
   fi
 
--- a/src/platform.js
+++ b/src/platform.js
@@ -84,7 +84,7 @@ function getArchitecture() {
 }
 
 /**
- * Determine whether or not we're running inside a contianer (e.g. Docker).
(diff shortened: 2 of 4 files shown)


run-app.sh
```diff
@@ -7,7 +7,7 @@ args=""
 start_task="run-only"
 
 is_container() {
-  if [ -f /.dockerenv ]; then
+  if [ -f /.dockerenv ] || [ -f /run/.containerenv ]; then
```
mrstegeman, line 10: Please break this into a separate block, just to keep things consistent.

mrstegeman (changes requested):
There’s also a check in the source that needs to be updated: https://github.com/WebThingsIO/gateway/blob/4606c6d73854afae90cc92a8b07ae1c18670763e/src/platform.js#L89

foodev:
https://github.com/WebThingsIO/gateway/blob/4606c6d73854afae90cc92a8b07ae1c18670763e/src/platform.js#L89

Then I would change the name to `isContainer`, too. This way it's consistent with the Bash script.

Furthermore I think the user-agent should contain `(container)` instead of `(docker)`.

https://github.com/WebThingsIO/gateway/blob/d84c0561385876269d703445f3944911c3870624/src/utils.js#L70

Is that OK?

mrstegeman:
That sounds good to me.

src/platform.js
```diff
@@ -84,10 +84,11 @@ function getArchitecture() {
 }
 
 /**
- * Determine whether or not we're running inside Docker.
+ * Determine whether or not we're running inside a contianer (e.g. Docker).
```
mrstegeman, line 87: nit: `container`

mrstegeman (changes requested):

mrstegeman (approved):

PR #2610, opened 2020-10-15
added jwt unit test

Kevchan85:
Created a simple unit test. I think the next part is to use mock data so it always returns the same private and public keys. Not really sure what libraries I should use for this. Did you have a preference? I probably have to lint this, but I thought I would get some feedback on mocks first. I'm trying to understand all the small pieces that make up the integration tests before I attempt to take those on. Thanks!

### DIFF
--- a/src/test/jsonwebtoken-test.js
+++ b/src/test/jsonwebtoken-test.js
@@ -1,0 +1,10 @@
+const JSONWebToken = require('../models/jsonwebtoken');
+
+describe('Json Web Token', () => {
+  const id = 12345
+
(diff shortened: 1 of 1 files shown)


mrstegeman:
One quick note is that this should be moved into `src/test/integration/`.

As for libraries, feel free to use whatever you want. I don't have a strong preference.

mrstegeman:
@Kevchan85 Still planning to work on this?

Kevchan85:
Yes sorry, I've been busy. I will block out some time this weekend!

Issue #2614, opened 2020-10-23
Subdomain registration form updates


benfrancis:
During first time setup users can tick a box which says "Please keep me updated about new features and contribution opportunities."

Some changes are needed to this form:
- [X] The base domain needs changing from .mozilla-iot.org to .webthings.io (#2620)
- [X] The mailing list opt-in checkbox should be unchecked by default (#2624)
- [X] The text next to the checkbox needs changing to "Keep me updated with news about WebThings." (#2624)
- [X] There needs to be a separate checkbox to agree to the WebThings privacy policy (also unchecked by default) with the text "Agree to the WebThings privacy policy" (#2627)
- [X] The Privacy policy link should now link to https://webthings.io/privacy (#2625)
- [X] Mailing list registration needs to subscribe to the mailing list directly rather than set a flag on the registration server (#2631)

PR #2620, opened 2020-11-06
Switch tunneling domain from mozilla-iot.org to webthings.io.

mrstegeman:


benfrancis:
@mrstegeman I was going to review this and make reference to the points in #2614, but they can be addressed separately if that's easier.

mrstegeman:
@benfrancis I'm going to do this piece by piece. I updated the top-level comment in #2614.

benfrancis (commented):
LGTM

PR #2627, opened 2020-11-09
Add a checkbox to agree to privacy policy.

mrstegeman:
References #2614

### DIFF
--- a/static/css/subdomain-form.css
+++ b/static/css/subdomain-form.css
@@ -114,7 +114,7 @@ button:active:enabled {
 
 #agree-group,
 #opt-in-group {
-  padding-top: 2rem;
+  padding-top: 0;
   display: grid;
   grid-template-columns: 8rem auto;
   width: 100%;
@@ -167,9 +167,9 @@ button:active:enabled {
   padding-top: 2rem;
 }
 
-#agree-info > a:link,
-#agree-info > a:visited,
-#agree-info > a:hover,
-#agree-info > a:active {
+#agree-info > span > a:link,
+#agree-info > span > a:visited,
+#agree-info > span > a:hover,
+#agree-info > span > a:active {
   color: white;
 }
--- a/static/css/subdomain-form.css
+++ b/static/css/subdomain-form.css
@@ -112,20 +112,24 @@ button:active:enabled {
   background-color: #658196;
 }
 
+#agree-group,
 #opt-in-group {
   padding-top: 2rem;
-  display: table;
+  display: grid;
+  grid-template-columns: 8rem auto;
   width: 100%;
 }
 
+#agree,
 #opt-in {
   position: absolute;
   left: -1000em;
 }
 
+#agree + label::before,
 #opt-in + label::before {
+  display: inline-block;
   content: '';
-  display: table-cell;
   vertical-align: middle;
   width: 3rem;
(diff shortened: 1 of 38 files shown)


benfrancis (commented):
Let's hold off on this one for a moment while I figure out whether we also need a Terms of Service. Ideally we could just add "and Terms of Service"  to the string, but that may not be GDPR compliant.

static/fluent/en-US/main.ftl
```diff
@@ -534,7 +534,7 @@ tunnel-setup-choose-address = Choose a secure web address for your gateway:
 tunnel-setup-input-subdomain =
     .placeholder = subdomain
 tunnel-setup-email-opt-in = Keep me updated with news about WebThings.
-tunnel-setup-privacy-policy = Privacy Policy
+tunnel-setup-agree-privacy-policy = Agree to the <a href="https://webthings.io/privacy">WebThings privacy policy</a>
```
benfrancis, line 537: `Agree to the WebThings <a href="https://webthings.io/privacy">Privacy Policy</a> and <a href="https://webthings.io/legal">Terms of Service</a>.`

benfrancis (changes requested):
OK, my conclusion is that it's OK to agree to the Privacy Policy and Terms of Service with the same checkbox, as long as the newsletter signup is separate.

static/fluent/en-US/main.ftl
mrstegeman: Updated.

mrstegeman:
@benfrancis is this good to go now?

static/css/subdomain-form.css
```diff
@@ -112,20 +112,24 @@ button:active:enabled {
   background-color: #658196;
 }
 
+#agree-group,
 #opt-in-group {
   padding-top: 2rem;
```
benfrancis, line 117: I suggest `padding-top: 0;`

static/fluent/en-US/main.ftl
```diff
@@ -534,7 +534,7 @@ tunnel-setup-choose-address = Choose a secure web address for your gateway:
 tunnel-setup-input-subdomain =
     .placeholder = subdomain
 tunnel-setup-email-opt-in = Keep me updated with news about WebThings.
-tunnel-setup-privacy-policy = Privacy Policy
+tunnel-setup-agree-privacy-policy = Agree to the WebThings <a href="https://webthings.io/privacy">Privacy Policy</a> and <a href="https://webthings.io/legal">Terms of Service</a>.
```
benfrancis, line 537: Unfortunately the hyperlinks are not rendering for me. I see a fluent error in the logs...
`An element of forbidden type "a" was found in the translation. Only safe text-level elements and elements with data-l10n-name are allowed.`

benfrancis (changes requested):
Sorry, I missed the new review request. A couple of comments inline.

static/css/subdomain-form.css
mrstegeman: Done.

static/fluent/en-US/main.ftl
mrstegeman: Fixed.

benfrancis (approved):
Great, thanks.

PR #2635, opened 2020-11-13
Change base directory for gateway.

mrstegeman:


benfrancis (approved):
All this upgrade logic is going to require some thorough testing

PR #2640, opened 2020-11-17
Update gateway-addon.

mrstegeman:


mrstegeman:
@tim-hellhake looks good

Issue #2253, opened 2019-11-05
Offer beta-test images for add-on developers?


flatsiedatsie:
I read that version 0.10.0 is 'feature complete' and is being tested internally. Perhaps at these points in development a test image could be shared online, so that add-on developers can work on their add-ons more easily, and make them ready for the new version too?

mrstegeman:
One of the issues with making them public is that we'd then have to generate and push out OTA updates for each beta release we do.

flatsiedatsie:
Couldn't part of the beta img be that this is not a feature?

I was just thinking that
- Following all the steps to do a manual build is a lot of work.
- You probably also have test images internally.
- Some test images were shared in the past, and they were very handy.

I (personally) wouldn't expect any support, and understand that the beta image should not be expected to become a normal system through updates. It's just that I would like to avoid manual installation for every release.

Alternatively, would it be possible to communicate when the code for a new version is theoretically 'feature complete'? Then this manual generation would only have to be done once per minor version.

mrstegeman:
How would you feel about a nightly and/or beta Docker image as a test target?

flatsiedatsie:
I don't really know, I've never used docker. But if you think it would be a solution (not just for me, for everyone) then I'm all for trying it.

benfrancis:
FWIW you can find a 0.10-pre1 image [here](https://s3-us-west-1.amazonaws.com/mozillagatewayimages/images/gateway-0.10.0-pre1.img.zip), just please don't use it in production and remember to re-flash with the final 0.10 release when it comes out.

I expect there will be a pre2 image today or tomorrow which includes some recent bugfixes.

flatsiedatsie:
Thanks, that's very useful!

benfrancis:
[0.10-pre2](https://s3-us-west-1.amazonaws.com/mozillagatewayimages/images/gateway-0.10.0-pre2.img.zip)

flatsiedatsie:
Thank you! This makes development so much easier!

mrstegeman:
After upgrading to 1.0 (coming soon), you'll be able to set the new `updates.allowPrerelease` config flag.

PR #2641, opened 2020-11-19
Allow users to install prerelease updates.

mrstegeman:
References #2253

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -52,7 +52,10 @@ module.exports = {
     certemail: 'noreply@webthings.io',
   },
   bcryptRounds: 2,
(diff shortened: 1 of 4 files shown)


benfrancis (approved):
Clever. So does this require users to edit the config file and rebooting to enable pre-releases? Would you mind documenting this somewhere (e.g. on the wiki)?

mrstegeman:
@benfrancis You can set it in `~/.webthings/config/local.json` and the daily update check service will just pick it up there. If you want the gateway to see it via the UI, you'd have to restart the gateway after setting the flag.

Issue #2612, opened 2020-10-23
Transition UI


benfrancis:
As a user I want to transfer my gateway to use webthings.io services so that I can continue to receive software updates and access my gateway remotely.

The Mozilla IoT remote access service and automatic software updates are being [discontinued](https://discourse.mozilla.org/t/an-important-update-on-mozilla-webthings/67764/11). This UI in the gateway notifies users of this change and guides them through the transfer to replacement community-run services at webthings.io.

Draft UX specification [here](https://docs.google.com/presentation/d/1X5pfyQX-4oeRcNxxPf5wtpt2MgYi5VBBENIZkJX6Axc/edit?usp=sharing).

SVG assets incoming...

benfrancis:
Icons [icons.zip](https://github.com/WebThingsIO/gateway/files/5431155/icons.zip)

benfrancis:
![transition_banner5](https://user-images.githubusercontent.com/552417/99839386-8b5f4c80-2b62-11eb-8ade-569b490bd377.png)

![transition_prompt5](https://user-images.githubusercontent.com/552417/99839402-90240080-2b62-11eb-9a7f-a461dd78e29c.png)

PR #2644, opened 2020-11-20
Implement base transition dialog.

mrstegeman:
References #2612


benfrancis (approved):
The strings look spot on to me if you want to land those for localisation.

You might want to add #transition-dialog to the .hidden CSS selector before landing though, as it's currently visible.

mrstegeman:
Good call. Done.

PR #2647, opened 2020-11-22
Support M-JPEG video streams.

mrstegeman:
References WebThingsIO/onvif-adapter#4

### DIFF
--- a/static/js/schema-impl/property/video.js
+++ b/static/js/schema-impl/property/video.js
@@ -24,6 +24,7 @@ class VideoDetail {
 
     this.dashHref = null;
     this.hlsHref = null;
+    this.mjpegHref = null;
     this.player = null;
 
     for (const link of property.links) {
@@ -34,6 +35,10 @@ class VideoDetail {
         } else if (this.hlsHref === null &&
                    link.mediaType === 'application/vnd.apple.mpegurl') {
           this.hlsHref = link.href;
+        } else if (this.mjpegHref === null &&
+                   (link.mediaType === 'video/x-motion-jpeg' ||
+                    link.mediaType === 'video/x-jpeg')) {
+          this.mjpegHref = link.href;
         }
       }
     }
@@ -69,7 +74,16 @@ class VideoDetail {
     const element = document.createElement('div');
     element.classList.add('media-modal-backdrop');
 
-    if (!shaka.Player.isBrowserSupported()) {
+    if (this.mjpegHref !== null) {
+      element.innerHTML = `
+        <div class="media-modal">
+          <div class="media-modal-frame">
+            <div class="media-modal-close"></div>
+            <img class="media-modal-video"></img>
+          </div>
+        </div>
+        `;
+    } else if (!shaka.Player.isBrowserSupported()) {
       element.innerHTML = `
         <div class="media-modal">
           <div class="media-modal-frame">
@@ -93,7 +107,7 @@ class VideoDetail {
     document.body.appendChild(element);
     document.querySelector('#things').style.display = 'none';
 
-    if (!shaka.Player.isBrowserSupported()) {
+    if (this.mjpegHref || !shaka.Player.isBrowserSupported()) {
       this.positionButtons();
     }
 
@@ -113,7 +127,15 @@ class VideoDetail {
 
     window.addEventListener('resize', this.positionButtons);
 
-    if (shaka.Player.isBrowserSupported() && (this.dashHref || this.hlsHref)) {
+    if (this.mjpegHref) {
+      element.querySelector('.media-modal-video').addEventListener(
+        'load',
+        this.positionButtons
+      );
+      element.querySelector('.media-modal-video').src =
+        `${this.mjpegHref}?jwt=${API.jwt}`;
+    } else if (shaka.Player.isBrowserSupported() &&
+               (this.dashHref || this.hlsHref)) {
       element.querySelector('.media-modal-video').addEventListener(
         'loadeddata',
         this.positionButtons


mrstegeman:
@atirage as requested 😄

benfrancis (approved):
I'm not able to test this, but the code looks OK.

atirage:
I'll test it with a Raspberry Pi camera in the near future.

mrstegeman:
I tested by hijacking the VideoCamera device in the Virtual Things add-on and using this: https://github.com/blueimp/mjpeg-server

atirage:
@mrstegeman could you please share with me your Virtual Thing code you used to test mjpeg streams. I'm getting some Tornado errors in my PiCamera version.
Thanks in advance!

mrstegeman:
It was just something I hacked together with virtual-things-adapter, using `mjpeg-server`. I just ran that in the background, on loop, rather than using `ffmpeg`, and I swapped out the `href` and `mediaType`. I never built an actual web thing to do it.

What kind of issues are you seeing?

atirage:
I'm getting this error, (jwt replaced with ...):

> ERROR:tornado.application:Uncaught exception GET /media/screenshots/stream.vid?jwt=...
(192.168.0.178)
HTTPServerRequest(protocol='http', host='gateway.local:8900', method='GET', uri='/media/screenshots/stream.vid?jwt=...
version='HTTP/1.1', remote_ip='192.168.0.178')
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/tornado/web.py", line 1703, in _execute
    result = await result
  File "/usr/local/lib/python3.7/dist-packages/tornado/web.py", line 2653, in get
    await self.flush()
  File "/usr/local/lib/python3.7/dist-packages/tornado/web.py", line 1101, in flush
    return self.request.connection.write(chunk)
  File "/usr/local/lib/python3.7/dist-packages/tornado/http1connection.py", line 497, in write
    self._pending_write = self.stream.write(self._format_chunk(chunk))
  File "/usr/local/lib/python3.7/dist-packages/tornado/http1connection.py", line 474, in _format_chunk
    "Tried to write more data than Content-Length"
tornado.httputil.HTTPOutputError: Tried to write more data than Content-Length
ERROR:tornado.general:Cannot send error response after headers written

I'm recording a 10 fps mjpeg stream into that file from the uri, basically each new frame is written to the file.
I'm suspecting some kind of access synch issue.

mrstegeman:
You need to make sure tornado is handling a stream, rather than a static resource. I found an example (it's a couple years old) that should provide a good starting point: https://github.com/wildfios/Tornado-mjpeg-streamer-python

atirage:
After reworking the code you referenced to use coroutines, I'm getting this:

>     await self.flush()
tornado.iostream.StreamClosedError: Stream is closed
ERROR:tornado.general:Cannot send error response after headers written 

I added some debug code and I could see that this is not the first execution of the flush method, it comes after the 2nd or 3rd time. Also, the real_error part of the exception is None. :(

The same happens if I access the stream in the browser.

mrstegeman:
Can you share your code?

atirage:
@mrstegeman, I finally managed to get it working! :) It was a classic case of RTFM, fortunately I've found this link which explains how this multipart response is supposed to be structured:

> https://stackoverflow.com/questions/47729941/mjpeg-over-http-specification

I still need to do some refactoring and testing, but there'll be a PiCamera WebThing soon. :)

mrstegeman:
Great! I’d love to see a demo when you get it working.

atirage:
Here's the latest version:

> https://github.com/atirage/picamera-webthing/blob/master/picamera-webthing.py

I get a nice continuous, low latency 720p stream @30fps with local access.
@mrstegeman, in what form would you like see a demo?

mrstegeman:
This is fantastic, thanks for sharing! I don't need a demo, but I do have a few questions.

1. Is this working just fine with the gateway and Web Thing add-on, with no changes to either?
2. Does it work if you're accessing your gateway remotely, e.g. through the `*.webthings.io` tunnel?
3. What kind of latency are you seeing? We were always at 30-60 seconds with the live video streams.

The biggest downside of M-JPEG is no audio.

atirage:
1. Yes, no changes needed, it works "out of the box".
2. Unfortunately, the stream is not working through the *.webthings.io tunnel, I didn't check further yet. The snapshot is working.
3. Accessing locally, there's very low latency, not empirically determined, but I would say below 1s.

mrstegeman:
Ah, the tunneling part is what I was worried about. Let me know if you can figure out what’s wrong!

atirage:
I've added a debug print when my ´StreamHandler` is invoked and I noticed that it happened only the first time I tried the tunnel access. Subsequent accesses through the tunnel don't seem to reach my handler at all and the web console shows the following response:

> Status 200
OK
Version HTTP/1.1
Transferred 44.31 MB (44.31 MB size)
Referrer Policyno-referrer-when-downgrade
HTTP/1.1 200 OK
X-Powered-By: Express
Strict-Transport-Security: max-age=31536000; includeSubDomains
Content-Security-Policy: frame-ancestors 'none'
Vary: Accept
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Authorization
Access-Control-Allow-Methods: GET,HEAD,PUT,PATCH,POST,DELETE
Surrogate-Control: no-store
Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate
Pragma: no-cache
Expires: 0
server: TornadoServer/6.0.4
content-type: multipart/x-mixed-replace;boundary=--jpgboundary
date: Wed, 30 Dec 2020 16:28:10 GMT
connection: close
transfer-encoding: chunked

In the UI this results in displaying a still image.
I'm not sure where to go next from here...

On a side note, it seems that the Gateway does not close the stream if the stream UI page is left/closed as a browser would.

mrstegeman:
This is curious. Given that things work locally, it seems like the local proxy is doing what it needs to do. If you watch the network requests in your browser's developer tools, do you see anything weird happening?

atirage:
Well, like I wrote earlier if I access now the stream through the tunnel service I see the above response header in my browser, as a payload I have a bunch of data separated by the `--jpegboundary`, so I guess this is why a still image is shown in the UI, but the image is not a current one.
More over, this response is not coming from my `WebThing`, since my `StreamHandler` is not invoked.

PR #2649, opened 2020-11-23
Transition dialog CSS

benfrancis:
References #2612


mrstegeman:
Please rebase on master.

benfrancis:
Notes:
1. I'm assuming the existing scrim element can be used when the dialog is displayed
2. I've added a form element and changed the submit button to an input type="submit" in order to take advantage of HTML form validation (highlighting required fields), but this requires an L10n change so please check this is what you want. You might want to set "required" on the subdomain text box if the checkbox is ticked, and do `event.preventDefault()` on the submit event of the form.
3. I think you may have missed off the success message (sorry I missed that), and I wasn't sure how you wanted to implement that

Let me know if I missed anything!

static/index.html
```diff
@@ -632,32 +632,34 @@ <h3 class="logs-header" data-l10n-id="logs"></h3>
...
+          <label id="transition-dialog-newsletter-label"
+            data-l10n-id="transition-dialog-newsletter-label"
+            for="transition-dialog-newsletter-checkbox">
+          </label>
+          <input id="transition-dialog-agree-tos-checkbox" name="transition-dialog-agree-tos"
+            class="transition-checkbox" type="checkbox" required>
+          <label id="transition-dialog-agree-tos-label"
+            data-l10n-id="transition-dialog-agree-tos-label"
+            for="transition-dialog-agree-tos-checkbox">
+            <a data-l10n-name="transition-dialog-privacy-policy-link" href="https://webthings.io/privacy" target="_blank"></a>
```
mrstegeman, line 655: These should also have `rel="noopener"`.

mrstegeman (changes requested):

static/fluent/en-US/main.ftl
```diff
@@ -631,7 +631,8 @@ transition-dialog-newsletter-label = Keep me updated with news about WebThings
 transition-dialog-agree-tos-label = Agree to the WebThings <a data-l10n-name="transition-dialog-privacy-policy-link">Privacy Policy</a> and <a data-l10n-name="transition-dialog-tos-link">Terms of Service</a>.
 transition-dialog-email =
     .placeholder = Email address
-transition-dialog-register = Register
+transition-dialog-register =
```
mrstegeman, line 634: You'll need to make the same change in all of the other translations, since quite a few have already been translated.

mrstegeman:
> 3. I think you may have missed off the success message (sorry I missed that), and I wasn't sure how you wanted to implement that

Ah, shoot. Can you go ahead and add the element(s) you need for that? The fluent string can be something like:
```
transition-dialog-success = Navigate to <<domain>> to continue.
```

I'll then fill in the `<<domain>>` via JS. We do something similar for some of the OAuth strings.

static/fluent/en-US/main.ftl
benfrancis: Ah OK, I wasn't sure whether we were supposed to edit them all directly.

mrstegeman (approved):

PR #2655, opened 2020-11-25
Implement logic for transition dialog.

### LINKED ISSUE #2612: Transition UI

mrstegeman:
Closes #2612

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -158,7 +158,20 @@ function startHttpGateway() {
 
 function stopHttpGateway() {
   servers.http.removeListener('request', httpApp);
+
+  if (httpsApp) {
+    servers.http.removeListener('request', httpsApp);
+  }
+
   servers.http.close();
+}
+
+function stopHttpsGateway() {
+  if (servers.https && httpsApp) {
+    servers.https.removeListener('request', httpsApp);
+    servers.https.close();
+    servers.https = null;
+  }
 }
 
 function startWiFiSetup() {
@@ -401,6 +414,7 @@ process.on('SIGINT', () => {
 // function to stop running server and start https
 TunnelService.switchToHttps = () => {
   stopHttpGateway();
+  stopHttpsGateway();
   startHttpsGateway().then((server) => {
     TunnelService.setServerHandle(server);
   });
--- a/src/certificate-manager.js
+++ b/src/certificate-manager.js
@@ -57,9 +57,11 @@ function writeCertificates(certificate, privateKey, chain) {
  * @param {string} fulldomain - The full domain being registered
  * @param {boolean} optout - Whether or not the user opted out of emails
  * @param {function} callback - Callback function
+ * @param {function?} newsletterCallback - Callback function for newsletter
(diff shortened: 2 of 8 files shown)


benfrancis (approved):
Nice work on getting this working. I had a bunch of comments about the code but realised they were all about coding style and API design so decided they're not really important given this code is only temporary.

I've tested this quite thoroughly and can't find any major bugs.

One slight UI bug: It's possible to get the UI into a state where the #update-message-area and #action-required-area divs both appear and overlap each other when the page first loads after a 1.0 upgrade, which looks broken (I'm working on steps to reproduce because it didn't happen the second time I tried it and I'm not sure what I did differently).

Ideally we'd have them stack on top of each other along the y axis but I can't spot a quick fix for that. Worst case we could just give the update dialog a darker (and opaque) background colour (#48779a) with darker buttons (#315169) so the boundary between the two can be seen.

I did spot an exception in the console at one point about a 500 error, but I haven't been able to reproduce it again. I'll let you know if I manage it.

I have a couple of comments about the transition emails that are sent out, but I'll share those separately as they don't relate to this PR.

benfrancis:
> One slight UI bug: It's possible to get the UI into a state where the #update-message-area and #action-required-area divs both appear and overlap each other

FWIW, I haven't been able to reproduce this again.

mrstegeman:
> Nice work on getting this working. I had a bunch of comments about the code but realised they were all about coding style and API design so decided they're not really important given this code is only temporary.

Yeah, it was definitely quick and dirty.
 
> One slight UI bug: It's possible to get the UI into a state where the #update-message-area and #action-required-area divs both appear and overlap each other when the page first loads after a 1.0 upgrade, which looks broken (I'm working on steps to reproduce because it didn't happen the second time I tried it and I'm not sure what I did differently).

That's easy to reproduce. Just run `npm start` again and soft-refresh the UI. I thought about adding some logic to not show the transition banner when the reload banner is active. What do you think about that?

benfrancis:
> That's easy to reproduce. Just run npm start again and soft-refresh the UI. I thought about adding some logic to not show the transition banner when the reload banner is active. What do you think about that?

Yep that sounds ideal.

mrstegeman:
> > That's easy to reproduce. Just run npm start again and soft-refresh the UI. I thought about adding some logic to not show the transition banner when the reload banner is active. What do you think about that?
> 
> Yep that sounds ideal.

Done.

PR #2659, opened 2020-11-30
Add <img> tags for all transition images.

mrstegeman:
This prevents issues with displaying the proper image during the
transition process.


benfrancis (approved):
Haven't had chance to test this but the code looks good. Will be good to test it on the actual image to see how it behaves on the subdomains in practice.

PR #2700, opened 2021-01-08
Translate several files to TypeScript.

mrstegeman:


tim-hellhake (approved):

benfrancis:
What is the intention here? Is it to eventually move the whole code base to TypeScript?

I don't necessarily think that's a bad idea, but we discussed doing that a couple of years back and decided against it at the time. If we're going to re-visit that decision I suggest it might be a good idea to have an open discussion about the tradeoffs involved (e.g. on Discourse), so that it's an explicit architectural decision rather than just a gradual process of erosion over time.

Would you be OK with that?

mrstegeman:
Yes, I would very much like to move the whole code base to TypeScript, although it's going to take some time. However, to be clear, it's not changing the architecture at all, and in most cases the transpiled code is identical to the original. The only major change I've made was to stop running webpack on the server source, which we had no reason to do anyway. I don't remember any conversations about moving things to TypeScript, only Rust, so maybe that was before my time.

Benefits:
* Code is more understandable.
* Static typing helps find lots of issues (I've already found and fixed several). This leads to fewer bugs.
* The webthing-node and gateway-addon-node libraries have already been moved over to TypeScript. Additionally, part of the gateway source is already TypeScript. Moving everything will lead to a less confusing code base.
* Moving things over has already brought in major contributions to the other libraries mentioned above, so I'm hopeful it will encourage new contributors to the gateway as well. @tim-hellhake is already helping me out, which is much appreciated.

Drawbacks:
* TypeScript is new to some people. However, I've been able to pick it up without ever reading any TypeScript docs, so it's a pretty small learning curve.

With all that said, I've posted on Matrix and [Discourse](https://discourse.mozilla.org/t/transitioning-gateway-source-to-typescript/73722) as well, in case anyone wants to weigh in.

PR #2702, opened 2021-01-10
Drop router and OpenWrt code.

mrstegeman:


static/images/wifi-secure.svg
```diff
@@ -1,49 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
benfrancis, line 1: Is this image not still used in Wi-Fi setup? https://github.com/WebThingsIO/gateway/blob/master/src/wifi-setup.js#L136

static/css/lib/highlight-9.12.0.min.css
```diff
@@ -1 +0,0 @@
-.hljs{display:block;overflow-x:auto;padding:0.5em;background:#F0F0F0}.hljs,.hljs-subst{color:#444}.hljs-comment{color:#888888}.hljs-keyword,.hljs-attribute,.hljs-selector-tag,.hljs-meta-keyword,.hljs-doctag,.hljs-name{font-weight:bold}.hljs-type,.hljs-string,.hljs-number,.hljs-selector-id,.hljs-selector-class,.hljs-quote,.hljs-template-tag,.hljs-deletion{color:#880000}.hljs-title,.hljs-section{color:#880000;font-weight:bold}.hljs-regexp,.hljs-symbol,.hljs-variable,.hljs-template-variable,.hljs-link,.hljs-selector-attr,.hljs-selector-pseudo{color:#BC6060}.hljs-literal{color:#78A960}.hljs-built_in,.hljs-bullet,.hljs-code,.hljs-addition{color:#397300}.hljs-meta{color:#1f7199}.hljs-meta-string{color:#4d99bf}.hljs-emphasis{font-style:italic}.hljs-strong{font-weight:bold}
```
benfrancis, line 1: Is this not still used by the local token developer settings UI? https://github.com/WebThingsIO/gateway/blob/master/src/views/local-token-service.handlebars#L14

static/things.json
```diff
@@ -1,3 +0,0 @@
-[
-
-]
```
benfrancis, line 3: LOL, what's that still doing there? Oh, I left it there 😂

benfrancis (changes requested):

static/images/wifi-secure.svg
mrstegeman: Good catch! I was looking in the wrong place.

static/things.json
mrstegeman: Haha I don't know how I've completely missed that file for over 3 years.

static/css/lib/highlight-9.12.0.min.css
mrstegeman: Good catch! We were using a very outdated one, too, so I've updated it.

benfrancis (approved):

PR #2705, opened 2021-01-12
Migrate plugin dir to typescript

tim-hellhake:


tim-hellhake:
Split-up into multiple PRs

PR #2708, opened 2021-01-13
Migrate controllers to TypeScript.

mrstegeman:


tim-hellhake (approved):

benfrancis:
I can't currently provide a meaningful review of TypeScript code, so you and @tim-hellhake are on your own with that for now I'm afraid 😟

PR #2712, opened 2021-01-14
Drop debug controller.

mrstegeman:


tim-hellhake (approved):

davehylands:
I know that I found this to be really useful feature for the zigbee and zwave adapters.

mrstegeman:
@davehylands A couple points here that led to this decision:
1. It was leading to some circular logic in the gateway during the TypeScript migration.
2. I've been trying to reduce the maintenance burden, dropping (mostly) unused features like the OpenWrt code, the virtual assistant, etc.
3. I'm fairly certain you're the only person that has ever used this 😆 

It should be pretty straightforward to build a debug API into those two adapters if we end up needing it. Seems more logical to do that than to have a feature in the gateway that's only used for two adapters.

PR #2717, opened 2021-01-15
Migrate controllers

tim-hellhake:


src/router.ts
```diff
@@ -10,28 +10,54 @@
...
+import RootController from './controllers/root_controller';
+import SettingsController from './controllers/settings_controller';
+import ThingsController from './controllers/things_controller';
+import UpdatesController from './controllers/updates_controller';
+import UploadsController from './controllers/uploads_controller';
+import UsersController from './controllers/users_controller';
+import RuleEngine from './rules-engine/index';
+import ProxyController, {ProxyStore} from './controllers/proxy_controller';
+import ExtensionsController from './controllers/extensions_controller';
 const nocache = require('nocache')();
```
mrstegeman, line 41: Missed a couple `require()` calls.

src/router.ts
```diff
@@ -105,74 +130,52 @@ const Router = {
...
 
     // Handle proxied resources
-    app.use(APP_PREFIX + Constants.PROXY_PATH, nocache, auth,
-            this.proxyController);
+    app.use(APP_PREFIX + Constants.PROXY_PATH, nocache, auth, proxyController);
 
     // Let OAuth handle its own rendering
-    app.use(APP_PREFIX + Constants.OAUTH_PATH, nocache,
-            require('./controllers/oauth_controller'));
+    app.use(APP_PREFIX + Constants.OAUTH_PATH, nocache, OAuthController());
```
mrstegeman, line 136: You could probably just create a single instance of this and reuse it on line 168.

src/router.ts
```diff
@@ -105,74 +130,52 @@ const Router = {
...
+  addProxyServer(thingId: string, server: string) {
+    this.proxyStore?.addProxyServer(thingId, server);
+  }
+
+  removeProxyServer(thingId: string) {
+    this.proxyStore?.removeProxyServer(thingId);
+  }
+}
+
+export = new Router();
```
mrstegeman, line 181: I've been trying to use `export default`, rather than `export =`. You'd just need to change the `require()` in `thing.js` and `things.js`.

mrstegeman (changes requested):

src/router.ts
tim-hellhake: Done

src/router.ts
tim-hellhake: I had trouble with this one before.
But it seems to work now.

src/router.ts
mrstegeman: Still need to update the Thing and Things models.

mrstegeman (dismissed):

src/router.ts
tim-hellhake: `app.js` was missing

tim-hellhake:
resolved by #2708

PR #2720, opened 2021-01-15
Migrate effects

tim-hellhake:


src/rules-engine/Property.ts
```diff
@@ -6,22 +6,43 @@
...
 'use strict';
 
-const assert = require('assert');
-const AddonManager = require('../addon-manager');
-const Constants = require('../constants');
-const Things = require('../models/things');
-const {EventEmitter} = require('events');
-const Events = require('./Events');
+import assert from 'assert';
+import AddonManager from '../addon-manager';
```
mrstegeman, line 10: If you switch AddonManager, Thing, and Things back to `require()` for now, the tests will work better.

mrstegeman (approved):

PR #2725, opened 2021-01-17
Migrate notifier proxy

tim-hellhake:


src/plugin/notifier-proxy.ts
```diff
@@ -11,25 +11,25 @@
 
 'use strict';
 
-const Deferred = require('../deferred').default;
+import Deferred from '../deferred';
 const {MessageType} = require('gateway-addon').Constants;
```
mrstegeman, line 15: `MessageType` can be switched over, as you did with the APIHandlerProxy.

src/plugin/outlet-proxy.ts
```diff
@@ -9,33 +9,37 @@
 
 'use strict';
 
+import {Level} from 'gateway-addon/lib/schema';
+
 const {MessageType} = require('gateway-addon').Constants;
```
mrstegeman, line 14: Ditto for the `MessageType`.

mrstegeman (changes requested):

mrstegeman (approved):

PR #2733, opened 2021-01-18
Migrate a great many things to TypeScript

tim-hellhake:


tim-hellhake:
@mrstegeman only together can we turn him to the dark side...

src/models/things.ts
```diff
@@ -45,15 +47,15 @@ const Things = {
    * added is when the addon discovers a thing, not when the model is
    * instantiated
    */
-  emitter: new EventEmitter(),
+  private emitter = new EventEmitter();
```
mrstegeman, line 50: Side note: I tried taking this out and just making the Things class extend EventEmitter and it failed miserably. Not sure why.

mrstegeman (approved):

PR #2738, opened 2021-01-20
Try switching browser tests to Chrome.

mrstegeman:
I'm wondering if the random test failures may be caused by the
WebSocket issues we see from time to time in Firefox.

References : https://bugzilla.mozilla.org/show_bug.cgi?id=1360753

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt -qq update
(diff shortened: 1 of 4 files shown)


benfrancis:
Was your conclusion that this wasn't the cause?

mrstegeman:
I still think it may be. However, the Chrome tests were throwing some other error that I didn't understand, and I'm presently unable to get the browser to tests to run at all locally.

PR #2748, opened 2021-01-23
fix single quote in wifi network name

stutrek:
This handlebars helper that escapes the wifi network name operates on single quotes, but is only used inside double quotes. Additionally, it had an extra backslash, and because the regex was missing a `g`, it would not work as intended if there was a second single quote.

I was not able to test this code because I don't know the steps involved in making a pi image from source, but I did make sure the regex works.


stutrek:
The linting error appears to be incorrect.

image

stutrek:
linting issues are unrelated.

mrstegeman (approved):

PR #2754, opened 2021-01-25
Move unload logic to plugin

tim-hellhake:


src/plugin/plugin.ts
```diff
@@ -817,4 +818,52 @@ export default class Plugin {
...
   }
+
+  unloadComponents(): Promise<void> {
+    const adapters = Array.from(this.adapters.values());
+    const notifiers = Array.from(this.notifiers.values());
+    const apiHandlers = Array.from(this.apiHandlers.values());
+
+    const unloadPromises = [];
+
+    for (const [, adapter] of this.adapters.entries()) {
```
mrstegeman, line 829: `for (const adapter of this.adapters.values())` would be cleaner. Ditto for the notifier block below.

mrstegeman (changes requested):

src/plugin/plugin.ts
tim-hellhake: Done

mrstegeman (approved):

PR #2755, opened 2021-01-25
Prohibit usage of `any` type.

mrstegeman:


src/test/browser/page-object/elements.ts
```diff
@@ -11,6 +11,7 @@ class Elements {
 
   protected rootElement?: webdriverio.Element;
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [name: string]: any;
```
tim-hellhake, line 15: Did you consider using `unknown` here?
That won't trigger the `any` rule.

tim-hellhake (changes requested):

tim-hellhake (approved):

src/test/browser/page-object/elements.ts
mrstegeman: Yeah... That ends up giving me 82 new errors. I'm not worried about it since it's test code.

PR #2763, opened 2021-01-29
Fix mDNS hostname/status logic.

mrstegeman:
We've apparently been in a weird, semi-functional hybrid state
since dropping the built-in mDNS server. This makes the code rely
solely on the platform functionality, and drops the useless config
settings.


src/db.ts
```diff
@@ -67,10 +67,6 @@ class Database {
     this.db.serialize(() => {
       this.createTables();
       this.migrate();
-      // If database newly created, populate with default data
-      if (!exists) {
-        this.populate();
-      }
```
benfrancis, line 73: Was this the only default data in the database?

config/default.js
```diff
@@ -46,18 +46,6 @@ module.exports = {
...
-    defaults: {
-      mdns: {
-        // Whether or not to enable mDNS advertisements
-        enabled: true,
-
-        // Domain to advertise via mDNS
-        domain: os.hostname().split('.')[0],
-      },
-    },
-  },
```
benfrancis, line 59: How should this be customised now? Can we document this somewhere?

benfrancis (approved):
Eek. A couple of questions, but this looks good to me.

src/db.ts
mrstegeman: Yes.

config/default.js
mrstegeman: This is kind of the whole point. These defaults actually did nothing. They were there, and were being read, but their values never actually affected the host.

Just as before, the proper way to configure the hostname and the mDNS state is through the domain settings page in the UI.

PR #2771, opened 2021-02-03
Use prettier for formatting.

mrstegeman:


mrstegeman:
@benfrancis @tim-hellhake this doesn't really need a full review, as almost all of the changes were automated, just getting some buy-in on using this tool. I strongly prefer to have a code base with a consistent style, and this generates code that is very close to what we were writing anyway.

tim-hellhake (approved):

.github/workflows/build.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       - name: Install node dependencies
         run: |
           npm ci
+      - name: Check formatting
+        run: |
+          npx prettier -u -c '*.ts' '*.js' config src static tools
```
benfrancis, line 35: Is this treating .js/.ts differently to .html/.css in the GitHub workflow?

benfrancis (approved):
I agree with having a consistent coding style. I'm not sure how I feel about code being re-formatted automatically (as opposed to hinted at via a linter), because sometimes humans break formatting rules for good reasons! Having said that, I do think most of the changes the tool made to our existing code were improvements.

I'm not a big fan of all the extra white-space around parentheses and some of the HTML edits are a bit extreme, but they're really just a matter of taste.

Happy to give it a go and see how it goes. It seems simple enough to remove it if it causes issues.

.github/workflows/build.yml
mrstegeman: No, they're just globs for the files that should be formatted.

PR #2778, opened 2021-02-05
Fix text input CSS after minimization.

mrstegeman:


benfrancis (commented):
It's a bit difficult to tell what this is fixing. Which commit broke it? The context might help.

mrstegeman:
Upgrading webpack. The HTML minimizer changed, so it seems to strip the `type="text"` from text inputs, since it's the default.

benfrancis (approved):
I hate webpack.

Issue #2806, opened 2021-03-08
Change `links` to `forms` in API


benfrancis:
Blocks: #2802

The gateway's API currently exposes `links` for each interaction affordance (property, action and event) which provide API endpoints for getting/setting properties, invoking actions etc. E.g.

```json
  "properties": {
    "on": {
      "links": [{"href": "/things/switch/properties/on"}]
    }
```

The W3C WoT Thing Description specification specifies the use of `forms` rather than `links` for many of these use cases. E.g.

```json
  "properties": {
    "on": {
      "forms": [{"href": "/things/switch/properties/on"}]
    }
```

I still do not like the forms approach of the W3C spec (see https://github.com/w3c/wot-thing-description/issues/85 and https://github.com/w3c/wot-thing-description/issues/88 for some previous discussions), but given for simple use cases the two are functionally equivalent, let's pick our battles and make this conformant with the W3C spec.

**This includes:**
1. Changing instances of `links` to `forms` where appropriate
2. Changing `mediaType` to `contentType` where used
3. Removing `rel` (e.g. `rel: "property"`) from links which become forms
4. Adding an `op` member of a form where needed (this is not needed in all instances due to [default value definitions](https://www.w3.org/TR/wot-thing-description/#sec-default-values))
5. Updating the gateway's front end to expect this

**This excludes:**
1. Changing the payloads of API requests/responses to match data schemas in the Thing Description (e.g. removing object wrappers),  or changing data schemas in Thing Descriptions to exactly match payloads (e.g. adding object wrappers or `response` metadata). This will be handled separately.
2. Representing [action queues](https://github.com/w3c/wot-thing-description/issues/302) and [event logs](https://github.com/w3c/wot-thing-description/issues/892) in forms, which is not currently supported in the W3C spec. For the purposes of this issue we should just provide a form with a `href` endpoint for these interaction affordances and continue to follow the conventions of the Web Thing API.
3. The alternate link relation in the top level [links](https://webthings.io/api/#links-member) member used to link to an HTML UI should stay as a link:

```json
"links": [
  {
    "rel": "alternate",
    "mediaType": "text/html",
    "href": "/things/lamp"
  }
]
```

**Open questions:**

1. Should a top level link to a WebSocket API endpoint be a link or a form? For example:

```json
"links": [
  {
    "rel": "alternate",
    "href": "wss://mywebthingserver.com/things/lamp"
  }
]
```

could become:

```json
"forms": [
  {
[...]

PR #2811, opened 2021-03-29
W3C compliance - Use forms instead of links in Properties, Actions, and Events

relu91:
This PR covers the changes described in #2806 but it still does not tackle the open questions. It is merely a refactor of the current code base divided into small commits. I also introduced a test to verify the correct behavior of the Camera component in the Web UI. 

The PR is still in the draft because we still need to understand how much this change will impact the addons and how to cope with the _root level_ forms (for further details refer to #2806). 

*Note* this change requires the updated version of https://github.com/WebThingsIO/gateway-addon-node in my forked repository. If you want to test this PR locally do the following:
1. clone https://github.com/relu91/gateway-addon-node (be careful to clone the correct submodule: https://github.com/relu91/gateway-addon-ipc-schema)
2. npm link
3. cd your gateway folder 
4. npm link gateway-addon


benfrancis:
I'm having problems running this:

```
$ npm start

> webthings-gateway@1.0.0 start /home/tola/Code/relu91/gateway
> npm run build && node build/app.js


> webthings-gateway@1.0.0 build /home/tola/Code/relu91/gateway
> rm -rf build && cp -rL src build && find build -name '*.ts' -delete && tsc -p . && webpack

src/models/thing.ts:144:28 - error TS2571: Object is of type 'unknown'.

144           property.forms = property.forms.map((form) => {
                               ~~~~~~~~~~~~~~

src/models/thing.ts:144:48 - error TS7006: Parameter 'form' implicitly has an 'any' type.

144           property.forms = property.forms.map((form) => {
                                                   ~~~~

src/models/thing.ts:159:9 - error TS2571: Object is of type 'unknown'.

159         property.forms.push({
            ~~~~~~~~~~~~~~

src/models/thing.ts:229:24 - error TS2571: Object is of type 'unknown'.

229         action.forms = action.forms.map((form) => {
                           ~~~~~~~~~~~~

src/models/thing.ts:229:42 - error TS7006: Parameter 'form' implicitly has an 'any' type.

229         action.forms = action.forms.map((form) => {
                                             ~~~~

src/models/thing.ts:242:7 - error TS2571: Object is of type 'unknown'.

242       action.forms!.push({
          ~~~~~~~~~~~~~

src/models/thing.ts:255:23 - error TS2571: Object is of type 'unknown'.

255         event.forms = event.forms.map((form) => {
                          ~~~~~~~~~~~

src/models/thing.ts:255:40 - error TS7006: Parameter 'form' implicitly has an 'any' type.

255         event.forms = event.forms.map((form) => {
                                           ~~~~

src/models/thing.ts:268:7 - error TS2571: Object is of type 'unknown'.

268       event.forms.push({
          ~~~~~~~~~~~

src/models/thing.ts:635:28 - error TS2571: Object is of type 'unknown'.

635           property.forms = property.forms.map((form) => {
                               ~~~~~~~~~~~~~~

src/models/thing.ts:635:48 - error TS7006: Parameter 'form' implicitly has an 'any' type.

635           property.forms = property.forms.map((form) => {
                                                   ~~~~

src/models/thing.ts:650:9 - error TS2571: Object is of type 'unknown'.

650         property.forms.push({
            ~~~~~~~~~~~~~~

src/models/thing.ts:667:24 - error TS2571: Object is of type 'unknown'.

667         action.forms = action.forms.map((form) => {
                           ~~~~~~~~~~~~

src/models/thing.ts:667:42 - error TS7006: Parameter 'form' implicitly has an 'any' type.

667         action.forms = action.forms.map((form) => {
                                             ~~~~

src/models/thing.ts:680:7 - error TS2571: Object is of type 'unknown'.

680       action.forms!.push({
          ~~~~~~~~~~~~~

src/models/thing.ts:695:23 - error TS2571: Object is of type 'unknown'.

695         event.forms = event.forms.map((form) => {
                          ~~~~~~~~~~~

src/models/thing.ts:695:40 - error TS7006: Parameter 'form' implicitly has an 'any' type.

695         event.forms = event.forms.map((form) => {
                                           ~~~~

src/models/thing.ts:709:7 - error TS2571: Object is of type 'unknown'.

709       event.forms.push({
          ~~~~~~~~~~~

src/plugin/property-proxy.ts:85:12 - error TS2339: Property 'setForms' does not exist on type 'PropertyProxy'.

85       this.setForms(propertyDict.forms);
              ~~~~~~~~


Found 19 errors.

npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! webthings-gateway@1.0.0 build: `rm -rf build && cp -rL src build && find build -name '*.ts' -delete && tsc -p . && webpack`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the webthings-gateway@1.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tola/.npm/_logs/2021-03-30T11_58_11_237Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! webthings-gateway@1.0.0 start: `npm run build && node build/app.js`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the webthings-gateway@1.0.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tola/.npm/_logs/2021-03-30T11_58_11_419Z-debug.log
```

These look like TypeScript errors to  me (I'm new to TypeScript).

Maybe I didn't link the dependency correctly though?

relu91:
Yep, it is our friendly typescript compiler. Did you build the gateway-addon? sorry I forgot to mention it, my bad.

benfrancis:
This code all looks good to me so far, thanks.

Not sure if @tim-hellhake wants to take a look from a TypeScript point of view since that's new to me?

In testing this I installed the virtual things adapter and noticed lots of type errors in the console like the one below:
```
2021-03-30 15:37:34.106 ERROR  : virtual-things-adapter: strict mode: missing type "object" for keyword "properties" at "https://raw.githubusercontent.com/WebThingsIO/gateway-addon-ipc-schema/master/messages/definitions.json#/allOf/1" (strictTypes)
```

I was able to add a virtual thing, but trying to set properties results in a 400 error with the payload

`title parameter required`

It looks like virtual-things-adapter might need to be updated :/

I haven't tested it with any other adapters yet.

benfrancis:
For the record, in addition to the steps above, I needed to build gateway-addon-node with a custom build command which doesn't do a submodule update and switch away from @relu91's w3c-compliance branch:
`
node generate-version.js && node generate-types.js && ./node_modules/typescript/bin/tsc -p .`

before doing the npm link step.

relu91:
Some updates on this PR. I tried to figure out the impact of changing `links` to `forms` at the IPC layer. Since the list of addons is quite long I tried a naive automated approach. Putting simply I cloned every single addons locally and searched for the string `link`. In [this gists](https://gist.github.com/relu91/23b5d3833274253b83d31f08a968d4aa), you can find the script that I used to generate a [report](https://gist.github.com/relu91/b31c03ecc91b34b137f7a9b741bfdcfa). 

The  [report](https://gist.github.com/relu91/b31c03ecc91b34b137f7a9b741bfdcfa) was then manually reviewed to eliminate false positives. So currently I find out 9 addons that should be updated after merging this PR. The addons marked as **~~yes~~ no** were found not really impacted by the change and therefore discarded.

Since I am aware of the weaknesses of this process I would rate the different marks as follows:
- no: very likely the project would not be affected -> it never uses the links property directly
- ~~yes~~ no: it might be affected cause it contained usages of links. but most of the times they manipulated links at the root level
- yes: something might breaks -> we should open an issue to warn the maintainers about the upcoming changes

benfrancis:
That's extremely useful, thank you @relu91.

So my understanding is that if we want to rename links to forms at the IPC layer as well as the web layer, the following add-ons would be impacted:
- followers: @flatsiedatsie
- frontier-silicon-adapter: @bewee
- highlights: @flatsiedatsie
- onvif-adapter: @WebThingsIO
- sonos-adapter: @tim-hellhake
- spotify-adapter: @tim-hellhake
- thing-url-adapter: @WebThingsIO
- virtual-things-adapter: @WebThingsIO
- zwave-adapter: @WebThingsIO

This seems like good news to me.

@flatsiedatsie @bewee @tim-hellhake would you be willing to rename "links" to "forms" in your add-ons before 2.0 is released?

We can make the change to the web layer only, but it would be much better to have it be consistent all the way down the stack.

We'd need to figure out an update strategy so that add-ons keep working while the upgrade is still rolling out.

flatsiedatsie:
Ofcourse, I'd be happy to. 

Would the addons still work for people on the old version of the gateway is the addon started using the new names?

relu91:
> Would the addons still work for people on the old version of the gateway is the addon started using the new names?

I think it depends on addons maintainers. If in the logic of your addon you want to maintain backward compatibility you can easily check for links or forms and behave differently. There is no other change in the IPC layer so I think they will just continue to work. 

But @benfrancis has the final word here :)

benfrancis:
> check for links or forms and behave differently

If this is straightforward then it would be a good practice to recommend, so that add-ons can work with both 1.x and 2.x until 2.x becomes more widely adopted.

tim-hellhake:
@benfrancis Sure!

tim-hellhake:
@relu91 @benfrancis 
I merged WebThingsIO/gateway-addon-ipc-schema#18 and WebThingsIO/gateway-addon-node#100 and released the new version [2.0.0-alpha.1](https://github.com/WebThingsIO/gateway-addon-node/commit/19ae94ac829663ed0fab3caebc2908a38bbf9a7d).

tim-hellhake:
@relu91 @benfrancis 
Would [that](https://github.com/tim-hellhake/sonos-adapter/pull/25) be sufficient?

relu91:
@tim-hellhake Thanks! let me update the dependency and see if the cli is ok with it 👍🏻

relu91:
I tried to dig a little bit about why on node 10 we go a timeout in the tests. Locally, I can sometimes reproduce the issue. It seems that the virtual-things-adapter once installed it causes problems in the clean-up procedures. In particular, AFIK the `addonManger.unistall` is called before calling `addonManager.unloadAddons`. Since in the `unistall` we kill the process when `virtual-things-adapter` is unloaded, it also attempts to kill the process, but I think that there are everything blocks. 

Not sure why this problem didn't pop up also in all the other PRs.

benfrancis:
Note: While I'm waiting for #2871 to land I've been continuing work on the links -> forms change in https://github.com/benfrancis/gateway/tree/links-to-forms which also includes those changes.

static/js/schema-impl/capability/thing.js
```diff
@@ -301,9 +301,9 @@ class Thing {
           const action = description.actions[name];
 
           let href;
-          for (const link of description.actions[name].links) {
-            if (link.rel === 'action') {
-              href = link.href;
+          for (const form of description.actions[name].forms) {
+            if (!form.op || form.op === Constants.WoTOperation.INVOKE_ACTION) {
+              href = form.href;
```
benfrancis, line 306: In contrast to the above I think this is probably safe since it should pick the form generated by the gateway which is the last in the array and contains no op member.

static/js/views/things.js
```diff
@@ -216,9 +216,9 @@ const ThingsScreen = {
         }
 
         let href;
-        for (const link of description.actions[actionName].links) {
-          if (link.rel === 'action') {
-            href = link.href;
+        for (const form of description.actions[actionName].forms) {
+          if (!form.op || form.op === Constants.WoTOperation.INVOKE_ACTION) {
+            href = form.href;
```
benfrancis, line 221: Same here.

benfrancis (changes requested):
A couple of nits and I have one concern about how the forms are selected in the front end. I'm also going to suggest a follow-up to add backwards compatibility for Thing Descriptions which use links instead of forms so that we don't break existing add-ons. But otherwise I think this is very close to being landed!

benfrancis:
I've filed a follow-up issue regarding backwards-compatibility https://github.com/WebThingsIO/gateway/issues/2881

benfrancis:
I've also filed a follow-up for the WebSocket endpoint, which I've realised is currently still a link #2882

static/js/models/thing-model.js
```diff
@@ -194,13 +194,7 @@ class ThingModel extends Model {
...
     const property = this.propertyDescriptions[name];
 
-    let href;
-    for (const link of property.links) {
-      if (!link.rel || link.rel === 'property') {
-        href = link.href;
-        break;
-      }
-    }
+    const href = property.forms[0].href;
```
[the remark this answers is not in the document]
relu91: Please check again the selection function that I built

static/js/utils.ts
```diff
@@ -264,3 +265,37 @@ export function adjustInputValue(value: number, schema: Record<string, unknown>)
...
+      if (error instanceof TypeError) {
+        // URL is relative or not well formatted
+        return (
+          !selectedForm.op || selectedForm.op === operation || selectedForm.op?.includes(operation)
+        );
+      }
+      throw error;
+    }
+  })?.href;
+}
```
relu91, line 301: Is this file the right place to introduce this utility function?

I use the spread operator and the `reverse` function, although it might be not so efficient for a large `forms` array I think it is more expressive than a `for` loop that searches the `form` backward. 

About the selection logic, tell me if you find it satisfactory. I choose to check also for protocol, maybe is an overkill? wot-adapter may publish TDs with other protocols in forms..

static/js/models/thing-model.js
```diff
@@ -216,7 +217,7 @@ class ThingModel extends Model {
     if (typeof this.propertiesHref === 'undefined') {
       const urls = Object.values(this.propertyDescriptions).map((v) => {
         if (v.forms) {
-          return v.forms[0].href;
+          return Utils.selectFormHref(v.forms, Constants.WoTOperation.WRITE_PROPERTY);
```
relu91, line 220: I think for extensibility it might be right to look for `writeproperty` operation here even if now it is just reduntat.

static/js/utils.ts
relu91: I also wanted to provide unit tests for this particular function, but it is not easy to test the frontend code so I gave up. Feel free to suggest how to properly test this.

static/js/schema-impl/capability/thing.js
```diff
@@ -300,13 +294,7 @@ class Thing {
...
           const action = description.actions[name];
 
-          let href;
-          for (const form of description.actions[name].forms) {
-            if (!form.op || form.op === Constants.WoTOperation.INVOKE_ACTION) {
-              href = form.href;
-              break;
-            }
-          }
+          const href = Utils.selectFormHref(action.forms, Constants.WoTOperation.INVOKE_ACTION);
```
relu91, line 297: For consistency, I always used `selectFormHref` when looking for the gateway form.

static/js/utils.ts
```diff
@@ -264,3 +265,37 @@ export function adjustInputValue(value: number, schema: Record<string, unknown>)
...
+  return [...forms].reverse().find((selectedForm) => {
+    try {
+      const { protocol } = new URL(selectedForm.href);
+      return (
+        protocol === 'http:' &&
+        (!selectedForm.op || selectedForm.op === operation || selectedForm.op?.includes(operation))
+      );
+    } catch (error) {
+      if (error instanceof TypeError) {
+        // URL is relative or not well formatted
```
benfrancis, line 293: I think I'm right in saying that all the form `href`s added by the gateway are just paths rather than absolute URLs (e.g. `/things/foo/properties/on`). That means this logic would always be triggered and the protocol check would never be applied.

Rather than just ignore the protocol if not an absolute URL, I think if you're going to check the protocol what you need to do here is resolve the `href` against the `base` member of the Thing Description first.

This is probably strictly not necessary since the last form entry (added by the gateway) will always be HTTP, so we could just remove the protocol check if you prefer (see below).

benfrancis (changes requested):
This looks good, but I suggest either removing the protocol check or first resolving `href `against `base` before checking for http: **or https:**. (See further comments inline).

I think we could make this all more efficient by parsing the forms once on instantiation and storing a normalised model of the thing's affordances which has a URL endpoint per operation. Then we don't need to parse the forms every time we want to use them. However, this could be part of wider work to consolidate the Thing Description parsing logic and I'm happy to file a follow-up issue for that.

flatsiedatsie:
I've been trying Gateway 1.1.0, and have run into this change. Some of my adons look for the `links` array.

I noticed that there is not a forms array. But the links array is also still there, just.. empty.

Perhaps the gateway could populate both the forms and links array? Then it would remain backwards compatible with these addons.

PR #2818, opened 2021-04-27
Update dependencies

benfrancis:
Note the new package `@types/jest` which was needed to resolve TypeScript compilation errors after upgrading other packages.


benfrancis:
@tim-hellhake I'm struggling to get integration tests to pass on this PR, but I just tried the integration tests on master and they fail there too.

Are integration tests currently passing for you, and do you have any idea what might be causing these failures?

PR #2821, opened 2021-04-29
Bump gateway version to 1.1.0

benfrancis:


tim-hellhake (approved):
Looks good by I wonder how we tag a second pre-release.
Do we remove the current pre-release first?

benfrancis:
> Looks good by I wonder how we tag a second pre-release.
> Do we remove the current pre-release first?

If I've understood correctly I think we create a new git tag for each pre-release but this metadata stays the same https://github.com/WebThingsIO/wiki/wiki/How-To-Release-a-Gateway-OTA-Update

Since we now have new user-facing features on master (groups) I renamed the release from 1.0.1 to 1.1.0 and will create a branch to continue development on that release, so that work on 2.0 can start to land on master.

Issue #2357, opened 2019-12-09
Thing Settings


benfrancis:
As a user I want to configure a Thing so I can:
- Show/hide the device on floorplan view
- Show/hide individual properties and actions in detail view
- etc.

PR #2835, opened 2021-06-04
Add "show on floorplan" option to edit thing page

bewee:
This allows to hide a thing from the floorplan view via its edit page, as suggested in #2357.

image


tim-hellhake (approved):

tim-hellhake:
Good idea!

PR #2850, opened 2021-07-07
Migrate to rocket 5

bewee:


src/main.rs
```diff
@@ -19,49 +18,53 @@ mod process_manager;
...
+        actix::spawn(async {
+            AddonManager::from_registry()
+                .send(LoadAddons)
+                .await
+                .expect("Sending LoadAddons message ");
+        });
 
-    thread::spawn(|| {
-        rocket().launch();
+        system.run().expect("Running system");
```
tim-hellhake, line 50: This is a blocking call.
I think we need to pull that out of the `spawn` call.
I'll do it after the PR is merged.

Issue #2817, opened 2021-04-23
Remote access isn't working


uqlel:
Hello,
After installing webthings gateway on raspbian using .deb file and setuping, when i try to connect to my domain, Firefox throws SSL_ERROR_ACCESS_DENIED_ALERT error. 
`grep -i pagekite run-app.log` shows this:
```bash
2021-04-23 10:56:30.525 INFO   : [pagekite] stderr: /bin/sh: 1: /opt/webthings-gateway/pagekite.py: not found
2021-04-23 10:56:30.532 INFO   : [pagekite] process exited with code 127
2021-04-23 11:04:05.528 INFO   : [pagekite] stderr: /bin/sh: 1: /opt/webthings-gateway/pagekite.py: not found
2021-04-23 11:04:05.533 INFO   : [pagekite] process exited with code 127
2021-04-23 11:10:45.770 INFO   : [pagekite] stderr: /bin/sh: 1: /opt/webthings-gateway/pagekite.py: not found
2021-04-23 11:10:45.774 INFO   : [pagekite] process exited with code 127
```
Sorry for my bad english :/

uqlel:
I can detect pagekite.net:
```bash
root@rpi:/etc/webthings-gateway/log# ping pagekite.net
PING pagekite.net (69.164.211.158) 56(84) bytes of data.
64 bytes from ns1.pagekite.net (69.164.211.158): icmp_seq=1 ttl=53 time=102 ms
64 bytes from ns1.pagekite.net (69.164.211.158): icmp_seq=2 ttl=53 time=100 ms
64 bytes from ns1.pagekite.net (69.164.211.158): icmp_seq=3 ttl=53 time=100 ms
^C
--- pagekite.net ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2ms
rtt min/avg/max/mdev = 100.049/100.675/101.603/0.762 ms```

uqlel:
I have also changed `--inh-caps=-all` to `--inh-caps=-chown` in `/usr/bin/webthings-gateway` because i had the same issue as in https://github.com/WebThingsIO/gateway-aur/issues/4

uqlel:
`find / -name "*pagekite*"` returns nothing, i think that i'm missing something

benfrancis:
I'm afraid I've never actually used the .deb package so I don't have much experience with that and it's currently lacking a maintainer, but what I can say is:
- You shouldn't need to access pagekite.net
[...]

PR #2853, opened 2021-07-20
Tweak recommended dependencies to avoid PageKite issue discovered in #2817

benfrancis:
I discovered that #2817 reproduces when building from source too if python2 is not installed, because PageKite depends on python2.

The current list of recommended dependencies includes python-pip which doesn't exist on Ubuntu. I propose removing this and instead adding "python-six" which exists on both Debian and Ubuntu and should pull in python2 as well, then removing the pip install of python-six.

I'm not sure how different the Debian package of python-six is to the pip one, but this seemed to work for me. We should separately figure out why the Debian package doesn't seem to be pulling python2 in as a dependency in #2817.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,8 @@ $ sudo apt install \
     libudev-dev \
     libusb-1.0-0-dev \
     pkg-config \
-    python-pip \
+    python-six \
     python3-pip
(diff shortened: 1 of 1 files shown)


benfrancis:
r=me

PR #2863, opened 2021-08-11
Server-Sent Events support - closes #2830

### LINKED ISSUE #2830: Add support for Server Sent Events
In order to make the Web Thing API exposed by the gateway W3C compliant, one thing we may need to do is to change current [Event](https://webthings.io/api/#event-resource) and [Events](https://webthings.io/api/#events-resource) resources to return a Server Sent Events `EventSource` which pushes events to clients, rather than the current static JSON log of events which needs to be polled.

See also https://github.com/WebThingsIO/gateway/issues/2807#issuecomment-825907028

benfrancis:


benfrancis:
Work-in-progress Server-Sent Events implementation for W3C WoT compliance (#2830), which is currently blocking #2806.

This still needs tests. It's a bit tricky to test manually because Postman doesn't support Server-Sent events, but you can use curl to see the raw output of the event stream.

Note that JavaScript Consumers using `EventSource` you will need to provide a JWT in a query string (as is currently the case with WebSockets), because the `EventSource` API doesn't allow sending an `Authorization` header. I've used the same approach in the example curl command below, you just need to insert your own JWT.


`curl http://localhost:8080/things/virtual-things-10/events/virtualEvent?jwt={jwt} -H 'Accept: text/event-stream'`

For testing you can use the Virtual Actions & Events Thing from the Virtual Things add-on, which emits an event when you invoke the "no input" action.

benfrancis:
OK, I've added integration tests. This is now ready for review.

benfrancis:
@relu91 I'd be interested in your review here too if you get chance.

src/test/integration/things-test.ts
```diff
@@ -1021,6 +1035,70 @@ describe('things/', function () {
     expect(res.body[0].a).toHaveProperty('timestamp');
   });
 
+  it('should be able to subscribe to events using EventSource', async () => {
+    await addDevice(EVENT_THING);
+    if (!httpServer.address()) {
```
relu91, line 1040: Shouldn't this check be redundant? Servers are already initialized in `../common`. Is there any compelling reason to keep this initialization?

src/test/integration/things-test.ts
```diff
@@ -1021,6 +1035,70 @@ describe('things/', function () {
...
+    const overheatedEvent = new Event('overheated', 101, EVENT_THING.id);
+    const [, event] = await Promise.all([
+      Events.add(overheatedEvent),
+      e2p(eventSource, 'overheated'),
+    ]);
+    expect(event.type).toEqual('overheated');
+    expect(JSON.parse(event.data)).toEqual(101);
+    eventSource.close();
+
+    // Test events subscription
```
relu91, line 1061: I'd create a dedicated test for this

src/test/integration/things-test.ts
```diff
@@ -1021,6 +1035,70 @@ describe('things/', function () {
...
+    const overheatedEvent2 = new Event('overheated', 101, EVENT_THING.id);
+    const [, event2] = await Promise.all([
+      Events.add(overheatedEvent2),
+      e2p(eventsSource, 'overheated'),
+    ]);
+    expect(event2.type).toEqual('overheated');
+    expect(JSON.parse(event2.data)).toEqual(101);
+    eventsSource.close();
+
+    // Test non-existent thing errors
```
relu91, line 1076: same here

src/test/integration/things-test.ts
```diff
@@ -1021,6 +1035,70 @@ describe('things/', function () {
...
+      `http://127.0.0.1:${addr.port}${Constants.THINGS_PATH}` +
+      `/non-existent-thing/events/overheated?jwt=${jwt}`;
+    const thinglessEventSource = new EventSource(eventSourceURL) as EventTarget & EventSource;
+    thinglessEventSource.onerror = jest.fn();
+    thinglessEventSource.onopen = jest.fn();
+    await e2p(thinglessEventSource, 'error');
+    expect(thinglessEventSource.onopen).not.toBeCalled();
+    expect(thinglessEventSource.onerror).toBeCalled();
+
+    // Test non-existent event errors
```
relu91, line 1087: here too.

src/test/integration/things-test.ts
```diff
@@ -1021,6 +1035,70 @@ describe('things/', function () {
...
+      `${EVENT_THING.id}/events/non-existentevent?jwt=${jwt}`;
+    const eventlessEventSource = new EventSource(eventSourceURL) as EventTarget & EventSource;
+    eventlessEventSource.onerror = jest.fn();
+    eventlessEventSource.onopen = jest.fn();
+    await e2p(eventlessEventSource, 'error');
+    expect(eventlessEventSource.onopen).not.toBeCalled();
+    expect(eventlessEventSource.onerror).toBeCalled();
+
+    // Clean up
+    httpServer.close();
```
relu91, line 1099: As above server lifecycle is somewhat handled in `../common` :
```ts
afterAll(async () => {
  Logs.close();
  await addonManager.unloadAddons();

  if (servers.https) {
    servers.https.close();
    await e2p(servers.https, 'close');
  }

  servers.http.close();
  await e2p(servers.http, 'close');

  removeTestManifest();
});
```
Do we really need this?

relu91 (changes requested):
The PR looks good, aside from some comments below I'd refactor the unit test provided in smaller tests. I understand that logically they all belong to the same category (sse tests) but if we want to group them logically we can just use the describe function:
```js
describe("Server Sent Event testing", () => {
  it("should subscribe", ()=>{})
  it("should subscribe to multiple events", ()=>{})
  it("should fail for not existing events", ()=>{})
}); 
```
What do you think?

src/controllers/events_controller.ts
```diff
@@ -1,42 +1,147 @@
...
+      const eventId = Date.now();
+
+      // Push event to client via event stream
+      response.write(`id: ${eventId}\n`);
+      response.write(`event: ${event.getName()}\n`);
+      response.write(`data: ${JSON.stringify(event.getData())}\n\n`);
+    }
+
+    // Subscribe to events from the specified Thing
+    if (thing) {
```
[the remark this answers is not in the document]
benfrancis: Ah yes, you're correct.

It gets a bit confusing because the gateway can have events which are associated with itself rather than a thing. I decided not to provide an SSE endpoint for those since we'll probably want to handle that if and when we implement the Directory Service API, but I forgot to remove the conditional after adding those checks.

src/test/integration/things-test.ts
benfrancis: I got in a bit of a pickle trying to get a reference to the server port, with certificate errors and then problems with jest hanging on a server that wasn't closed, so I copied this approach from websocket-util.js. But I just tried again and you're right, there's a much simpler way of doing this. Phew.

benfrancis:
Thanks for the review @relu91, great feedback. I think I've addressed all your comments.

relu91 (approved):
Perfect 👍🏻

benfrancis:
@tim-hellhake agreed to make @relu91 a peer of the [gateway module](https://github.com/WebThingsIO/wiki/wiki/Modules#webthings-gateway), so landing based on @relu91's review.

@tim-hellhake comments still welcome if you want to review.

tim-hellhake (approved):

PR #2864, opened 2021-08-16
Update dependencies

benfrancis:
Squashing together lots of pending pull requests (including those from dependabot) to update dependencies.

PR #2865, opened 2021-08-17
Update dependencies

benfrancis:
This is a second attempt at the dependency updates from #2864. It currently excludes the update to postcss proposed by dependabot in #2842 which is generating errors.


benfrancis:
r=me

Issue #2807, opened 2021-03-08
Remove object wrappers from API payloads


benfrancis:
Blocks: #2802

Currently for a property definition in a Thing Description exposed by the gateway as below:

```json
...
    "temperature": {
      "type": "number",
      "links": [{"href": "/things/pi/properties/temperature"}]
    }
...
```

 a response to a GET request on the Property resource looks something like:

```json
{
  "temperature": 21
}
```

where the payload is a JSON object with a map representation of the property's name and value. The payloads of PUT requests and responses also follow the same format.

To be compliant with the W3C Thing Description specification, let's change payloads to match the data schema in the Thing Description, e.g. for the above example this would simply be:

```json
21
```

See https://github.com/WebThingsIO/api/issues/143 for some background on this topic.

**This includes:**
1. Removing object wrappers from HTTP requests and responses on a [Property](https://webthings.io/api/#property-resource) resource when reading and writing properties
2. Removing object wrappers from HTTP requests and responses when invoking an action on an [Action](https://webthings.io/api/#action-resource) resource
3. Updating the gateway's front end to expect this

**The excludes:**
1. Removing object wrappers from [Properties](https://webthings.io/api/#properties-resource) resources (I seem to recall the W3C specification actually defined an object wrapper for `readallproperties` and `writeallproperties` ops with values keyed by property names, but I can't find it in the current draft, can anyone confirm this?)
2. Changing the payload of the [Things](https://webthings.io/api/#things-resource) resource, which will hopefully be defined in the WoT Discovery specification as part of the Directory Service API
3. Changing payloads of any messages in the WebSocket API, which need the additional context

**Open questions:**
1. What should we do about [Event](https://webthings.io/api/#event-resource) resources, which currently return an array of objects in their HTTP response?
2. What should we do about [Events](https://webthings.io/api/#events-resource) resources, which currently don't have an obvious representation which is W3C compliant?
3. What should we do about [Action](https://webthings.io/api/#action) resources which represent an action queue?
4. What should we do about [Actions](https://webthings.io/api/#actions-resource) resources?

benfrancis:
Some thoughts on the open questions...

> 1. What should we do about [Event](https://webthings.io/api/#event-resource) resources, which currently return an array of objects in their HTTP response?
> 2. What should we do about [Events](https://webthings.io/api/#events-resource) resources, which currently don't have an obvious representation which is W3C compliant?

[...]

PR #2871, opened 2021-09-07
W3C compliant actions protocol binding

benfrancis:
This is an implementation to make the invoke action part of the gateway's API W3C compliant and is a partial fix to #2807.

- [x] Remove object wrapper from invoke action API requests (to make action inputs directly match the action input data schema in Thing Descriptions)
- [x] Update the front end to consume the new API
- [x] Fix integration tests

My current thinking is to leave the rest of the API intact until the [proposed action protocol binding](https://github.com/w3c/wot-profile/pull/89) for WoT Profile is finalised, so that we don't have to implement it twice. The first step is to simply make the `invokeaction` operation API match the Thing Description.

~~One thing I'm not sure about is what to do with the /actions endpoint where no action name is specified in the URL, but one option is to leave it as it is and just remove that endpoint from Thing Descriptions until the Thing Description specification has a way to describe it (see https://github.com/w3c/wot-thing-description/issues/1200).~~

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -273,7 +273,12 @@ function createApp(isSecure: boolean): express.Application {
       extended: false,
     })
   );
-  app.use(bodyParser.json({ limit: '1mb' }));
+  app.use(
+    bodyParser.json({
+      limit: '1mb',
+      strict: false,
+    })
+  );
 
   // Use fileUpload to handle multi-part uploads
   app.use(fileUpload());
--- a/src/controllers/actions_controller.ts
+++ b/src/controllers/actions_controller.ts
@@ -95,47 +95,26 @@ function build(): express.Router {
    */
   controller.post('/:actionName', async (request, response) => {
     const actionName = request.params.actionName;
-
-    const keys = Object.keys(request.body);
-    if (keys.length != 1) {
-      const err = 'Incorrect number of parameters.';
-      console.log(err, request.body);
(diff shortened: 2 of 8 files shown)


benfrancis:
I've fixed the integration tests and this is now ready for review.

What this PR does:
- Removes the object wrapper from `invokeaction` operations on Action endpoints  (e.g. /things/mything/actions/foo) so that the payload directly matches the input schema for the action affordance in the Thing Description

What this PR does not do:
- Change the API for querying or cancelling an event
- Change the API for invoking an action via an Actions endpoint (e.g. /things/mything/actions/)

My thinking is that in this first phase we will just make the `invokeaction` and `queryallactions` operations W3C compliant in order to be able to provide a W3C compliant Thing Description. The API for the other operations will still exist, it just won't be described by the Thing Description.

In a later phase we can fully implement the new actions protocol binding from the WoT Profile specification and also expose `queryaction` and `cancelaction` operations in Thing Descriptions.

Note that we will need to add the new `queryallactions` operation to Forms in #2811.

benfrancis:
@relu91 FYI I think we need to land this before landing https://github.com/relu91/gateway/pull/2 and https://github.com/WebThingsIO/gateway/pull/2811 because they are starting to conflict with one another.

relu91 (approved):
No objections! looks good.

PR #2868, opened 2021-08-24
W3C compliant properties protocol binding

benfrancis:
This is a work-in-progress implementation to make the properties part of the gateway's API W3C compliant and is a partial fix to #2807.
- [x] Remove object wrapper from property API requests (to make API payloads directly match property data schema in Thing Descriptions)
- [x] Provide an empty response to set property PUT requests (to match the [properties protocol binding](https://w3c.github.io/wot-profile/#properties) of the Core Profile, which will make it easier to conform with that profile in the future)
- [x] Update the front end to consume the new API
- [ ] Fix the litany of failing integration tests as a result of changing the API

PR #2875, opened 2021-09-20
W3C compliant properties protocol binding

benfrancis:
This is a replacement for #2868 which makes the properties part of the gateway's API W3C compliant and is a partial fix to #2807. Unlike #2868, set property operations still have a payload in their response since it turns out the front end depends on this.

- [x] Remove object wrapper from property API requests (to make API payloads directly match property data schema in Thing Descriptions)
- [x] Update the front end to consume the new API
- [x] Fix the litany of failing integration tests as a result of changing the API

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -273,7 +273,12 @@ function createApp(isSecure: boolean): express.Application {
       extended: false,
     })
   );
-  app.use(bodyParser.json({ limit: '1mb' }));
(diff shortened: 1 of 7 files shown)


benfrancis:
OK, I finally got all the tests to pass which was a bit of a mission. Ready for review.

relu91 (approved):

PR #2880, opened 2021-09-29
Enable wot-adapter as alternative of thing-url-adapter

relu91:
As planned in the wot compliance roadmap, this PR is introducing the https://github.com/WebThingsIO/wot-adapter as an alternative of the https://github.com/WebThingsIO/thing-url-adapter. Basically, the gateway treats the https://github.com/WebThingsIO/thing-url-adapter as a special addon-on and enables dedicated UI elements and features to it. This PR **extends** this special treatment also for the https://github.com/WebThingsIO/wot-adapter.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -68,3 +68,17 @@ export function getGatewayUserAgent(): string {
 
   return `${primary} ${secondary}${tertiary}`;
 }
+
+export function isWebthingsThingDescription(thingDescription: {
+  '@context': string | string[];
+}): boolean {
+  if (typeof thingDescription['@context'] === 'string') {
+    return thingDescription['@context'] === 'https://webthings.io/schemas';
+  }
+
+  if (Array.isArray(thingDescription['@context'])) {
+    return thingDescription['@context'].includes('https://webthings.io/schemas');
+  }
+
+  return false;
+}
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -68,3 +68,17 @@ export function getGatewayUserAgent(): string {
...
+  if (typeof thingDescription['@context'] === 'string') {
+    return thingDescription['@context'] === 'https://webthings.io/schemas';
+  }
+
+  if (Array.isArray(thingDescription['@context'])) {
+    return thingDescription['@context'].includes('https://webthings.io/schemas');
+  }
+
+  return false;
+}
```
benfrancis, line 84: This unfortunately doesn't work because the [`@context`](https://webthings.io/api/#web-thing-description) member in a Web Thing Description is optional. It's only needed if a developer wants to add semantic annotations using schemas from the schema repository at https://webthings.io/schemas/. There are also some adapters (like the virtual things adapter) which still set `@context` to https://iot.mozilla.org/schemas/

I think this function needs to work the other way around and check whether a Thing Description is a W3C Thing Description by looking for a `@context` of 

1. https://www.w3.org/2019/wot/td/v1
2. http://www.w3.org/ns/td
3. Whatever the 1.1 context is going to be

 as per https://github.com/w3c/wot-scripting-api/blob/main/typescript/thing-description/schema/td-json-schema-validation.json#L63

benfrancis (changes requested):
This looks good to me, except for the logic to detect a Mozilla Web Thing Description vs. a W3C Thing Description.

benfrancis (approved):
That's better, thanks.

PR #2884, opened 2021-10-13
Log graph improvements

### LINKED ISSUE #2366: data logging autoscale sometimes odd
Input from [MicroBlocks](microblocks.fun) team:

The gateway attempts to auto-scale the data log graphs, but it sometimes does strange things.

Here's one example:
image

One strange thing is that the base and maximum points are fractional numbers but the data are integral values of temperature. It would be nice if the autoscaler code could notice that the data only contained integers and pick integers for the base and max of the graph scale.

The other strange thing is that it didn't set the base value to be the minimum value in the displayed range. As it is, the graph display falls below the base line and overlaps with the scroll bar. (Maybe this would have been fixed with a page reload.)

If more dynamic auto-scaling is not easily accomplished, then there could be a "refresh button" to force re-scaling.

Finally, a feature that would make data logging more useful to science classes would be a way to download the logged data as a CSV file so it could be imported into other programs for analysis.

The graphing is already a terrific feature. Fixing these issues would make it even better!

hobinjk:
Fixes #2366

👋

### DIFF
--- a/static/js/logs/log.js
+++ b/static/js/logs/log.js
@@ -418,12 +418,15 @@ class Log {
   }
 
   determineBounds() {
-    if (this.hasOwnProperty('valueMin')) {
-      return;
-    }
     const bounds = this.valueBounds();
-    this.valueMin = bounds.min;
-    this.valueMax = bounds.max;
+    if (this.hasOwnProperty('valueMin')) {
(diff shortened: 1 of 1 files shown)


benfrancis:
Done.

PR #2886, opened 2021-10-24
Fix #2878

### LINKED ISSUE #2878: bug: TypeError: rsp.getStatus() is not a function
Hello there,
i've spent the whole day trying to figure out this error and i can't seem to track why it's happening:
`Error calling API handler: TypeError: rsp.getStatus is not a function`

how to reproduce:
using post man call this API with auth token: for example i will call it against the webhook-events adapter:
`https://mypage.webthings.io/extensions/webhook-events/api/test`

then in your console log you will get the following error:
`Error calling API handler: TypeError: rsp.getStatus is not a function`
`build/controllers/extensions_controller.js:61:33`
so it seems as if api-handler.ts isn't being loaded as that's where the `rsp.getStatus()` call resides and which is a simple wrapper to rsp.status . i've removed .nvm, reinstalled node 12.x directly and still to no avail.

the old version of the `extensions_controller.js` called rsp.status diretly without wrapper and worked as advertised...

can someone please check and let me know if it's a bug or me doing something wrong in my setup.

btw the change happened on commit Migrate to typescript https://github.com/WebThingsIO/gateway-addon-node/commit/ab4ae73f1cc265874913b3d4d55fd34522baa050

Jay

bewee:
Closes #2878

Not elegant, but it works.

### DIFF
--- a/src/controllers/extensions_controller.ts
+++ b/src/controllers/extensions_controller.ts
@@ -38,10 +38,7 @@ function build(): express.Router {
     });
 
     try {
-      const rsp = Object.assign(
-        new APIResponse(),
-        await apiHandler.handleRequest(req)
-      );
+      const rsp = Object.assign(new APIResponse(), await apiHandler.handleRequest(req));
       response.status(rsp.getStatus());
 
       if (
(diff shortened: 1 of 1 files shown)


tim-hellhake (approved):

benfrancis (approved):

benfrancis:
Thanks @bewee

Since this was reported on 1.0.X too, would you mind creating a PR which applies this same patch to that branch?

PR #2897, opened 2021-11-23
`writeproperty` operation - closes #2892

### LINKED ISSUE #2892: [Core Profile] Implement `writeproperty` operation
As a developer consuming the gateway API, I want to be able to write a property so that I can change the state of a device.

Implement the `writeproperty` protocol binding from the W3C WoT Profile Core Profile, as defined in https://w3c.github.io/wot-profile/#writeproperty

Since this is mostly already implemented, I think all that's actually needed here is to change the response code from `200 OK` to `204 No Content`

benfrancis:
This PR makes the `writeproperty` operation on the gateway's API compliant with the [W3C Core Profile protocol binding](https://w3c.github.io/wot-profile/#writeproperty), essentially by changing the HTTP response code form 200 to 204.

This is a bit awkward for a couple of reasons:
1. The front end expects an HTTP response data payload following the same format as the WebSocket API, which is no longer the case, so that payload has to be reconstructed for the `onPropertyStatus()` method
2. This is the only PUT request in the API which responds with an empty body, and TypeScript is strict about these things, so I needed to create a special API method specifically for PUT requests which don't return a body

### DIFF
--- a/src/controllers/things_controller.ts
+++ b/src/controllers/things_controller.ts
@@ -346,8 +346,8 @@ function build(): express.Router {
     const value = request.body;
     try {
       // Note: updatedValue may differ from value
-      const updatedValue = await Things.setThingProperty(thingId, propertyName, value);
-      response.status(200).json(updatedValue);
+      await Things.setThingProperty(thingId, propertyName, value);
+      response.sendStatus(204);
     } catch (err) {
       console.error('Error setting property:', err);
       response.status(err.code || 500).send(err.message);
--- a/src/test/browser/things-view/thing-test.ts
+++ b/src/test/browser/things-view/thing-test.ts
@@ -169,7 +169,8 @@ describe('Thing', () => {
     expect(stringId).toEqual(`string-${escapeHtmlForIdClass('spaced string')}`);
   });
 
-  it('should reset property value when setProperty is rejected', async () => {
+  // TODO: Fix or remove test case (https://github.com/WebThingsIO/gateway/issues/2906)
+  it.skip('should reset property value when setProperty is rejected', async () => {
     const browser = getBrowser();
     const desc = {
       id: 'UnknownThing',
--- a/static/js/api.ts
+++ b/static/js/api.ts
@@ -79,8 +79,24 @@ class API {
       if (!res.ok) {
         throw new Error(`${res.status}`);
       }
-
       return res.json();
+    });
+  }
+
+  putJsonWithEmptyResponse(url: string, data: Record<string, unknown>): Promise<void> {
+    const opts = {
+      method: 'PUT',
+      headers: {
+        Authorization: `Bearer ${this.jwt}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(data),
+    };
+
+    return fetch(url, opts).then((res) => {
+      if (!res.ok) {
+        throw new Error(`${res.status}`);
+      }
     });
   }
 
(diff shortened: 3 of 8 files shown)


benfrancis:
@relu91 This is basically ready for review, but there's still one browser test that's failing ([this one](https://github.com/WebThingsIO/gateway/blob/master/src/test/browser/things-view/thing-test.ts#L172)). This is the same test that we had a problem with in #2868.

@sogaani Do you have any thoughts on why this change causes this test to fail? You said in https://github.com/WebThingsIO/gateway/pull/2868#issuecomment-915311005 that the gateway no longer has "the logic to revert property value when setProperty failed", but this PR doesn't remove that logic. Do you have any idea why it passes before this PR but fails after?

If it's testing logic which no longer exists then I am tempted to remove the test, but I'd like to better understand what's going on.

benfrancis:
@relu91 Since I haven't heard back from @sogaani and it seems the broken test case is testing a feature that may longer exist, I have disabled the test for now and filed #2906 to fix or remove it.

That means this PR is ready for your review if you have the time. Thanks.

src/controllers/things_controller.ts
```diff
@@ -346,8 +346,8 @@ function build(): express.Router {
     const value = request.body;
     try {
       // Note: updatedValue may differ from value
```
relu91, line 348: Side note: with the new changes this comment seems a little bit obscure. Should we keep it?

src/test/browser/things-view/thing-test.ts
```diff
@@ -169,6 +169,9 @@ describe('Thing', () => {
     expect(stringId).toEqual(`string-${escapeHtmlForIdClass('spaced string')}`);
   });
 
+  /**
+  TODO: Fix or remove test case (https://github.com/WebThingsIO/gateway/issues/2906)
+
   it('should reset property value when setProperty is rejected', async () => {
```
relu91, line 175: instead of commenting if it does not have compilation errors I would `it.skip` it and leave the todo comment.

static/js/api.ts
```diff
@@ -79,11 +79,24 @@ class API {
...
       }
-
       return res.json();
     });
   }
 
+  putJsonWithEmptyResponse(url: string, data: Record<string, unknown>): Promise<void> {
+    const opts = {
+      method: 'PUT',
+      headers: this.headers('application/json'),
```
relu91, line 89: Minor note, in my understanding `this.headers` functions also set `Accept: 'application/json'` which is not needed in this case.

static/js/models/thing-model.js
```diff
@@ -202,10 +202,10 @@ class ThingModel extends Model {
       this.base
     );
 
-    return API.putJson(href, value)
-      .then((json) => {
+    return API.putJsonWithEmptyResponse(href, value)
+      .then(() => {
         const result = {};
-        result[name] = json;
+        result[name] = value;
```
relu91, line 208: I think it is worth adding a comment in the function documentation about the fact that the returned result is just for API backward compatibility and it might not reflect the actual remote property value.

relu91 (approved):
LGTM, I left some comments that if addressed would improve the resulting code a little bit.

src/controllers/things_controller.ts
relu91: Just noted that you actually removed this line in https://github.com/WebThingsIO/gateway/pull/2900/commits/1e76f8a1f9fb177eae7519bc076597772018bd5a#diff-b512196f53f6d1d852ec7d013d2070dce8f985f423299122a104947db6465551L348 .

benfrancis:
Thanks for the review, @relu91! I believe I have now addressed all your review comments, so merging.

PR #2900, opened 2021-12-08
`writemultipleproperties` operation - closes #2891

### LINKED ISSUE #2891: [Core Profile] Implement `writemultipleproperties` operation
As a developer consuming the gateway API, I want to be able to write multiple properties at the same time so that I don't need to make a separate request for each property.

Implement `writemultipleproperties` protocol binding from W3C WoT Profile Core Profile, as defined in https://w3c.github.io/wot-profile/#writemultipleproperties

benfrancis:
Implementation of the [`writemultipleproperties` operation](https://w3c.github.io/wot-profile/#writemultipleproperties) from WoT Core Profile.

Currently builds on #2897 so will need rebasing against master

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -84,6 +84,7 @@ export enum LogSeverity {
 
 export enum WoTOperation {
   READ_ALL_PROPERTIES = 'readallproperties',
+  WRITE_MULTIPLE_PROPERTIES = 'writemultipleproperties',
   SUBSCRIBE_ALL_EVENTS = 'subscribeallevents',
   UNSUBSCRIBE_ALL_EVENTS = 'unsubscribeallevents',
   QUERY_ALL_ACTIONS = 'queryallactions',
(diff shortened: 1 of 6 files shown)


benfrancis:
OK, I think this is ready for review but it is still built on #2897, so it's only the last commit (https://github.com/WebThingsIO/gateway/pull/2900/commits/1e76f8a1f9fb177eae7519bc076597772018bd5a) which requires review here.

One open question I'm not sure about (because the WoT Profile specification doesn't specify it) is what should happen if when setting multiple properties some succeed and some fail, i.e. is the operation atomic? Currently what will happen is that if some are set successfully and some are not they will be left that way and a 500 error will be returned.

Another option might be to try to roll back the successful property writes if one of the other writes fails, but that would be more complicated.

relu91 (approved):
LGTM, but I'm also concerned about how we are handling failures in `writemultipleproperties` operations, the current implementation will likely cause troubles. I suggest opening a tracking issue after this PR has been merged.

benfrancis:
Thanks, I have filed follow-up issues #2913 and https://github.com/w3c/wot-profile/issues/162.

PR #2907, opened 2022-01-12
Rename mediaType to type in links - closes #2808

### LINKED ISSUE #2808: Change `mediaType` to `type` for `links` in API
Blocks: #2802

Currently [Link](https://webthings.io/api/#link-object) objects in a Thing Description exposed by the gateway can define a `mediaType`, but in the W3C specification [it is called](https://www.w3.org/TR/wot-thing-description/#link) `type`. Let's change it to be compliant.

**This includes:**
1. Changing remaining links in Thing Descriptions exposed by the gateway to use `type` rather than `mediaType`
2. Updating the gateway's front end to expect this

See https://github.com/WebThingsIO/api/issues/125 for previous discussion

benfrancis:


relu91 (approved):
Well done, kudos for keeping backward compatibility.

PR #2911, opened 2022-01-20
`invokeaction` operation - closes #2893

### LINKED ISSUE #2893: [Core Profile] Implement `invokeaction` operation
As a developer consuming the gateway API, I want to be able to invoke an action so that I can perform some function on a device.

Implement the `invokeaction` protocol binding from the W3C WoT Profile Core Profile, as defined in https://w3c.github.io/wot-profile/#invokeaction

This was partially implemented in https://github.com/WebThingsIO/gateway/pull/2871 which resulted in WoT Thing Description compliance. To fully implement the `invokeaction` protocol binding from the Core Profile, the response payload needs to be updated to conform to the ActionStatus format defined in https://w3c.github.io/wot-profile/#async-action-response.

benfrancis:
Closes #2893.

This PR brings the protocol binding of the `invokeaction` operation in line with the [HTTP Basic Profile](https://w3c.github.io/wot-profile/#http-basic-profile) in the latest draft of the W3C WoT Profile specification.

Specifically:
- Aligns the values of the `status` member of the `ActionStatus` object, and extracts them out as constants
- Deprecates the existing use of `POST` on the top level `/actions` endpoint of a Thing, in favour of only allowing a `POST` on `/actions/{actionName}` (the functionality is left in place, but no longer used internally and a deprecation warning is printed on the console)
- Adds the URL of an action resource to the `Location` header of a response to an `invokeaction` request
- Removes object wrappers from various payloads
- Uses the Problem Details format for error messages used by the gateway's own actions
- Updates integration tests to reflect the changes

TODO:
- [x] Implement back end
- [x] Check if front end needs updating
- [x] Extract constants for status member of ActionStatus
- [x] Update format of error member of ActionStatus
- [x] Update integration tests

### DIFF
--- a/src/models/action.ts
+++ b/src/models/action.ts
@@ -16,12 +16,11 @@ import { ActionDescription as AddonActionDescription, Any } from 'gateway-addon/
 import Thing from './thing';
 
 export interface ActionDescription {
-  input: Any;
   href: string;
   status: string;
   timeRequested: string;
   timeCompleted?: string;
-  error?: string;
+  error?: Record<string, unknown>;
 }
 
 export default class Action extends EventEmitter {
@@ -41,7 +40,7 @@ export default class Action extends EventEmitter {
 
   private timeCompleted: string | null;
 
-  private error: string;
+  private error: Record<string, unknown> | null;
 
   /**
    * Create a new Action
@@ -62,15 +61,14 @@ export default class Action extends EventEmitter {
       this.href = `${Constants.ACTIONS_PATH}/${name}/${this.id}`;
       this.thingId = null;
     }
-    this.status = 'created';
+    this.status = Constants.ActionStatusValues.PENDING;
     this.timeRequested = Utils.timestamp();
     this.timeCompleted = null;
-    this.error = '';
+    this.error = null;
   }
 
   getDescription(): ActionDescription {
     const description: ActionDescription = {
-      input: this.input,
       href: this.href,
       status: this.status,
       timeRequested: this.timeRequested,
@@ -96,7 +94,7 @@ export default class Action extends EventEmitter {
       return;
     }
 
-    if (newStatus === 'completed') {
+    if (newStatus === Constants.ActionStatusValues.COMPLETED) {
       this.timeCompleted = Utils.timestamp();
     }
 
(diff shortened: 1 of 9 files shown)


src/models/action.ts
```diff
@@ -16,12 +16,11 @@ import { ActionDescription as AddonActionDescription, Any } from 'gateway-addon/
 import Thing from './thing';
 
 export interface ActionDescription {
-  input: Any;
   href: string;
   status: string;
   timeRequested: string;
   timeCompleted?: string;
-  error?: string;
+  error?: Record<string, unknown>;
```
relu91, line 23: Have you considered making a Problem Details interface? I'm reading from the profile spec that also other methods MAY explain error codes using that format. It might be nice to have such an interface handy in the future. (here is [zion's Problem Details](https://github.com/vaimee/zion/blob/main/src/common/interfaces/problem-details.ts) interface as a reference)

static/js/views/add-thing.js
```diff
@@ -93,7 +93,7 @@ const AddThingScreen = {
 
     API.startPairing(timeout)
       .then((json) => {
-        this.actionUrl = json.pair.href;
+        this.actionUrl = json.href;
```
relu91, line 96: does this take into account the `base` field?

relu91 (approved):
I think the PR achieved the declared goals above. I have just added two minor comments that are mostly suggestions/questions rather than blockers.

src/models/action.ts
benfrancis: Good idea, I've done that.

static/js/views/add-thing.js
benfrancis: That's a good question, but I don't think it actually matters in the case of the gateway's front end because it's not a general purpose WoT Consumer, it only consumes Thing Descriptions exposed by the gateway. All Thing Descriptions exposed by the gateway have the [base set to the root of the host](https://github.com/WebThingsIO/gateway/blob/master/src/models/thing.ts#L638) and all values of `href` are absolute paths. Therefore the action URL will always be an absolute path, relative to the gateway's own host.

The WebThings Cloud implementation I'm working on next _is_ a general purpose WoT Consumer, and so will have to resolve the base URL properly.

PR #2932, opened 2022-05-25
Removes Chrome complaints, and everything still seems to work fine

flatsiedatsie:


benfrancis (approved):
Thanks for this. (It's a bit strange because [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#passive) `passive` already defaults to true for `touchstart` events in Chrome, but if it works it works!).

PR #2950, opened 2022-09-20
Security fixes

benfrancis:
These are security fixes which can be applied automatically using `npm audit fix`.

There are further security vulnerabilities reported which would require manual intervention.


rzr:
LGTM

benfrancis:
I've set this as draft because it will probably need to be re-done once https://github.com/WebThingsIO/gateway/pull/2951 has landed.

PR #2952, opened 2022-09-27
1.0.1 security fixes

benfrancis:
This PR adds security fixes from `npm audit fix`, then fixes the fallout of all of those upgrades in order to make the TypeScript compiler and linter happy and fix test regressions.

This turned out to be a bit of a mission and I regret that I couldn't do this on master first and then backport (so we wouldn't have to rebase all of this work for master in the future), but the latest package-lock.json on master was generated with npm7 so uses version 2 of the package lock file format, which complicates things. Patching the 1.0.X branch seems like the quickest path to getting out a 1.0.1 release.

With this PR landed there are only 6 remaining vulnerabilities reported by `npm audit`, which all either require breaking changes or manual interventions, but they are all in image libraries which as far as I can tell are only used by webpack at build time so aren't a big problem.

Let me know if you have any questions!

It might be easier to review the commits separately, but they need to land together in order to not cause regressions.


tim-hellhake (approved):

tim-hellhake:
Looks good!

PR #2955, opened 2022-09-27
Master security fixes

benfrancis:
This is the counterpart to https://github.com/WebThingsIO/gateway/pull/2952 but re-based for the master branch.

~~Still WIP because it needs further testing and re-basing on top of #2954 once that lands.~~


tim-hellhake (approved):

benfrancis:
There's one failing browser test on CI which appears to be unrelated.

Browser tests continue to be flaky and I get a different error each time it runs. I've manually tested the functionality it is meant to be testing and it's working as expected so this does seem to be a bug with the test rather than the actual code.

I'm going to land this patch because the security fixes are important and dependabot is freaking out, but we are going to need to figure out how to fix these tests.

benfrancis:
See https://github.com/WebThingsIO/gateway/issues/2866#issuecomment-1265321086 regarding the known intermittent test failure.

PR #2970, opened 2022-10-03
feat(CI): upgrade checkout and setup-node actions of build workflow

josephyi:
Successful run - https://github.com/josephyi/gateway/actions/runs/3171386118

I attempted to enable the caching feature, but I think it doesn't like the matrix build. 🤷

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,8 +19,8 @@ jobs:
           14,
         ]
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


benfrancis:
@josephyi Thank you for this contribution.

I'm not familiar with this workflow, so out of curiosity would you mind briefly explaining the motivation for the upgrade and what it's intended to fix so I know what to look out for?

josephyi:
> @josephyi Thank you for this contribution.
> 
> I'm not familiar with this workflow, so out of curiosity would you mind briefly explaining the motivation for the upgrade and what it's intended to fix so I know what to look out for?

The motivation for the upgrade is to treat actions like dependencies that should be kept up to date for bug fixes, performance improvements, new features, etc. Functionally, everything should behave the same.

PR #2978, opened 2022-10-04
Use latest Buster as base for 1.0.1 Raspberry Pi image

### LINKED ISSUE #2981: Need to tell Raspberry Pi image build script to use buster rather than bullseye
Using Raspberry Pi OS bullseye as a base distribution for the Raspberry Pi image will probably cause the OTA update not to work for exiting gateways.

Given buster is still being supported, we should use that as a base for the 1.0.1 release.

benfrancis:
This PR uses the latest Raspberry Pi OS Buster as a base for the Raspberry Pi image, instead of the latest master from pi-gen.

Fixes #2981

### DIFF
--- a/image/build.sh
+++ b/image/build.sh
@@ -13,8 +13,8 @@ compgen -G "node_modules*" >/dev/null && rm -rf node_modules* || true
 _temp=$(mktemp -d)
 cp -a "${here}/.." "${_temp}"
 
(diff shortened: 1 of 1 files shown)


benfrancis:
@tim-hellhake It would be great if you could review this for me, thanks.

tim-hellhake (approved):

PR #2990, opened 2022-10-23
Fix various bugs concerning groups UI

### LINKED ISSUE #2940: Dragging the last thing slightly to the left makes it disappear?
drag-removes

This also happens to things inside groups.

A refresh of the page restores everything, and the things are then in a reshuffled position.

This is on a Candle controller, which means its Gateway version 1.1

bewee:
This fixes:
- The first/last thing of a group / of the ungrouped things disappearing when being dropped on itself (Closes #2940)
- Overlapping things and groups when actually scrolling should take place. As a part of this, groups are fixed-width now.

### DIFF
--- a/static/css/things.css
+++ b/static/css/things.css
@@ -11,32 +11,25 @@
 
 #things-container {
   position: relative;
-  top: 50%;
-  transform: translateY(-50%);
-  max-height: 100%;
-  display: flex;
-  flex-direction: column;
+  margin: auto;
+  width: calc(100% - 10rem);
 }
 
 #things {
   text-align: center;
   font-size: 1.6rem;
   color: #fff;
-  flex-direction: column;
-  margin: auto;
-  padding-left: 5px;
-  padding-right: 5px;
   min-height: 50px;
+  margin: auto;
   width: calc(100% - 10rem);
 }
 
 #groups {
   text-align: center;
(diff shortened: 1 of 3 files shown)


benfrancis:
Great, thanks for this! I'll take a look as soon as we get 1.0.1 out of the door.

static/css/things.css
```diff
@@ -11,32 +11,25 @@
 
 #things-container {
   position: relative;
-  top: 50%;
-  transform: translateY(-50%);
-  max-height: 100%;
-  display: flex;
-  flex-direction: column;
+  margin: auto;
+  width: calc(100% - 10rem);
```
benfrancis, line 15: This breaks the vertical alignment of Things, and the "Click + to add Thing" hint in the empty state.

You could use flexbox for the vertical aligment instead of the old transform approach:

```suggestion
  display: flex;
  flex-direction: column;
  align-items: center;
  justify-content: center;
```

benfrancis (changes requested):
Thanks for fixing these bugs, unfortunately the fix breaks vertical alignment which is a part of the UI design. I've proposed an alternative which I think should work, see what you think.

benfrancis:
Closing in favour of #3099, thanks for your work on this.

PR #2991, opened 2022-10-23
Refine groups

### LINKED ISSUE #2938: Refine Groups UI
I'd like to suggest some refinements to the Groups UI before we ship 1.1.

groups

Key changes:
1. Add groups via an overflow menu on the add button, rather than via the Add Thing screen. Because a) Adding a group feels like a different action to adding a thing so it feels out of place in the add thing screen and b) Clicking the + button currently does a lot of heavy lifting behind the scenes (like putting all the gateway's radios in pairing mode) which isn't necessary to just add a group
2. Tweaks to the add thing, edit thing and remove thing screens to make them consistent with similar screens
3. A bit of polish on the group UI like updated icons and layout (e.g. full width rather than dynamic width)

(I'm not sure yet about the proposed new thing and group icons)

I've also noticed some bugs with the existing implementation that we need to make sure are fixed before release. I've included a couple of examples of the UI in a broken state, below.

Screenshot from 2022-07-07 14-45-47
Screenshot from 2022-07-07 16-22-49

Hopefully they can be fixed at the same time.

CC @bewee

bewee:
Closes #2938

- [x] Refine layout of edit group dialog
- [x] Move "new group" into own dialog
- [ ] Add/update thing and group icons


flatsiedatsie:
check for "Gruop"

PR #3000, opened 2022-11-10
Use docker hub password and add manual trigger

tim-hellhake:
@benfrancis
I assume that `DOCKER_HUB_REAL_PASSWORD` was the actual docker hub user password.
Maybe this was necessary to update the description.
We can use the existing access token for this.
I also added a manual trigger for the workflow.


benfrancis (approved):

benfrancis:
Thanks!

benfrancis:
That seemed to work, the description on DockerHub has been updated :)

FYI, after merging this commit I had to update my personal access token to include the workflow permission before it would let me push branches to GitHub.

PR #3011, opened 2022-11-17
Add a pre-release workflow - fixes #2987

### LINKED ISSUE #2987: Create a workflow for pre-release tags
This is a follow-up to #2980 in which we discovered that tagging a release automatically publishes a new image to DockerHub.

Currently when a new tag is created on the repo which looks like a semver version tag, a GitHub [workflow](https://github.com/WebThingsIO/gateway/blob/master/.github/workflows/release.yml) creates a new release and tries to build a Raspberry Pi image, apt and rpm packages, and a docker image.

The release is initially set as a pre-release on GitHub and production gateways will not pull down an OTA update until the pre-release flag is removed. However, the current release workflow will automatically try to publish the docker image to DockerHub, even if the pre-release flag is set. This may result in an un-tested image being pushed to DockerHub.

As far as I can tell it isn't possible to create a workflow which only pushes an image to DockerHub when the pre-release flag is removed, but what we could do is set up a separate workflow which is triggered when a tag ending in "-pre" is added. This workflow could do everything the current release workflow does, but without publishing anything to DockerHub.

benfrancis:
This is a proposed workflow for pre-release images, which fixes #2987.

To use this workflow you create a semver style tag with a suffix of `-alpha` or `-beta` followed by a dot and then a number. E.g. `1.0.1-alpha.1` or `1.0.1-beta.2`.

The only differences between this workflow and the "release" workflow are:
1. It's only triggered by tags with an alpha or beta pre-release suffix as described above, and not by production release tags like `1.0.1`
2. It still pushes a built image to DockerHub (with the pre-release suffix in its name), but it doesn't reset the `latest` tag to the pre-release image

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,9 +1,10 @@
-name: Build and push release images
+name: Build and push pre-release images
 
 on:
   push:
     tags:
-    - '[0-9]+.[0-9]+.[0-9]+'
+    - '[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+'
+    - '[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+'
 
 jobs:
   create-release:
@@ -39,7 +40,7 @@ jobs:
           publish: true
           imageName: webthingsio/gateway
           platform: linux/amd64,linux/arm/v7,linux/arm64
-          tag: latest,${{ env.RELEASE_VERSION }}
+          tag: ${{ env.RELEASE_VERSION }}
           dockerUser: ${{ secrets.DOCKER_HUB_USER }}
           dockerPassword: ${{ secrets.DOCKER_HUB_PASSWORD }}
 


benfrancis:
One thing I don't know is whether gateways will pull down pre-release OTA updates with a version number in this format, and how (or if) they will compare them against production versions.

Note that I've left the "pre-release" option on GitHub releases in the "release" workflow as an extra safety step, to make sure we don't accidentally push OTA updates out to gateways with an untested pre-release image just by using the wrong tag format.

tim-hellhake (approved):

tim-hellhake:
> One thing I don't know is whether gateways which are set to download pre-release updates will pull down pre-release OTA updates with a version number in this format, and how (or if) they will compare them against production versions.

The gateway will consider whatever comes first on the list
https://github.com/WebThingsIO/gateway/blob/300bd36067b423abc63751dad1feed51d0e6738a/tools/check-for-update.js#L29
https://github.com/WebThingsIO/gateway/blob/300bd36067b423abc63751dad1feed51d0e6738a/tools/check-for-update.js#L38
and will compare it using the [semver](https://www.npmjs.com/package/semver) lib with the current release
https://github.com/WebThingsIO/gateway/blob/300bd36067b423abc63751dad1feed51d0e6738a/tools/check-for-update.js#L47

> Note that I've left the "pre-release" flag on GitHub releases created by the "release" workflow as an extra safety step, to make sure we don't accidentally push OTA updates out to gateways with an untested pre-release image just by using the wrong tag format.

Sounds like a good idea.

Issue #3007, opened 2022-11-17
Fix or disable intermittently failing browser tests


benfrancis:
On almost every PR at least one of the test runs in CI fails on some intermittently failing browser tests. This makes it look like the build is always broken, but it's almost certainly the tests that are broken.

We need to fix these tests, but may need to disable them in the meantime.

```
2022-11-12T12:27:46.4970418Z Summary of all failing tests
2022-11-12T12:27:46.4971210Z FAIL test/browser/things-view/thing-test.js (138.944 s)
2022-11-12T12:27:46.4972082Z   ● Thing › should render an unknown thing and be able to change properties
2022-11-12T12:27:46.4972411Z 
2022-11-12T12:27:46.4972715Z     val.split(...).flatMap(...) is not a function or its return value is not iterable
2022-11-12T12:27:46.4973239Z     TypeError: val.split(...).flatMap(...) is not a function or its return value is not iterable
2022-11-12T12:27:46.4973545Z 
2022-11-12T12:27:46.4973878Z       at typeInput.reduce (../node_modules/devtools/build/commands/elementSendKeys.js:38:33)
2022-11-12T12:27:46.4974298Z           at Array.reduce (<anonymous>)
2022-11-12T12:27:46.4974822Z       at DevToolsDriver.elementSendKeys (../node_modules/devtools/build/commands/elementSendKeys.js:36:31)
2022-11-12T12:27:46.4975150Z 
2022-11-12T12:27:46.4975548Z   ● Thing › should render an unknown thing and be able to change properties
2022-11-12T12:27:46.4975829Z 
2022-11-12T12:27:46.4976453Z     waitUntil condition failed with the following reason: Can't call getCSSProperty on element with selector "#menu-scrim.hidden" because element wasn't found
2022-11-12T12:27:46.4976829Z 
2022-11-12T12:27:46.4977163Z       at timer.catch (../node_modules/webdriverio/build/commands/browser/waitUntil.js:70:15)
2022-11-12T12:27:46.4977451Z 
2022-11-12T12:27:46.4977790Z   ● Thing › should render a thing with spaced property names
2022-11-12T12:27:46.4978048Z 
2022-11-12T12:27:46.4978462Z     val.split(...).flatMap(...) is not a function or its return value is not iterable
2022-11-12T12:27:46.4979025Z     TypeError: val.split(...).flatMap(...) is not a function or its return value is not iterable
2022-11-12T12:27:46.4979443Z 
[...]

PR #3014, opened 2022-11-18
Disable intermittently failing browser tests

benfrancis:
This PR proposes disabling some intermittently failing browser tests which are failing about 1/3rd of the time, as discussed in #3007.

Obviously it would be better to fix these tests, but they have been broken for many months. I would prefer to have a green test run which excludes these known broken tests so that we can more easily spot real problems.

### DIFF
--- a/src/test/browser/load-test.ts
+++ b/src/test/browser/load-test.ts
@@ -13,7 +13,7 @@ describe('basic browser tests', () => {
     }
   });
 
-  it('creates a user', async () => {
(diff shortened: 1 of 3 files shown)


benfrancis:
Waiting to see if we get a green test run with these tests disabled.

benfrancis:
Well, that didn't go well.

I have no idea why disabling those tests is causing new timeouts.

benfrancis:
Superseded by #3100.

PR #3034, opened 2022-12-08
Standardise `@context` of Thing Descriptions - closes #2809

### LINKED ISSUE #2809: Change `@context` of Thing Descriptions
Blocks: #2802

Once the Thing Descriptions exposed by the gateway are W3C compliant, we should update the `@context` member to include the string `https://www.w3.org/2019/wot/td/v1` to reflect that.

**Open questions:**
1. We still want to be able to provide `https://webthings.io/schemas/` as a context in order to add semantic annotations from the WebThings capability schemas. Does this mean we'll need to add a prefix to all of those annotations? e.g.

```json
{
  "@context": ["https://www.w3.org/2019/wot/td/v1", {"wt": "https://webthings.io/schemas/"} ],
  "@type": ["wt:Light", "wt:OnOffSwitch"],
  "id": "https://mywebthingserver.com/things/lamp",
  "title": "My Lamp",
  "description": "A web connected lamp",
  "properties": {
    "on": {
      "@type": "wt:OnOffProperty",
    },
    "brightness" : {
      "@type": "wt:BrightnessProperty",
    }
  },
  "actions": {
    "fade": {
      "@type": "wt:FadeAction"
    }
  },
  "events": {
    "overheated": {
      "@type": "wt:OverheatedEvent"
    }
  }
}
```

Note: This should only be done once we've made all the other changes to make Thing Descriptions W3C compliant.

See https://github.com/WebThingsIO/api/issues/157 for some past discussion

benfrancis:
This PR makes sure that all Thing Descriptions exposed by the gateway have both the W3C WoT TD 1.1 and WebThings JSON-LD contexts, in order to work towards W3C compliance. It:

1. Makes sure the `@context` is an array
2. Removes the legacy `iot.mozilla.org/schemas` context if used
3. Makes sure both `https://www.w3.org/2022/wot/td/v1.1` and `https://webthings.io/schemas` are present in the array, and if not adds the W3C context to the start and/or adds the WebThings context to the end of the array. E.g.

```json
"@context": [
  "https://www.w3.org/2022/wot/td/v1.1",
  "https://webthings.io/schemas",
]
```

I had written some code to migrate existing Thing Descriptions in the gateway's database, so that they don't have to be modified at runtime every time they are read, but it turned out that code was redundant. Those Thing Descriptions already get updated on startup as a side effect of the `updateFromDescription()` method being called for every existing device when handling `deviceAdded` messages from adapters. I don't think this is necessarily desirable, so I plan to revisit this in a follow-up and figure out what's going on there.

Note: Currently the gateway only supports an array of strings in the `@context`, not maps which define JSON-LD prefixes. That feature is planned for later, when we'll ideally also [add prefixes to all non-standard members](https://github.com/WebThingsIO/gateway/issues/1919).

TODO:
- [x] Remove legacy context
- [x] Add W3C context
- [x] Add WebThings context
- [x] ~~Migrate old Thing Descriptions~~
- [x] Fix integration tests

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -11,6 +11,7 @@ import fs from 'fs';
 import * as Platform from './platform';
 import pkg from './package.json';
 import { Event as EventSchema } from 'gateway-addon/lib/schema';
+import * as Constants from './constants';
 
 /**
  * Compute a SHA-256 checksum of a file.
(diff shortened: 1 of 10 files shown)


benfrancis:
Note that we should also update the default used by [gateway-addon-node](https://github.com/WebThingsIO/gateway-addon-node/issues/121) and [gateway-addon-python](https://github.com/WebThingsIO/gateway-addon-python/issues/93).

src/utils.ts
```diff
@@ -74,17 +75,59 @@ export function isW3CThingDescription(thingDescription: {
...
+ * @returns string[] The standardised context array or null if already OK.
+ */
+export function standardizeContext(context: string | string[]): string[] | null {
+  // If expected contexts already present then return null
+  if (
+    Array.isArray(context) &&
+    context.includes(Constants.WOT_TD_1_1_CONTEXT) &&
+    context.includes(Constants.WEBTHINGS_CONTEXT)
+  ) {
+    return null;
```
relu91, line 109: Why return `null`? can't we just return the context as it is? it should simplify https://github.com/WebThingsIO/gateway/pull/3034/files#diff-298926bb41b428e204128605377a50dec34c4e90cc601ac06bd172563a91a399R717 . Given the fact that the contract of the function is to produce a standardized context, if it is already standard it makes sense to return it without touching.

relu91 (approved):
The changes look good, I have a minor question about the business logic of the `standardizeContext` function.

src/utils.ts
benfrancis: Good point! Fixed.

PR #3040, opened 2022-12-13
Allow no exec member for extension add-ons (v1.1) - closes #3037

### LINKED ISSUE #3037: Make `exec` member of add-on manifests optional for extensions again
In the TypeScript re-write for 1.1 a change was made that made the exec member of add-on manifests mandatory.

This means it's no longer possible to create an extension add-on which is just CSS and JavaScript content with no back end executable.

We should consider making exec optional again to allow extension add-ons (e.g the square theme).

benfrancis:


src/addon-utils.ts
```diff
@@ -82,13 +82,12 @@ function validateObject(
  * @returns {string|null} Error string, or null if no error.
  */
 function validateManifestJson(manifest: Record<string, unknown>): string | null {
-  const manifestTemplate = {
+  const manifestTemplate: Record<string, any> = {
```
tim-hellhake, line 85:
```suggestion
  const manifestTemplate: Record<string, unknown> = {
```
This would make the linter happy

tim-hellhake (changes requested):

tim-hellhake:
> Hmm, that didn't make it happy.

The linter is happy now.
Now it's a compile error.
The `Record` type is the problem.
Before the object type was inferred through the data.
The compiler can no longer guarantee that the properties exist.

> What is the reason for the "no explicit any" rule anyway? It's kind of a pain.

It defeats the purpose of a typesafe language.

src/addon-utils.ts
```diff
@@ -82,13 +82,12 @@ function validateObject(
  * @returns {string|null} Error string, or null if no error.
  */
 function validateManifestJson(manifest: Record<string, unknown>): string | null {
-  const manifestTemplate = {
+  const manifestTemplate: Record<string, unknown> = {
```
tim-hellhake, line 85:
```suggestion
  const manifestTemplate = {
```

tim-hellhake (changes requested):
Unfortunately, I cannot edit the correct line because it's outside the diff.
But if you want a quick solution it would be
```
    (<Record<string, unknown>>(
      (<Record<string, unknown>>manifest.gateway_specific_settings).webthings
    )).exec = '';
```
in line 125.

tim-hellhake:
> What is the reason for the "no explicit any" rule anyway? It's kind of a pain.

Yes, it's painful but for a reason.
It's a potentially unsafe operation.
The correct way would be to define an interface for the manifest json.
This allows you to specify optional properties.
Or even better: use json schema instead of writing your own validator :grimacing:

tim-hellhake (approved):

benfrancis:
> if you want a quick solution it would be
> 
> ```
>     (<Record<string, unknown>>(
>       (<Record<string, unknown>>manifest.gateway_specific_settings).webthings
>     )).exec = '';
> ```
> 
> in line 125.

Thank you! Done.

> Yes, it's painful but for a reason.
> It's a potentially unsafe operation.
> The correct way would be to define an interface for the manifest json.
> This allows you to specify optional properties.

Makes sense, thank you.

Ironic that this rule is probably what caused this bug in the first place... ;)

> Or even better: use json schema instead of writing your own validator grimacing

Yeah it's weird that all the add-on IPC messages are validated using JSON Schema but the add-on manifest is validated like this. I don't know what the history of that is.

flatsiedatsie:
Awesome, cool that the Square theme will continue to work.

PR #3047, opened 2022-12-15
Screw you TypeScript (1.1) - closes #3046

### LINKED ISSUE #3045: ERROR  : Failed to get property on: getProperty: device: zb-000d6f000400f652
STR:
- Install 1.0 and pair a Zigbee smart plug
- Verify that smart plug can be turned on and off
- Upgrade to 1.1.0-beta.1 pre-release

Expected:
- Able to read and write "on" property of smart plug

Actual:
- Web thing icon is greyed out in UI
- Error in syslog:

`Dec 15 11:36:56 gateway run-app.sh[3219]: #033[31m2022-12-15 11:36:56.492 ERROR  : Failed to get property on: getProperty: device: zb-000d6f000400f652 not found.#033[39m`

### LINKED ISSUE #3046: Unable to add any new Things in 1.1 beta
STR:
- Upgrade to 1.1.0-beta.1
- Install Virtual Things adapter add-on and DateTime adapter add-on
- Click "+" on the things screen

Expected:
- New devices appear

Actual:
- No devices appear

benfrancis:
https://github.com/WebThingsIO/gateway/pull/3040 appears to have caused a regression (#3046) which breaks all adapter add-ons and prevents users from adding or using any things.

Here is a version of that patch which is not completely typesafe, but has the benefit that it works.

closes #3046 and closes #3045

### DIFF
--- a/src/addon-utils.ts
+++ b/src/addon-utils.ts
@@ -235,21 +235,26 @@ function loadManifestJson(packageId: string): [Record<string, unknown>, Record<s
   let min = manifest.gateway_specific_settings.webthings.strict_min_version;
   let max = manifest.gateway_specific_settings.webthings.strict_max_version;
 
+  const gatewayVersion = semver.coerce(pkg.version);
+  if (gatewayVersion === null) {
+    throw new Error(`Unable to compare with non-semver gateway version ${pkg.version}`);
+  }
+
   if (typeof min === 'string' && min !== '*') {
     min = semver.coerce(min);
-    if (semver.lt(pkg.version, min)) {
+    if (semver.lt(gatewayVersion, min)) {
       throw new Error(
         // eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


benfrancis:
The second commit fixes integration tests by ignoring pre-release strings when comparing gateway versions.

Previously addon-utils was stripping the pre-release string from the version provided in the add-on manifest, but not the version provided in package.json, which lead to identical versions being interpreted as being different.

tim-hellhake (approved):

tim-hellhake:
> #3040 appears to have caused a regression (#3046) which breaks all adapter add-ons and prevents users from adding or using any things.
> 
> Here is a version of that patch which is not completely typesafe, but has the benefit that it works.
> 
> closes #3046 and closes #3045

I'm very curious why that makes a difference.
The transpiled code should be exactly the same.

Issue #3065, opened 2023-01-15
Error when starting docker image in Raspberry Pi 3 (Raspbian)


carlosanunes:
When starting the docker image I get the following error message:

TypeError: Cannot read property 'on' of undefined
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/things.js:377:25)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/actions.js:40:34)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
Process 10 died: No such process; trying to remove PID file. (/run/avahi-daemon//pid)
+ WEBTHINGS_HOME=/home/node/.webthings
+ args=
+ start_task=run-only
+ is_container
+ '[' -f /.dockerenv ']'
+ return 0
++ node --version
++ egrep -o '[0-9]+'
++ head -n1
+ _node_version=12
+ [[ ! -f /home/node/.webthings/.node_version ]]
+ cd /home/node/webthings/gateway
+ mkdir -p /home/node/.webthings/config
+ ./tools/update-addons.sh
Opening database: /home/node/.webthings/config/db.sqlite3
/home/node/webthings/gateway/build/models/things.js:377
addon_manager_1.default.on(Constants.THING_ADDED, (thing) => {
                        ^

TypeError: Cannot read property 'on' of undefined
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/things.js:377:25)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/actions.js:40:34)
[...]

Issue #3043, opened 2022-12-14
Raspberry pi 4 bullseye not running docker container.


werto165:
I am trying to run the follow image: webthingsio/gateway:1.1.0-alpha.1 on bullseye
```
+ WEBTHINGS_HOME=/home/node/.webthings
+ args=
+ start_task=run-only
+ is_container
+ '[' -f /.dockerenv ']'
+ return 0
++ node --version
++ egrep -o '[0-9]+'
++ head -n1
+ _node_version=12
+ [[ ! -f /home/node/.webthings/.node_version ]]
+ cd /home/node/webthings/gateway
+ mkdir -p /home/node/.webthings/config
+ ./tools/update-addons.sh
Creating database: /home/node/.webthings/config/db.sqlite3
/home/node/webthings/gateway/build/models/things.js:377
addon_manager_1.default.on(Constants.THING_ADDED, (thing) => {
                        ^

TypeError: Cannot read property 'on' of undefined
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/things.js:377:25)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/node/webthings/gateway/build/models/actions.js:40:34)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
Process 10 died: No such process; trying to remove PID file. (/run/avahi-daemon//pid)
+ WEBTHINGS_HOME=/home/node/.webthings
+ args=
+ start_task=run-only
+ is_container
+ '[' -f /.dockerenv ']'
+ return 0
++ node --version
++ egrep -o '[0-9]+'
++ head -n1
+ _node_version=12
+ [[ ! -f /home/node/.webthings/.node_version ]]
+ cd /home/node/webthings/gateway
+ mkdir -p /home/node/.webthings/config
+ ./tools/update-addons.sh
Opening database: /home/node/.webthings/config/db.sqlite3
/home/node/webthings/gateway/build/models/things.js:377
[...]

PR #3071, opened 2023-02-11
Deactivate broken update script to fix docker containers

### LINKED ISSUE #3066: Webthings 1.1.0 Docker image new install failing to start
Having had problems with the 1.1.0 docker image on a clean install not starting so did a bit of digging and worked out that putting a text file in the /home/node/.webthings folder named .node_version and ensure it's content is 12 will result in the image starting properly. If the file is missing the image will not start.

tim-hellhake:
Fixes #3066 #3065 #3043

The update-addons.sh does two things:
- Run the [migrate](https://github.com/WebThingsIO/gateway/blob/master/src/migrate.ts) script
- Call the [updateAddons](https://github.com/WebThingsIO/gateway/blob/c378371d9f0dc5796287d469b8d67b5a6b56df07/src/addon-manager.ts#L1209) method

I'm not sure if we even still need the migrate script and the `updateAddons` method is called after the start anyways.

### DIFF
--- a/run-app.sh
+++ b/run-app.sh
@@ -55,7 +55,7 @@ else
         "$(< "${WEBTHINGS_HOME}/.node_version")" != "${_node_version}" ]]; then
     cd "${HOME}/webthings/gateway"
     mkdir -p "${WEBTHINGS_HOME}/config"
-    ./tools/update-addons.sh
+    # ./tools/update-addons.sh
     cd -
     echo "${_node_version}" > "${WEBTHINGS_HOME}/.node_version"
   fi


benfrancis:
Hmm, this would be a quick fix but I would ideally like to fix the underlying problem rather than just comment out this line and leave dead code behind. Do you have any idea why the script is failing?

I plan to use migrate.ts to migrate Thing Descriptions to the WoT Thing Description 1.1 format. I had actually written code to do that but I haven't submitted it yet because of a strange issue with `updateFromDescription()` that needs fixing first, see https://github.com/WebThingsIO/gateway/pull/3034

tim-hellhake:
> Thanks for looking into this. This would be a quick fix but I would ideally like to fix the underlying problem rather than just comment out this line and leave dead code behind. Do you have any idea why the script is failing?

I tried to fix it, but after some time, I at least wanted to provide a quick fix so we could release a working docker image.

> I plan to use migrate.ts to migrate Thing Descriptions to the WoT Thing Description 1.1 format. I had actually written code to do that but I haven't submitted it yet because of a strange issue with `updateFromDescription()` that needs fixing first, see #3034

I see.

PR #3072, opened 2023-02-16
Bugfix for NaN issue

flatsiedatsie:
- select small part of log manually
- select 'week' from the dropdown

In some cases this led to an error. This fixes the `NaN` issues from the int becoming a string.


benfrancis:
Thanks! As I understand it HTML attribute values are always read as strings, so presumably it needs casting to an integer on line 209 as well?

flatsiedatsie:
Yeah that makes sense actually. Thanks!

benfrancis (approved):

Issue #3079, opened 2023-03-21
static Ethernet settings not shown


adioanca-rpma:
https://github.com/WebThingsIO/gateway/blob/4dadba9a11ec2cadd30353eb9b227bccd5b2b655/static/js/views/settings.js#L876

Issues:
1. `ipdaddr` instead of `ipaddr`
2. `.options` is missing. should be `body.options.[ipaddr | netmask | gateway]`
3. default values for `ipaddr| netmask| gateway` should be `''` and `this.elements.network.client.ethernet.netmask.value = '255.255.255.0';` line be removed IMO.

also the elements are hidden so my ugly workaround was to add:
```javascript
    this.ethernetEls = ethernetEls;
```
in `setupNetworkElements` and
```javascript
      if (body.mode === 'static') {
        this.ethernetEls.forEach((el) => el.classList.remove('hidden'));
      }
```
in `showEthernetSettings` which now looks like this:
```javascript
  showEthernetSettings: function () {
    this.backButton.href = '/settings/network';
    this.view.classList.add('dark');
    this.hideNetworkElements();
    this.elements.network.main.classList.remove('hidden');

    this.elements.network.client.ethernet.main.classList.remove('hidden');

    API.getLanSettings().then((body) => {
      body = {mode: body.mode, ...body.options}; // <--- ugly but small change
      if (body.mode === 'static') {
        this.ethernetEls.forEach((el) => el.classList.remove('hidden'));
      }
      this.elements.network.client.ethernet.mode.value = body.mode || 'dhcp';
      this.elements.network.client.ethernet.ip.value = body.ipaddr || '';
      this.elements.network.client.ethernet.netmask.value = body.netmask || '';
      this.elements.network.client.ethernet.gateway.value = body.gateway || '';
      this.titlebar.setAttribute("data-l10n-id", "ethernet");
    }).catch((e) => {
      console.error(`Failed to get ethernet config: ${e}`);
    }).then(() => {
      this.validateEthernet();
    });
  },
```

adioanca-rpma:
Done.

@benfrancis @tim-hellhake could you please help with NodeJS 16 build tools docker image?
alongside these:
`webthingsio/toolchain-linux-[ARCH]-node-[8|10|12|14]`

Thank you

adioanca-rpma:
[...]

PR #3080, opened 2023-03-22
Fix for #3079 - static Ethernet settings not shown

adioanca-rpma:


static/js/views/settings.js
```diff
@@ -243,6 +243,8 @@ const SettingsScreen = {
       this.elements.network.client.ethernet.gatewayLabel,
       this.elements.network.client.ethernet.gateway,
     ];
+    // we'll use these in /settings/network/ethernet screen if 'static' IP is configured
+    this.ethernetEls = ethernetEls;
```
benfrancis, line 247: I haven't edited this part of the code before, but I note that the Settings class maintains a hierarchical map of all of the HTML elements based on their section within the Settings UI. This line creates a set of duplicate references outside of that `this.elements` map.

If this subset of elements needs sharing between methods, would it instead make more sense to create an additional layer of nesting within the `this.elements` data structure (e.g. `this.network.client.ethernet.static`), and iterate through the subsection of the map rather than create a new duplicate array?

static/js/views/settings.js
```diff
@@ -865,17 +867,17 @@ const SettingsScreen = {
     this.hideNetworkElements();
     this.elements.network.main.classList.remove('hidden');
 
-    // set default values in case things go bad
-    this.elements.network.client.ethernet.netmask.value = '255.255.255.0';
```
benfrancis, line 869: What's your rationale for removing this very common default?

static/js/views/settings.js
```diff
@@ -865,17 +867,17 @@ const SettingsScreen = {
...
     this.elements.network.client.ethernet.main.classList.remove('hidden');
 
     API.getLanSettings()
       .then((body) => {
         this.elements.network.client.ethernet.mode.value = body.mode || 'dhcp';
-        this.elements.network.client.ethernet.ip.value = body.ipdaddr || '';
-        this.elements.network.client.ethernet.netmask.value = body.netmask || '255.255.255.0';
-        this.elements.network.client.ethernet.gateway.value = body.gateway || '';
+        if (body.mode === 'static') {
+          this.ethernetEls.forEach((el) => el.classList.remove('hidden'));
```
benfrancis, line 876: I'm sure there's a more efficient way of doing this, like grouping the inputs into a fieldset and showing/hiding the fieldset instead, but I think this is OK if you'd rather not get into modifying the HTML structure.

adioanca:
Hi Ben,

I haven't noticed these comments, sorry. I will review my changes and your comments and be back in a few days(I'm in easter vacation these days) with an updated version.

Issue #3098, opened 2023-08-10
Re-factor "context menu" as separate dialogs


benfrancis:
Whilst working on #2938 I noticed there are currently UI elements referred to in the code as a "context menu" and "group context menu". They are not really menus, but a mixture of related dialogs.

I propose re-factoring these as separate dialog UI elements in a follow-up.

PR #3099, opened 2023-08-10
Refine groups UI - fixes #2938

### LINKED ISSUE #2938: Refine Groups UI

benfrancis:
Closes #2938. Supercedes #2991.

- Adds an overflow menu to the add thing button to add either a thing or a group
- Adds a dedicated add thing screen
- Refines the overflow menu design
- Refines the group "context menu" 
- Refines the group UI design including fixed width groups and new icons
- Fixes the overlapping groups bug and refines scrolling behaviour

Also:
- Replaces lots of nested divs and spans with more semantic HTML
- Uses HTML form elements for forms in order to take advantage of built-in client-side validation

TODO:
- [X] Fix re-arranging groups, which got a bit broken along the way
- [x] Optimise for mobile
- [x] Make linter happy

Follow-ups:
- Refactor the "context menu" and "group context menu" as separate dialogs (they're not really menus) #3098
- Use the refined overflow menu design everywhere #3104
- Potential for abstracting out a lot of common code in further follow-ups for cleaner code


benfrancis:
@bewee I did some work on polishing the groups UI with a view to getting it released (that was some great work you did BTW).

I've slightly broken re-arranging groups in the process (the white dotted separator doesn't appear in the right place when dragging and dropping). I'm hoping to take another look at it tomorrow before fixing any broken tests etc. and then submitting this PR for review. If you happen to have time to take a look you may figure out the problem before me!

benfrancis:
> I've slightly broken re-arranging groups

@bewee Never mind, I think I've fixed it.

tim-hellhake (approved):

Issue #2857, opened 2021-07-26
Implement W3C Directory Service API


benfrancis:
There is a proposed [Directory Service API](https://w3c.github.io/wot-discovery/#exploration-directory-api) which may provide a W3C standard equivalent to the [`Things`](https://webthings.io/api/#things-resource) resource in the Web Thing API.

Once this API has stabilised we should consider updating the gateway's API to match.

PR #3107, opened 2023-08-21
Only render things once - fixes #3103

### LINKED ISSUE #3103: Things get rendered twice
When debugging the gateway front end I have noticed when that when the Things screen is loaded, the collection of Things is actually rendered twice. Each Thing is rendered in turn, then the view is cleared, then they are rendered again.

I also note that the /things resource is fetched three times, and /groups twice. Something is causing a refresh during page load.

STR:
- Load /things

Expected:
- /things and /groups resources loaded once
- Things rendered once

Actual:
- /things resource loaded three times
- /groups resource loaded twice
- Things rendered twice

benfrancis:
This fixes #3103 by waiting for both the /things and /groups WebSocket connections to open before firing a refresh things event, rather than firing it twice.

I don't love this because it will slightly delay the rendering of Things on the screen until two TCP sockets have opened, but otherwise we're doing all the work twice.

There is still room for improvement because the /things endpoint is still fetched twice (it was fetched three times before), and I think in the thing detail view `refreshThing()` is still called a second time once the sockets have opened.

Further refinements could potentially be made by implementing the W3C WoT Discovery Directory Service API (#2857), which provides more information in event payloads so that the whole collection of things doesn't have to be re-rendered for every small change to a Thing.

### DIFF
--- a/static/js/models/gateway-model.js
+++ b/static/js/models/gateway-model.js
@@ -22,7 +22,7 @@ class GatewayModel extends Model {
     this.groups = new Map();
     this.onMessage = this.onMessage.bind(this);
     this.queue = Promise.resolve(true);
-    this.connectWebSocket();
+    this.connectWebSockets();
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


tim-hellhake:
If I'm not mistaken this will also change the behavior for reconnects.
Before `refreshThings()` was called for every `connect` event, now it's only called for the first `connect` event.
Is this intentional?

PR #3147, opened 2024-08-04
Implement OAuth 2.0 Authorization Server Metadata - closes #3143

### LINKED ISSUE #3143: Consider Implementing OAuth 2.0 Authorization Server Metadata specification
In order to support WoT Discovery [security bootstrapping](https://www.w3.org/TR/wot-discovery/#exploration-secboot) (e.g. for a third party app or cloud service to gain access to the Thing Description of a Thing), an OAuth2 client (i.e. the app or service) needs to discover the authorization and token endpoints of the authorisation server (i.e. gateway) in order to gain a user's authentication and request an access (and refresh) token.

[RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) defines a standard metadata format and well-known URL which a client can use to discover this information about an authorisation server.

Note that this technically doesn't fully solve the problem of discovering the authorization server from a Thing Description URL, because in OAuth2 the resource server (in this case the gateway serving Thing Descriptions) can exist at a separate origin to the authorization server (which issues authentication tokens). In the case of the gateway they exist at the same origin, but a third party OAuth2 client has know way of knowing this.

benfrancis:
Closes #3143.

- [x] Add a well-known controller
- [x] Implement Authorization Server Metadata
- [x] Integration tests

Following this implementation, a `GET` request to `/.well-known/oauth-authorization-server` will respond with a JSON document which looks like the following...

```
{
    "issuer": "http://localhost:8080",
    "authorization_endpoint": "http://localhost:8080/oauth/authorize",
    "token_endpoint": "http://localhost:8080/oauth/token",
    "response_types_supported": [
        "code"
    ],
    "scopes_supported": [
        "/things",
        "/things:readwrite"
    ]
}
```

...with the host set to match the host of the request.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -32,6 +32,7 @@ export const INTERNAL_LOGS_PATH = '/internal-logs';
 export const LOGS_PATH = '/logs';
 export const PUSH_PATH = '/push';
 export const PING_PATH = '/ping';
+export const WELL_KNOWN_PATH = '/.well-known';
 export const PROXY_PATH = '/proxy';
 export const EXTENSIONS_PATH = '/extensions';
 // Remember we end up in the build/* directory so these paths looks slightly
--- a/src/controllers/well-known_controller.ts
+++ b/src/controllers/well-known_controller.ts
@@ -1,0 +1,31 @@
+/**
+ * Well-Known Controller
+ *
+ * Handles HTTP requests to /.well-known
+ */
+
+import express from 'express';
+import * as Constants from '../constants';
+
+function build(): express.Router {
+  const controller = express.Router();
+
+  /**
+   * OAuth 2.0 Authorization Server Metadata (RFC 8414)
+   */
+  controller.get('/oauth-authorization-server', (request, response) => {
+    const origin = `${request.protocol}://${request.headers.host}`;
+    response.json({
+      issuer: origin,
+      authorization_endpoint: `${origin}${Constants.OAUTH_PATH}/authorize`,
+      token_endpoint: `${origin}${Constants.OAUTH_PATH}/token`,
+      response_types_supported: ['code'],
+      // Only expose top-level scopes to unauthenticated clients
+      scopes_supported: [Constants.THINGS_PATH, `${Constants.THINGS_PATH}:readwrite`],
+    });
+  });
+
+  return controller;
+}
+
+export default build;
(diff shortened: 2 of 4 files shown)


tim-hellhake (approved):

tim-hellhake:
@benfrancis 

> One concern I have with this implementation is that because a gateway may be served via multiple host names (e.g. foo.webthings.io or gateway.local or mydomain.com) it currently takes the host from the Host header in an HTTP request. I wonder if this might be vulnerable to a host header injection attack.

Technically, yes, but what's the point of fooling yourself?
Usually, the attacker fakes the `Host` header to trick the server into doing things it shouldn't.
Since the server does not rely on the `Host` header, the attacker is the only one affected.

There is one thing, though...
If a proxy caches the request, the attacker might be possible to poison the cache to lure users to a malicious server.

Keycloak, a pretty mature authorization server, also relies on the host header.
The key is to pin the hostname via the config parameter or configure the reverse proxy to override the `Host` header so that you can trust it.

The best way to mitigate this attack vector is to allow only `localhost` or the configured tunnel hostname.
In case of a custom setup, it would be nice to allow the user to configure the hostname.

> If so, is there a better way to determine the host the gateway is being served on? What if there are more than one or a custom static or dynamic DNS setup is in place?

Not that I know of.

tim-hellhake:
@benfrancis 

> One concern I have with this implementation is that because a gateway may be served via multiple host names (e.g. foo.webthings.io or gateway.local or mydomain.com) it currently takes the host from the Host header in an HTTP request. I wonder if this might be vulnerable to a host header injection attack.

Technically, yes, but what's the point of fooling yourself?
Usually, the attacker fakes the `Host` header to trick the server into doing things it shouldn't.
Since the server does not rely on the `Host` header, the attacker is the only one affected.

There is one thing, though...
If a proxy caches the request, the attacker might be possible to poison the cache to lure users to a malicious server.

Keycloak, a pretty mature authorization server, also relies on the host header.
The key is to pin the hostname via the config parameter or configure the reverse proxy to override the `Host` header so that you can trust it.

The best way to mitigate this attack vector is to allow only `localhost` or the configured tunnel hostname.
It would also be nice to allow the user to configure the hostname to support custom setups.

> If so, is there a better way to determine the host the gateway is being served on? What if there are more than one or a custom static or dynamic DNS setup is in place?

I don't see a way to achieve this without knowing the exact user setup.

benfrancis:
@tim-hellhake Thank you for the feedback. Let's look at providing a way to statically configure the canonical domain of the gateway rather than deriving it dynamically from a Host header in the future. I note there are other places in the codebase where we currently use the Host header too. This could be part of #82.

PR #3149, opened 2024-08-19
Initial snapcraft.yaml thanks to @ogra1 - closes #2147

### LINKED ISSUE #2147: Snap Package
As a user I want the gateway software to be available as a snap package so I can easily install it on my existing Linux computer.

This will involve creating a [snapcraft.yaml](https://snapcraft.io/docs/snapcraft-yaml-reference) file to generate a [snap package](https://snapcraft.io/docs) which can be installed on Linux, but also later bundled with an [Ubuntu Core based OS image](https://github.com/WebThingsIO/gateway/issues/3139).

benfrancis:
This is an initial snapcraft.yaml for WebThings Gateway created for us by @ogra1 at Canonical.

This is just a rough first pass and lots of things are expected to be broken, but it should be enough to publish a first edge release of a snap package that installs and starts.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,50 @@
+name: webthings-gateway
+base: core22
+version: '2.0.0'
+summary: WebThings Gateway
+description: |
+  A self-hosted web application for monitoring and controlling a building over
+  the web.
+
+grade: stable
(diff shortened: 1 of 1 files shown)


benfrancis:
r=me

PR #3165, opened 2024-09-05
Detect if running inside a snap or on Ubuntu Core

benfrancis:
This is the first step in porting WebThings Gateway to Ubuntu Core.

Ubuntu Core does not ship with the `lsb_release` command so we have to add some snap-specific code to check for the existence of a distro ID in `/var/lib/snapd/hostfs/etc/os-release`, which relates to the host OS. See https://forum.snapcraft.io/t/hosts-lsb-release-but-core-os-release/24351/4

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,6 +16,7 @@ apps:
     plugs:
       - network
       - network-bind
+      - system-observe
 
 parts:
   python-deps:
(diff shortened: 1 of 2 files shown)


benfrancis:
@ogra1 Is this the cleanest way of doing this?

dilyn-corner:
Yes, this is indeed the best way for a snap to know what the actual host environment is.

dilyn-corner (approved):

PR #3176, opened 2024-11-13
Upgrade from Node 10 to Node 20

benfrancis:
This PR upgrades the version of Node.js used by WebThings Gateway from version 10 to version 20.

This requires upgrading various npm packages and applying a bunch of TypeScript fixes, and updating the GitHub Action to run using Node.js 20.

TODO: 
- [x] Resolve merge conflicts since NetworkManager PR landed
- [x] Figure out why integration tests pass but then the process exits with an error
- [x] Update snap to use Node.js 20
- ~~[ ] Update docker image to use Node.js 20~~ Going to do this in a follow-up because the Docker image needs other work doing to it as well


benfrancis:
@tim-hellhake I think this is nearly ready, I just can't figure out why jest is exiting with an error in CI even though all the tests are passing. Any ideas?

benfrancis:
Finally got the tests to pass!

I'm going to split bb0a1ad592175ca63ea72ab27c8d09958ffadd64 out into a separate PR because I don't think it's strictly necessary and I'd like to better understand what we're losing, but otherwise I think this is ready to land.

tim-hellhake (approved):

PR #3229, opened 2025-12-15
Fix Thing names in rules engine UI - closes #3183

### LINKED ISSUE #3183: 'Rules' Editor difficult to use due to Things labelling inconsistencies
Name labels for Things used in rules change during edit.
For me at least, this makes efficient use of rules practically impossible.

Observed in 1.2.A on RPI3+ with PI OS and recent 2.0.0-alpha.1 on RPI5 with Ubuntu 24.04.

See attached pdf for description.

[RulesInconsistency.pdf](https://github.com/user-attachments/files/17918158/RulesInconsistency.pdf)

benfrancis:
This PR replaces some logic that was removed from a filtering function in RuleUtils.js in #2811 which is used to identify which Thing a Property belongs to.

This should fix the UI bugs that were reported in #3183.

### DIFF
--- a/static/js/rules/RuleUtils.js
+++ b/static/js/rules/RuleUtils.js
@@ -30,9 +30,16 @@ const RuleUtils = {
       console.warn('byProperty property undefined', new Error().stack);
       return false;
     }
-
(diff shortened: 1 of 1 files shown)


benfrancis:
Thank you @relu91 :)

Issue #3231, opened 2025-12-18
Replace node-getopt package with an alternative


benfrancis:
The node-getopt package is unmaintained and causing [errors on the console](https://github.com/WebThingsIO/gateway/issues/3232) due to a [bug](https://github.com/jiangmiao/node-getopt/issues/20).

```
2025-12-10 17:19:25.214 ERROR  : zigbee-adapter: (node:413459) [DEP0128] DeprecationWarning: Invalid 'main' field in 'gateway/node_modules/node-getopt/package.json' of './lib'. Please either fix that or report it to the module author
2025-12-10 17:19:25.214 ERROR  : zigbee-adapter: (Use `node --trace-deprecation ...` to show where the warning was created)
```

I suggest replacing it with something else like minimist, arg, clipanion, yargs or commander.

This library is used in addon-loader.ts

PR #3233, opened 2025-12-18
Temporarily fork node-getopt to fix error on command line - closes #3232

### LINKED ISSUE #3232: [DEP0128] DeprecationWarning: Invalid 'main' field
STR:
- Load Zigbee add-on version 2.0.0 on gateway built from master branch

Expected:
- No errors on the console

Actual:

```
2025-12-10 17:19:25.214 ERROR  : zigbee-adapter: (node:413459) [DEP0128] DeprecationWarning: Invalid 'main' field in 'gateway/node_modules/node-getopt/package.json' of './lib'. Please either fix that or report it to the module author
2025-12-10 17:19:25.214 ERROR  : zigbee-adapter: (Use `node --trace-deprecation ...` to show where the warning was created)
```

benfrancis:
This is a temporary workaround for #3232 by forking the unmaintained node-getopt package with a patch.

The long term fix of replacing the package will be tracked in #3231 but I don't want to block the 2.0.0 release on researching an alternative.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "netmask": "^2.0.2",
         "nocache": "^2.1.0",
         "node-fetch": "^2.6.7",
-        "node-getopt": "^0.3.2",
+        "node-getopt": "git://github.com/WebThingsIO/node-getopt.git",
         "promisepipe": "^3.0.0",
         "rimraf": "^3.0.2",
         "segfault-handler": "^1.3.0",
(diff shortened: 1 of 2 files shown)


benfrancis:
> Why not just remove that functionality instead? I've never used it, or had a need to.

That's a good point, and I don't generally run add-ons from the command line using addon-loader.js either. I'm definitely all for removing unnecessary code where possible.

However, now that I'm reminded it's possible to run addon-loader from the command line it occurs to me that this is potentially helpful when debugging add-ons (vs. just running the add-on's executable directly). 

The add-on loader sets up a database connection to load the add-on's saved settings, sets up an IPC connection with the gateway, loads the add-ons manifest and configuration and handles logging to the gateway. If you're just debugging the add-on's own logic you can do that by running the add-ons executable directly, but if you're debugging the add-ons communication with the gateway, testing with real settings or finding bugs with registration/initialisation then it's potentially useful to be able to execute addon-loader from the command line.

Debugging an add-on by running addon-loader on the command line vs. just letting the gateway start the add-on also presumably makes it easier to attach a debugger, see console output and stack traces directly and tweak environment variables without the gateway's logging system and process management getting in the way.

And now I've probably spent more time thinking about this than it would have taken to just replace the dependency 😅

PR #3253, opened 2026-02-02
Update make-self-signed-cert.sh

flatsiedatsie:
make the self-signed certificate last 10 years instead of the default 1 month

### DIFF
--- a/static/js/views/settings.js
+++ b/static/js/views/settings.js
@@ -590,7 +590,7 @@ const SettingsScreen = {
   },
 
   validateWifi: function () {
-    const valid = this.elements.network.client.wifi.password.value.length >= 8;
+    const valid = (this.elements.network.client.wifi.password.value.length == 0 || this.elements.network.client.wifi.password.value.length >= 8);
 
     this.elements.network.client.wifi.connect.disabled = !valid;
   },
@@ -1162,7 +1162,7 @@ const SettingsScreen = {
     const components = new Map();
 
     // First, get the list of installed add-ons. Do this separately so that the
-    // add-ons can be displayed when no internet connection is present.
+    // add-ons can be displayed when no internet ion is present.
     this.fetchInstalledAddonList(true)
       .then(() => {
         const addonList = document.getElementById('installed-addons-list');
--- a/tools/make-self-signed-cert.sh
+++ b/tools/make-self-signed-cert.sh
@@ -7,4 +7,4 @@ SSL_DIR="${WEBTHINGS_HOME}/ssl"
 [ ! -d "${SSL_DIR}" ] && mkdir -p "${SSL_DIR}"
 openssl genrsa -out "${SSL_DIR}/privatekey.pem" 2048
 openssl req -new -sha256 -key "${SSL_DIR}/privatekey.pem" -out "${SSL_DIR}/csr.pem" -subj '/CN=www.sgnihtbew.com/O=WebThings Gateway/C=US'
-openssl x509 -req -in "${SSL_DIR}/csr.pem" -signkey "${SSL_DIR}/privatekey.pem" -out "${SSL_DIR}/certificate.pem"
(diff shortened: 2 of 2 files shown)


benfrancis:
Is this currently causing a problem? I think this was originally added for automated testing. Are you using it for something else?

flatsiedatsie:
Interesting, I thought it was a feature. But the code suggests you're right 0_0.

In practise however, if you skipped the tunnel during the setup phase, then you can still access the gateway via https. As far as I'm aware this has always been possible? The https server will [still start](https://github.com/WebThingsIO/gateway/blob/0905f83ce393fc2ddea9a05746add2d3a5c89d53/src/app.ts#L85) if the certificates exist, even if the tunnel has been otherwise skipped.

(Candle has always disabled the tunnel feature, so the user doesn't even get the option to skip it)


**What's different now?**

- Candle 3.0 pushes users more actively to upgrade to https. For one, it's simply more secure. Secondly, they users can then use the QR code scanner that is built into the Matter adapter. There is an alternative scanner for http users too, but it involves communicating with a second browser window.
- On a read-only disk it's not easy to update the certificate, causing the gateway to only have a valid certificate for a month.
- A month is a very short period? I found multiple examples online that set the self-signed certificate to 3650 days by default.


This PR does not benefit Candle downstream, as it uses a much more modified version that will automatically increase encryption strengh in the coming years, and sets the domain to match the initial one.
https://github.com/createcandle/candle-controller/blob/master/tools/make-self-signed-cert.sh

I do hope you agree that using https on the local network is a useful feature to maintain. This PR aids that.

benfrancis (changes requested):
Whilst I understand the urge to set and forget a certificate so that local access is permanently "secure", I think it's probably misguided and there may be better solutions.

1. Modern browsers are getting stricter on certificate validity. They generally reject certificates with a validity period longer the 398 days, which is currently being tightened to 200 days, with plans to hit 47 days by 2029. Even if you manually "trust" the certificate, many Apple devices will still throw a "Connection Not Private" error. Whilst this may not always be a problem for self-signed certificates, browser engines are increasingly applying these sanity checks to all certificates to prevent old, weak encryption from lingering forever.
2. Public certificates can be revoked if a gateway is compromised, but self-signed certificates can not. A certificate with a 10 year window acts as a golden key for intercepting traffic for the next decade.
3. Ten years from now algorithms that are considered secure today may be easily crackable, not least because of quantum computing.

If using self-signed certificates for encrypted local access rather than just for unit tests a one year validity may be more reasonable as a default, which should hopefully currently be accepted by most browsers.

A more sophisticated approach might be to create a Local Certificate Authority (Root CA) with a root certificate with a validity of 10 years, then a leaf certificate with a validity of 1 year. Or even better, use certbot to automatically renew the certificate every 90 days.

flatsiedatsie:
> being tightened to 200 days, with plans to hit 47 days by 2029.

Thank you, I didn't know that. 

I also later realized that since the certificate is stored on ~/.webthings/ssl, it can in fact be eaily written and changed.

Coupled with what you explained I agrree the suggested approach in this PR is a bad idea.

flatsiedatsie:
> A more sophisticated approach might be to create a Local Certificate Authority (Root CA) with a root certificate with a validity of 10 years, then a leaf certificate with a validity of 1 year. Or even better, use certbot to automatically renew the certificate every 90 days.

I will have to look into that.

Again, thank you for thinking along!

flatsiedatsie:
Feel free to copy and modify this version if that's of interest:

https://github.com/createcandle/candle-controller/blob/master/tools/make-self-signed-cert.sh

flatsiedatsie:
And this script will check if the certificate is about to expire (in three days) and generate a fresh one:

https://github.com/createcandle/configuration-files/blob/main/etc/cron.daily/ensure_certificate.sh

PR #3254, opened 2026-02-05
Update settings.js

flatsiedatsie:
allow connecting to open wifi networks


flatsiedatsie:
That makes sense.

benfrancis (changes requested):
Should not require the user to enter a password if one is not required

PR #3259, opened 2026-02-13
Fix update settings UI on platforms where updates not configurable

benfrancis:
Currently the update settings UI does not really account for platforms (like Ubuntu Core) where automatic updates are enabled but are not user configurable or triggerable from the UI.

This PR adds a bit more nuance to the update settings UI so that it can accurately represent the status of that kind of platform.


benfrancis:
r=me

Issue #3275, opened 2026-03-24
UX - The thing overview can get crowded if there are many properties


flatsiedatsie:
Just wanted to share what the [IKEA Bilresa scroll wheel](https://duckduckgo.com/?q=ikea+bilresa+scroll+wheel&t=brave&ia=images&iax=images) Matter remote control looks like with the latest version of the Matter Adapter. 

<img width="1619" height="933" alt="Image" src="https://github.com/user-attachments/assets/1c503e70-c011-4cb1-8a04-8dcd5bfaffd6" />
(I had to zoom out to fit it all in one screenshot)

.

Granted:
 - This device does stand out in how many attributes it exposes, as IKEA has used the Matter standard "creatively' to create their scrollwheel's functionality, which the standard didn't technically have an implementation for. Their light bulbs are much more 'normal', for example.
- The adapter exposes a lot of Matter properties. But that is done so users can create rules from all possible data that the device spits out. In fact, users can choose to turn _even more_ of Matter's attributes into additional properties. This is just how the device ends up by default.

Because of how Candle tries to stay as compatible as possible with the Webthings Gateway, this squared design is 100% done through CSS overrides in the [Candle Theme](https://github.com/createcandle/candle-theme). If a user disables that theme they revert to the original Webthings Gateway UI. In which case they would not be able to see anything, as all the little circles overlap.

flatsiedatsie:
I looked into it: definitely not a one-line change to add `part` on all properties. But I could probably whip up a python script that automates duplicating class to part.

benfrancis:
@flatsiedatsie wrote:
>  ideally there would be an overarching visual distinction between "clicking on this circular icon toggles something directly" and "clicking on this circular icon takes you to a detail view of the thing"

To be clear, clicking the circular icon never navigates to the Thing detail view. It either toggles a state, or it does nothing - exactly the same as the current behaviour. I understand your point about differences in visual affordances between something that is clickable and something that is not though. That distinction could also be more subtle, like a border or box shadow on clickable elements to make them appear more like buttons.

> this would be a great opportunity to add native 'part' support to the capabilities

> Similarly, for theming it would rock if the shadowroot container element had a few extra classes to indicate what's inside, based on the property. For example, "has-read-only", "has-type-integer","has-type-string".

[...]

PR #3294, opened 2026-04-11
Simpler thing detail view

flatsiedatsie:
This replaces the 'octopus view' with a responsive grid-based design.

Screenshot 2026-04-11 at 17 40 28

### DIFF
--- a/static/css/thing.css
+++ b/static/css/thing.css
@@ -71,11 +71,19 @@
   position: relative;
 }
 
-.thing-detail-contents {
-  position: absolute;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
+#things.single-thing > .thing {
+  width:100%;
+  display: flex;
+  flex-wrap: wrap;
+}
+
+#things.single-thing > .thing > * {
+  display: flex;
+  padding: 1rem;
+  min-height: 14rem;
+  min-width: 14rem;
+  align-items: flex-start;
+  justify-content: center;
 }
 
 .thing-detail-label {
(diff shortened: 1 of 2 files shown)


benfrancis (changes requested):
Thank you for taking the time to work on this, but I am really not keen on this design. It looks a bit like a grid of app icons, but with one icon randomly bigger than the others, which doesn't make sense from an information hierarchy point of view. I was also hoping to move away from circles representing each interaraction affordance because they are tricky to design around.

Let's continue to discuss the user interface design in #3275 before jumping to implementation.

As I mentioned, I hope to address this as part of a wider UI refresh, which also aligns with [app](https://docs.google.com/presentation/d/1qRuAxqvSS4c_nq26Lpa_51wsutIILY1Q-LANxblI5qo/edit?slide=id.g1252755ab01_0_0#slide=id.g1252755ab01_0_0) and [cloud](https://docs.krellian.com/cloud/devices/view-device/).

flatsiedatsie:
I totally get it, and fully agree.

Candle does implement this, just to make the (rare) flash of the octopus before the Candle Theme CSS loads and/or adds the `part` attributes to the components less jarring. And since the Candle theme can still be disabled, users can now at least see all their device properties if they do so :-)

So I spent some time figuring out how to do it and had the code ready, so thought I'd make a PR in the hopes that it would be helpful towards that redesign in some way, as a starting point.

Also, I think in the actual code the Capability icon is no longer larger than the others. I may have picked the wrong screenshot.

PR #3297, opened 2026-04-12
Allow extensions to subscribe to things

flatsiedatsie:
A better implementation

See also: https://github.com/WebThingsIO/gateway/pull/3291

### DIFF
--- a/static/js/extension.js
+++ b/static/js/extension.js
@@ -19,7 +19,8 @@ class Extension {
       timezone: App.TIMEZONE,
       units: App.UNITS,
     };
-    this.subscriptions = {};
+    this.property_subscriptions = {};
+    this.event_subscriptions = {};
     this.view = App.registerExtension(this);
   }
 
@@ -89,7 +90,7 @@ class Extension {
   }
 
   /**
-   * Allow extensions to subscribe to updates from Things so that
+   * Allow extensions to subscribe to property changes from Things so that
    * they don't need to create and manage their own websocket clients
    */
   subscribeToThingProperties(thingId,handler) {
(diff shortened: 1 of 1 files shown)


flatsiedatsie:
Let me know if it's worth my time to prettify it? Fine either way.

flatsiedatsie:
Also, maybe the developer should just create their own dict to store the functions, to simplify it.

benfrancis:
> maybe the developer should just create their own dict to store the functions, to simplify it.

Unless I'm missing something, extensions should just be able to call `thingModel.subscribe()` directly to register a callback for changes to a property they are interested in.

That shouldn't require any modifications to the gateway, so I'm going to close this PR, but if I'm misunderstanding and that isn't possible then please re-open.

(In the future [`Thing.observeAllProperties()`](https://github.com/w3c/wot-scripting-api/issues/312) via the WoT Scripting API could be even better.)

flatsiedatsie:
Wait.. really? Could you explain how? 

I've tried calling `thingX.subscribe()` in the past, but it didn't work.

thing

```
typeof thing.subscribe:  undefined
typeof this.subscribe:  undefined
typeof thing.thingModel  undefined
typeof this.thingModel  undefined
typeof thing.APP:  undefined
typeof thing.App:  undefined
typeof this.APP:  undefined
typeof this.App:  undefined
```

benfrancis:
Ah yes, I see the problem now. I was thinking that GatewayModel was exposed via the window object somehow, but it isn't. The Extension class uses App internally (where App is a module-scoped singleton required locally in the extension class, but not exposed publicly to Extension instances, or globally on the window object).

So what you are trying to do here is reasonable, but I am a bit reluctant to start creating a temporary public API for this if we intend to change it in the next version, since that will just cause more migration headaches for add-ons.

I would honestly suggest just continuing to open new WebSocket connections in your extension for now, until we have given some proper thought to a long term public API for extensions.

flatsiedatsie:
Understood. 

I am already using this in my addons however.

I was thinking maybe the naming convention could already be put into place, and then this version could be seamlessly replaced later? But I understand if even that might be uncertain.

benfrancis:
Let's continue to discuss the shape of the API in https://github.com/WebThingsIO/gateway/issues/3292

